PR #54, opened 2017-04-03
Implement EeGenKtAlgorithm

Bastiantheone:


fmom/pxpypze.go
```diff
@@ -211,3 +211,17 @@ func (p4 *PxPyPzE) Set(p P4) {
 	p4[2] = p.Pz()
 	p4[3] = p.E()
 }
+
+func dotProduct(p1, p2 *PxPyPzE)float64{
+	dotproduct := p1.Px()*p2.Px()+p1.Py()*p2.Py()+p1.Pz()*p2.Pz()
+	return dotproduct
+}
+
+//return the cosine of the angle between the momentum of two jets
```
sbinet: Please use the canonical way of documenting Go code:

```go
// CosTheta returns the cosine of the angle between the momentum of two 4-vectors.
func CosTheta(...) ...
```

please also specify whether the returned angle is in radians or in degrees, and the expected range ([0,2pi), [-pi,+pi), ...)

fmom/pxpypze.go
```diff
@@ -211,3 +211,17 @@ func (p4 *PxPyPzE) Set(p P4) {
...
+func CosTheta(p1, p2 *PxPyPzE) float64 {
```
sbinet: rather than taking two `*PxPyPzE`, perhaps take 2 `P4` interfaces instead.
it may be slower but more general :) (and push comes to shove, we may implement fast-path later on)

fmom/pxpypze.go
```diff
@@ -211,3 +211,17 @@ func (p4 *PxPyPzE) Set(p P4) {
...
+	mag1 := math.Sqrt(p1.Px()*p1.Px()+p1.Py()*p1.Py()+p1.Pz()*p1.Pz())
```
sbinet: replace with:

```go
mag1 := p1.P()
mag2 := p2.P()
```

fmom/pxpypze.go
```diff
@@ -211,3 +211,17 @@ func (p4 *PxPyPzE) Set(p P4) {
 	p4[2] = p.Pz()
 	p4[3] = p.E()
 }
+
+func dotProduct(p1, p2 *PxPyPzE)float64{
```
sbinet: make `dotProduct` take 2 `P4` interfaces, like for `CosTheta`.

fastjet/clustersequence.go
```diff
@@ -400,3 +428,4 @@ func (cs *ClusterSequence) runN3Dumb() error {
 	}
 	return err
 }
+
```
sbinet: remove blank like.

fastjet/clustersequence.go
```diff
@@ -360,11 +369,30 @@ func (cs *ClusterSequence) runN3Dumb() error {
...
 					cs.jetScaleForAlgorithm(jjet),
 				)
-				y := jetscale * Distance(ijet, jjet) * cs.invR2
-				if y < ymin {
-					ymin = y
-					ii = i
-					jj = j
+				switch cs.alg {
+				case EeGenKtAlgorithm:	//distance between jets min(diB, djB)*(1-cos theta)/(1-cos R)
+					denominator := 1- math.Cos(cs.r)
```
sbinet: Go convention would be to rename `denominator` as just `den`

fmom/pxpypze.go
```diff
@@ -211,3 +211,17 @@ func (p4 *PxPyPzE) Set(p P4) {
...
+	dotproduct := p1.Px()*p2.Px()+p1.Py()*p2.Py()+p1.Pz()*p2.Pz()
+	return dotproduct
+}
+
+//return the cosine of the angle between the momentum of two jets
+func CosTheta(p1, p2 *PxPyPzE) float64 {
+	mag1 := math.Sqrt(p1.Px()*p1.Px()+p1.Py()*p1.Py()+p1.Pz()*p1.Pz())
+	mag2 := math.Sqrt(p2.Px()*p2.Px()+p2.Py()*p2.Py()+p2.Pz()*p2.Pz())
+	dotproduct := dotProduct(p1,p2)
```
sbinet: `s/dotproduct/dot/`

sbinet:
@ebusato could you also take a look at this (especially the `fastjet` part)? thanks.

sbinet:
@dhrou could you also take a look at this (especially the `fmom` part)? thanks.

fmom/pxpypze.go
Bastiantheone: @sbinet I will change it up. What do you mean by s/dotproduct/dot/ ?

fmom/pxpypze.go
Bastiantheone: It returns the cosine of an angle and not the angle itself, so it's just a value between -1 and 1. Would it be useful to have a function that returns the angle as well?

sbinet: arf. I was so laser-focused on the angles issue in another project...
we'll start with just the cosine of the angle and add a function for the angle if there's a real need.

fmom/pxpypze.go
sbinet: apologies for the slang.
I was just saying to rename the `dotproduct` variable into just `dot`.

dhrou:
You have  P() and  P2() functions.
You should use them consistently in M2(), Eta(), CosTheta() (and of course not use Px(), Py(), Pz() unless really necessary. 
So that if one day you are using a different internal representation (e.g. Pt, eta, phi, E) the implementation of the above remains optimal.

PR #60, opened 2017-04-06
Fmom: add Cos(Theta) and Dot Product between 2 four-vectors to utils

Bastiantheone:


Bastiantheone:
Great :)

sbinet:
needs A+C from go-hep/license#3

Issue #63, opened 2017-04-10
fastjet: implement EeKtAlgorithm


sbinet:
now that EeGenKtAlgorithm is in, we should also implement the e+e- kt algorithm to be on par with the set of features of the `C++` version of FastJet.

@Bastiantheone do you want to give it a try? :)

Bastiantheone:
Yes I can do that

Bastiantheone:
@sbinet can you provide me with testdata files for this algorithm?

sbinet:
yep. will do as soon as I am in front of a real keyboard :)

sbinet:
@Bastiantheone see be0fad3.
Note that, for e+e- kt, the original test from `C++` FastJet runs with exclusive jets. so this mode would need to be implemented first. (I've created go-hep/hep#65 to track that)

PR #66, opened 2017-04-12
Implement ExclusiveJets & EeKtAlgorithm 

Bastiantheone:
Solves #65 & #63 
I tried to model the exclusive jets after the c++ version

### DIFF
--- a/fastjet/clustersequence.go
+++ b/fastjet/clustersequence.go
@@ -5,6 +5,7 @@
 package fastjet
 
 import (
+	"errors"
 	"fmt"
 	"math"
 
@@ -35,6 +36,7 @@ type ClusterSequence struct {
 	r2       float64
 	invR2    float64
 	qtot     float64
+	initn    int
 
 	jets      []Jet
 	history   []history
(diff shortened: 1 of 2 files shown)


fastjet/clustersequence.go
```diff
@@ -69,8 +72,63 @@ func NewClusterSequence(jets []Jet, def JetDefinition) (*ClusterSequence, error)
 	return cs, err
 }
 
+//returns the number of exclusive jets that would have been obtained
```
sbinet: the Go style recommends to document functions like so:
```go
// NrExclusiveJets returns the number of ...
func (cs *ClusterSequence) NrExclusiveJets(...) { ...
```
also, `NrExclusiveJets` doesn't look too Go-ish.
see:
```
$> go doc flag.NArg 
func NArg() int
    NArg is the number of arguments remaining after flags have been processed.

$> go doc reflect.Type.NumField
// NumField returns a struct type's field count.
// It panics if the type's Kind is not Struct.
func NumField() int
```

so, well, even though the standard library (itself) doesn't follow a consistent style, `NrFoo` isn't part of that set of inconsistent options :)

I would go with `NumExclusiveJets`.

sbinet:
ah! one more thing.

please add in your commit message:

```
Fixes go-hep/hep#63
Fixes go-hep/hep#65
```

so github can automatically close these issues when the CL will be merged.

fastjet/clustersequence.go
```diff
@@ -69,8 +72,63 @@ func NewClusterSequence(jets []Jet, def JetDefinition) (*ClusterSequence, error)
...
+		err = errors.New("2*initial nr jets != size of history -- this endagers internal assumptions!")
+		panic(err)
+	}
+
+	//now go forwards and reconstitute the jets that we have --
+	//basically for any history element, see if the parent jets to
+	//which it refers were created before the stopping point -- if they
+	//were then add them to the list, otherwise they are subsequent
+	//recombinations of the jets that we are looking for
+	ljets := make([]Jet, 0)
```
Bastiantheone: In the C++ version it checks at the end if ljets is the same size as the min of initn and njets. Would it be better to make a slice with that capacity: ljets := make([]Jet,0, imin(njets,initn)) ?

sbinet: yes, if a good guess for the capacity can be provided, then we should file it.
that's that much less re-allocations to perform at runtime.

Bastiantheone:
All changed up :)  I wouldn't mind implementing another one

sbinet:
thanks a lot!

sbinet:
next would be to work on the jet areas (p36 of http://fastjet.fr/repo/fastjet-doc-3.2.1.pdf) -- which would be a nice segway into voronoi diagrams and delaunay triangulations.

(I am out of town until Monday, though)

Bastiantheone:
Sounds good

Issue #64, opened 2017-04-11
rootio/cmd/root-srv: implement https+letsencrypt


sbinet:
add support for https (and thus, HTTP/2) with Let's Encrypt.

sbinet:
@mughetto, do you want to give that a spin?

sbinet:
(it's relatively easy, see:

sbinet:
and it could even be easier with the new `NewListener` feature:
https://godoc.org/golang.org/x/crypto/acme/autocert#example-NewListener

mughetto:
Having a look at the moment

mughetto:
Got a nearly working version, just got to figure out how to configure a certificate on localhost.

Do you want to support both http and https or systematically redirect http connections to https?

mughetto:
Any idea how I could test witout having a public domain name certbot could use?

mughetto:
Got a working test version, just want to implement a proper http->https redirection in case the server runs https before making the PR

sbinet:
(sorry for the belated answer)

here is an example of this http->https redirection:
https://github.com/go-hep/go-hep.org/blob/master/main.go#L51

(but you can already send your PR, and we'll iterate)

PR #67, opened 2017-04-13
rootio/cmd/root-srv: add https+letsencrypt support

mughetto:
PR solving #64.

Few additional features to (maybe) add during review process:
- Redirect http traffic to https (WIP)
- Sanity check on flags (enforce only 2 possible values for serv flag (http/https), check that CA domain name is given for the https case)
- Do we want a single server instance to serve both http and https requests? (if even possible)


sbinet:
(apologies for the belated answer: I blame the easter bunny)

Thanks for the PR!

but... the TravisCI issues are real.
go-hep supports Go-1.6 (we try to follow what AppEngine supports and not move too fast wrt Go compiler versions, when that make sense)

it seems `golang.org/x/crypto/...` uses the standard library `context.Context` in its bowels.
I would feel a bit sad to require Go-1.7 "just" for this nice-to-have https+http/2 feature for `root-srv`.

could you perhaps arange your CL to only enable the compilation of the https+acme feature for Go>=1.7 ?

that's usually done with build tags: 
- https://golang.org/pkg/go/build/#hdr-Build_Constraints
- https://dave.cheney.net/2013/10/12/how-to-use-conditional-compilation-with-the-go-build-tool

the other failures (ie: for Go>=1.7) come from the fact you didn't gofmt'ed your code before submitting it :)
see: https://go-hep.org/contributing/#code-contribution
(especially the part about `goimports` https://godoc.org/golang.org/x/tools/cmd/goimports)

thanks again for the PR

mughetto:
Hey thanks for the command, thought I gofmt-ed it but indeed wasn't the case ^^ Still learning my way :D 

Working on the conditionnal build part.

rootio/cmd/root-srv/main.go
```diff
@@ -13,22 +14,28 @@
 //  $> root-srv -addr=:8080 &
 //  2017/04/06 15:13:59 server listening on :8080
 //
+
```
sbinet: please remove that blank line (otherwise the documentation for the `root-srv` command in godoc only reads `$> open localhost:8080` :))

rootio/cmd/root-srv/main.go
```diff
@@ -39,8 +46,9 @@ func main() {
 
 ex:
 
- $> root-srv -addr=:8080
- 2017/04/06 15:13:59 server listening on :8080
+
```
sbinet: perhaps also add this example to the `package main` documentation, above.

rootio/cmd/root-srv/main_go16.go
```diff
@@ -0,0 +1,77 @@
+// +build !go1.7
```
sbinet: same comment than for the other file :)
please move this at the correct location.

rootio/cmd/root-srv/main_go16.go
```diff
@@ -0,0 +1,77 @@
...
+// TGraphAsymmErrors, TDirectories, TTrees, ...).
+//
+// Usage: root-srv [options]
+//
+// ex:
+//
+//  $> root-srv -addr=:8080 &
+//  2017/04/06 15:13:59 server listening on :8080
+//
+
```
sbinet: please remove that blank line to not break the documentation blurb in 2.

rootio/cmd/root-srv/main_go16.go
```diff
@@ -0,0 +1,77 @@
...
+func main() {
+	flag.Usage = func() {
+		fmt.Fprintf(
+			os.Stderr,
+			`Usage: root-srv [options]
+
+ex:
+
+
+ $> root-srv -addr :8080 -serv https -host example.com
```
sbinet: hum... 
it doesn't seem very helpful to give an example that wouldn't work.
ie: you shouldn't show an example with `-serv https` in the executable that won't do anything when given that option :)

rootio/cmd/root-srv/main_go16.go
```diff
@@ -0,0 +1,77 @@
...
+
+	flag.Parse()
+	server.Init()
+
+	log.Printf("%s server listening on %s", *servFlag, *addrFlag)
+
+	if *servFlag == "http" {
+		log.Fatal(http.ListenAndServe(*addrFlag, nil))
+	} else if *servFlag == "https" {
+		//		m := autocert.Manager{
```
sbinet: perhaps, instead of doing this, we should just panic when `"https"` (and tell why.)

then, reverse the logic of the `if` to read:

```go
if *servFlag != "http" {
   log.Fatalf("serv=%q not supported. please use %q", *servFlag, "http")
}
```
or something along these lines.

mughetto: I'm actually working on implementing it for the 1.6 case, so it should be fixed in the next commit (although I can't fully test it), I didn't realize every push would ring bells on your side :D

sbinet:
(and thanks for tackling this!)

mughetto:
Latest commit should fix every comments from @sbinet 

Do you have any working setup to test the https implementation for 1.6 ?

rootio/cmd/root-srv/main_go16.go
```diff
@@ -0,0 +1,69 @@
...
+// ex:
+//
+//  $> root-srv -addr=:8080 &
+//  2017/04/06 15:13:59 server listening on :8080
+//
+//  $> open localhost:8080
+//
+//  $> root-srv -addr :8080 -serv https -host example.com
+//  2017/04/06 15:13:59 https server listening on :8080 at example.com
+
```
sbinet: please remove this empty line.
(otherwise the documentation won't be diplsayed by godoc)

sbinet:
(coming back to this...)
no, I don't have a working setup for https+1.6.

travis-ci will make sure it will at least compile with 1.6.

rootio/cmd/root-srv/main_go16.go
mughetto: Ok should be done assuming my git-fu didn't get worse during summer.

sbinet:
Closed via 0336890 (I just did a bunch of squashing+merging :))

PR #75, opened 2017-07-24
predicates: implement predicates for delaunay

Bastiantheone:
This PR is the first step for the delaunay implementation. It provides a package for the geometric predicates.

First this implementation uses float64 values to determine the orientation and if a point is in a circle. If the point is too close to 0 the calculation is reevaluated using big/Rat values. I still need to figure out a good value that determines when it is too close to 0. When researching this topic I saw that some people determine this value dynamically, based on the input values. Does anyone have any suggestion here?

### DIFF
--- a/fastjet/internal/predicates/float64Pred.go
+++ b/fastjet/internal/predicates/float64Pred.go
@@ -14,12 +14,12 @@ const (
 	// representing a number.
 	// If y is the machine representation of x then |(x-y)/x| <= macheps
 	// https://en.wikipedia.org/wiki/Machine_epsilon
-	// Golangs float64 type has a 52-bit fractional mantissa,
+	// Go's float64 type has a 52-bit fractional mantissa,
 	// therefore the value 2^-52
 	macheps = 1.0 / (1 << 52)
 )
 
-// float64Pred dynamically updates the potential error
+// float64Pred dynamically updates the potential error.
 //
 // If y is the machine representation of x then |(x-y)/x| <= macheps and |x-y| = e.
 // Since we want the max possible error we assume |(x-y)/x| = macheps
--- a/fastjet/internal/predicates/float64Pred.go
+++ b/fastjet/internal/predicates/float64Pred.go
@@ -1,0 +1,91 @@
+// Copyright 2017 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package predicates
+
+import "math"
+
+const (
+	// macheps is the machine epsilon aka unit roundoff
+	// The machine epsilon is an upper bound on the absolute relative true error in
+	// representing a number.
+	// If y is the machine representation of x then |(x-y)/x| <= macheps
+	// https://en.wikipedia.org/wiki/Machine_epsilon
+	// Golangs float64 type has a 52-bit fractional mantissa,
+	// therefore the value 2^-52
+	macheps = 1.0 / (1 << 52)
+)
+
+// float64Pred dynamically updates the potential error
+type float64Pred struct {
+	// n is the number
+	n float64
+	// e is the max rounding error possible
+	e float64
+}
+
+// newFloat64Pred returns a new float64Pred e set to 0.
+func newFloat64Pred(n float64) float64Pred {
+	return float64Pred{
+		n: n,
+		e: 0,
+	}
+}
+
+// addFloat64 adds f to p and updates the potential error
+//
+// If y is the machine representation of x then |(x-y)/x| <= macheps and |x-y| = e, therefore
+// e = macheps*|x|.
+func (p float64Pred) addFloat64(f float64) float64Pred {
+	p.n += f
+	p.e += macheps * math.Abs(p.n)
+	return p
+}
+
+// addFloat64Pred adds b to a and updates the potential error
+//
+// If y is the machine representation of x then |(x-y)/x| <= macheps and |x-y| = e, therefore
+// e = macheps*|x|.
+func (a float64Pred) addFloat64Pred(b float64Pred) float64Pred {
+	a.n += b.n
+	a.e += macheps*math.Abs(a.n) + b.e
+	return a
+}
+
+// subFloat64 subtracts f from p and updates the potential error
+//
+// If y is the machine representation of x then |(x-y)/x| <= macheps and |x-y| = e, therefore
+// e = macheps*|x|.
+func (p float64Pred) subFloat64(f float64) float64Pred {
+	p.n -= f
+	p.e += macheps * math.Abs(p.n)
+	return p
+}
+
+// subFloat64Pred subtracts f from p and updates the potential error
+//
+// If y is the machine representation of x then |(x-y)/x| <= macheps and |x-y| = e, therefore
+// e = macheps*|x|.
+func (a float64Pred) subFloat64Pred(b float64Pred) float64Pred {
+	a.n -= b.n
+	a.e += macheps*math.Abs(a.n) + b.e
+	return a
+}
+
+// mulFloat64 multiplies p with f and updates the potential error.
+//
+// If y is the machine representation of x then |(x-y)/x| <= macheps and |x-y| = e, therefore
+// e = macheps*|x|.
+//
+// mul(mul(a,b),c) = mul(a*b+error,c) = a*b*c + error*c + newError
+// sum(mul(a,b),c) = sum(a*b+error,c) = a*b+c + error + newError
+// Conclusively, when multiplications are chained, the error also depends on the value
+// of the number, but this does not apply to sums or subtractions.
+//
+//If this is not a chained multiplication p.e will be 0, making that part irrelevant.
+func (p float64Pred) mulFloat64(f float64) float64Pred {
+	p.n *= f
+	p.e += macheps*math.Abs(p.n) + p.e*f
+	return p
+}
--- a/fastjet/internal/predicates/float64Pred.go
+++ b/fastjet/internal/predicates/float64Pred.go
@@ -4,7 +4,9 @@
 
 package predicates
 
-import "math"
+import (
+	"math"
+)
 
 const (
 	// macheps is the machine epsilon aka unit roundoff
@@ -18,6 +20,19 @@ const (
 )
 
 // float64Pred dynamically updates the potential error
+//
+// If y is the machine representation of x then |(x-y)/x| <= macheps and |x-y| = e.
+// Since we want the max possible error we assume |(x-y)/x| = macheps
+// macheps*|x| = |x - y|
+// if (x-y)>=0 then macheps*|x| = x - y  ->  y = x- macheps*|x|
+// else macheps*|x| = -(x - y)  ->  macheps*|x|+x = y
+// Each one of these has two cases again. x can be positive or negative, resulting in the same two possible equations:
+// y/(1-macheps)=x or y/(1+macheps)=x.
+// Since x is unknown, we will use the larger number as factor, to avoid having an error greater than
+// the maxError value we have.
+// |y-x| = macheps*|x| -> e = macheps*|x| -> e = macheps* |y|/(1-macheps)
+// A Special case is when y = 0. Then we use the smallest nonzero float, because that is the max
+// possible error in this case.
 type float64Pred struct {
 	// n is the number
 	n float64
@@ -34,12 +49,13 @@ func newFloat64Pred(n float64) float64Pred {
 }
 
 // addFloat64 adds f to p and updates the potential error
-//
-// If y is the machine representation of x then |(x-y)/x| <= macheps and |x-y| = e, therefore
-// e = macheps*|x|.
 func (p float64Pred) addFloat64(f float64) float64Pred {
 	p.n += f
-	p.e += macheps * math.Abs(p.n)
+	if p.n == 0 {
+		p.e += math.SmallestNonzeroFloat64
+	} else {
+		p.e += macheps * math.Abs(p.n) / (1 - macheps)
+	}
 	return p
 }
 
@@ -49,34 +65,37 @@ func (p float64Pred) addFloat64(f float64) float64Pred {
 // e = macheps*|x|.
 func (a float64Pred) addFloat64Pred(b float64Pred) float64Pred {
 	a.n += b.n
-	a.e += macheps*math.Abs(a.n) + b.e
+	if a.n == 0 {
+		a.e += math.SmallestNonzeroFloat64 + b.e
+	} else {
+		a.e += macheps*math.Abs(a.n)/(1-macheps) + b.e
+	}
 	return a
 }
 
 // subFloat64 subtracts f from p and updates the potential error
-//
-// If y is the machine representation of x then |(x-y)/x| <= macheps and |x-y| = e, therefore
-// e = macheps*|x|.
 func (p float64Pred) subFloat64(f float64) float64Pred {
 	p.n -= f
-	p.e += macheps * math.Abs(p.n)
+	if p.n == 0 {
+		p.e += math.SmallestNonzeroFloat64
+	} else {
+		p.e += macheps * math.Abs(p.n) / (1 - macheps)
+	}
 	return p
 }
 
 // subFloat64Pred subtracts f from p and updates the potential error
-//
-// If y is the machine representation of x then |(x-y)/x| <= macheps and |x-y| = e, therefore
-// e = macheps*|x|.
 func (a float64Pred) subFloat64Pred(b float64Pred) float64Pred {
 	a.n -= b.n
-	a.e += macheps*math.Abs(a.n) + b.e
+	if a.n == 0 {
+		a.e += math.SmallestNonzeroFloat64 + b.e
+	} else {
+		a.e += macheps*math.Abs(a.n)/(1-macheps) + b.e
+	}
 	return a
 }
 
 // mulFloat64 multiplies p with f and updates the potential error.
-//
-// If y is the machine representation of x then |(x-y)/x| <= macheps and |x-y| = e, therefore
-// e = macheps*|x|.
 //
 // mul(mul(a,b),c) = mul(a*b+error,c) = a*b*c + error*c + newError
 // sum(mul(a,b),c) = sum(a*b+error,c) = a*b+c + error + newError
@@ -86,6 +105,10 @@ func (a float64Pred) subFloat64Pred(b float64Pred) float64Pred {
 //If this is not a chained multiplication p.e will be 0, making that part irrelevant.
 func (p float64Pred) mulFloat64(f float64) float64Pred {
 	p.n *= f
-	p.e += macheps*math.Abs(p.n) + p.e*math.Abs(f)
+	if p.n == 0 {
+		p.e += math.SmallestNonzeroFloat64 + p.e*math.Abs(f)
+	} else {
+		p.e += macheps*math.Abs(p.n)/(1-macheps) + p.e*math.Abs(f)
+	}
 	return p
 }
--- a/fastjet/internal/predicates/float64Pred.go
+++ b/fastjet/internal/predicates/float64Pred.go
@@ -86,6 +86,6 @@ func (a float64Pred) subFloat64Pred(b float64Pred) float64Pred {
 //If this is not a chained multiplication p.e will be 0, making that part irrelevant.
 func (p float64Pred) mulFloat64(f float64) float64Pred {
 	p.n *= f
-	p.e += macheps*math.Abs(p.n) + p.e*f
+	p.e += macheps*math.Abs(p.n) + p.e*math.Abs(f)
 	return p
 }
--- a/fastjet/internal/predicates/incircle.go
+++ b/fastjet/internal/predicates/incircle.go
@@ -1,0 +1,106 @@
+// Copyright 2017 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package predicates
+
+import (
+	"math/big"
+)
+
+const (
+	// FIXME
+	epsilonI = 1e-13
+)
+
+// Incircle determines whether the point (x,y) lies inside the circle formed
+// by the three points (x1,y1),(x2,y2) and (x3,y3)
+func Incircle(x1, y1, x2, y2, x3, y3, x, y float64) bool {
+	det := simpleIncircle(x1, y1, x2, y2, x3, y3, x, y)
+	if det > epsilonI {
+		return false
+	}
+	if det < -epsilonI {
+		return true
+	}
+	return robustIncircle(setBig(x1), setBig(y1), setBig(x2), setBig(y2), setBig(x3), setBig(y3), setBig(x), setBig(y))
+}
+
+// simpleIncircle finds the determinant of the matrix using the simple float64 type
+// |1 x1 y1 x1^2+y1^2|
+// |1 x2 y2 x2^2+y2^2|
+// |1 x3 y3 x3^2+y3^2|
+// |1 x  y  x^2 +y^2 |
+func simpleIncircle(x1, y1, x2, y2, x3, y3, x, y float64) (det float64) {
+	m := [][]float64{
+		{1, x1, y1, x1*x1 + y1*y1},
+		{1, x2, y2, x2*x2 + y2*y2},
+		{1, x3, y3, x3*x3 + y3*y3},
+		{1, x, y, x*x + y*y},
+	}
+	return m[0][3]*m[1][2]*m[2][1]*m[3][0] - m[0][2]*m[1][3]*m[2][1]*m[3][0] -
+		m[0][3]*m[1][1]*m[2][2]*m[3][0] + m[0][1]*m[1][3]*m[2][2]*m[3][0] +
+		m[0][2]*m[1][1]*m[2][3]*m[3][0] - m[0][1]*m[1][2]*m[2][3]*m[3][0] -
+		m[0][3]*m[1][2]*m[2][0]*m[3][1] + m[0][2]*m[1][3]*m[2][0]*m[3][1] +
+		m[0][3]*m[1][0]*m[2][2]*m[3][1] - m[0][0]*m[1][3]*m[2][2]*m[3][1] -
+		m[0][2]*m[1][0]*m[2][3]*m[3][1] + m[0][0]*m[1][2]*m[2][3]*m[3][1] +
+		m[0][3]*m[1][1]*m[2][0]*m[3][2] - m[0][1]*m[1][3]*m[2][0]*m[3][2] -
+		m[0][3]*m[1][0]*m[2][1]*m[3][2] + m[0][0]*m[1][3]*m[2][1]*m[3][2] +
+		m[0][1]*m[1][0]*m[2][3]*m[3][2] - m[0][0]*m[1][1]*m[2][3]*m[3][2] -
+		m[0][2]*m[1][1]*m[2][0]*m[3][3] + m[0][1]*m[1][2]*m[2][0]*m[3][3] +
+		m[0][2]*m[1][0]*m[2][1]*m[3][3] - m[0][0]*m[1][2]*m[2][1]*m[3][3] -
+		m[0][1]*m[1][0]*m[2][2]*m[3][3] + m[0][0]*m[1][1]*m[2][2]*m[3][3]
+}
+
+// robustIncircle finds the determinant of the matrix using the accurate big/Rat type
+// |1 x1 y1 x1^2+y1^2|
+// |1 x2 y2 x2^2+y2^2|
+// |1 x3 y3 x3^2+y3^2|
+// |1 x  y  x^2 +y^2 |
+func robustIncircle(x1, y1, x2, y2, x3, y3, x, y *big.Rat) bool {
+	m := [][]*big.Rat{
+		{one, x1, y1, bigAdd(bigMul(x1, x1), bigMul(y1, y1))},
+		{one, x2, y2, bigAdd(bigMul(x2, x2), bigMul(y2, y2))},
+		{one, x3, y3, bigAdd(bigMul(x3, x3), bigMul(y3, y3))},
+		{one, x, y, bigAdd(bigMul(x, x), bigMul(y, y))},
+	}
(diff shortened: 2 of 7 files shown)


Bastiantheone:
My thought on the breakpoint is: 
macheps is the machine epsilon, it is the upper bound on relative error due to rounding
macheps = 2.22e-16
Since we assume the input numbers are correct macheps comes into play after each operation, when the result is rounded. mul(a,b) = a * b +/- macheps, sub(a,b)=a - b +/- macheps and sum(a,b)=a + b +/- macheps. For orientation there are 11 operations:
`det := x1*y2 + x2*y + x*y1 - x1*y - x2*y1 - x*y2`
Therefore epsilon = 11 * macheps

Bastiantheone:
In incircle it would have to be calculated dynamically, because there are chained multiplications:
mul(mul(a,b),c) = mul(a * b + macheps,c) = a * b * c + macheps * c + macheps

Bastiantheone:
Can you guys give this PR a look? @sbinet, @kortschak, @vladimir-ch

sbinet:
yep. just back from holidays and finished handling my stack of emails. going through the PRs now/tomorrow :)

Bastiantheone:
Great :)

fastjet/internal/predicates/incircle.go
```diff
@@ -0,0 +1,114 @@
...
+package predicates
+
+import (
+	"math/big"
+)
+
+const (
+	// macheps is the machine epsilon aka unit roundoff
+	// it is the upper bound on relative error due to rounding
+	macheps = 2.22e-16
```
sbinet: I would perhaps add some reference as to where it is coming from.
https://en.wikipedia.org/wiki/Machine_epsilon ?

and also explain the value (2^-52)

fastjet/internal/predicates/incircle.go
```diff
@@ -0,0 +1,114 @@
...
+	return robustIncircle(setBig(x1), setBig(y1), setBig(x2), setBig(y2), setBig(x3), setBig(y3), setBig(x), setBig(y))
+}
+
+// simpleIncircle finds the determinant of the matrix using the simple float64 type
+// |1 x1 y1 x1^2+y1^2|
+// |1 x2 y2 x2^2+y2^2|
+// |1 x3 y3 x3^2+y3^2|
+// |1 x  y  x^2 +y^2 |
+func simpleIncircle(x1, y1, x2, y2, x3, y3, x, y float64) (det, e float64) {
+	m := [][]float64{
```
sbinet: `simpleIncircle` will supposedly be called in the tight loop.
I am a bit wary about the memory locality of `m` and the bounds checks that the computation of the determinant will incur.

a possible way would be to recast `m` as a slice:

```go
m := []float64{
    1, x1, y1, x1*x1 + y1*y1,
    1, x2, y2, x2*x2 + y2*y2,
    1, x3, y3, x3*x3 + y3*y3,
    1, x, y, x*x + y*y,
}
```

or (leveraging [gonum.org/v1/gonum/mat.Det](https://godoc.org/gonum.org/v1/gonum/mat#Det)):
```go
m := mat.NewDense(4, 4, []float64{
    1, x1, y1, x1*x1 + y1*y1,
    1, x2, y2, x2*x2 + y2*y2,
    1, x3, y3, x3*x3 + y3*y3,
    1, x, y, x*x + y*y,
})
det := mat.Det(m)
```

`gonum/mat` is an additional dependency, but `mat.Det` is presumably more numerically stable (although, for a `4x4` matrix...)

fastjet/internal/predicates/incircle.go
```diff
@@ -0,0 +1,114 @@
...
+	if det > e {
+		return false
+	}
+	if det < -e {
+		return true
+	}
+	return robustIncircle(setBig(x1), setBig(y1), setBig(x2), setBig(y2), setBig(x3), setBig(y3), setBig(x), setBig(y))
+}
+
+// simpleIncircle finds the determinant of the matrix using the simple float64 type
```
sbinet: you might want to also describe what is `e` being returned.

fastjet/internal/predicates/incircle.go
```diff
@@ -0,0 +1,114 @@
...
+	// e is the maximum possible error due to rounding for the result of the determinant.
+	// mul(mul(a,b),c) = mul(a*b+macheps,c) = a*b*c + macheps*c + macheps
+	// sum(mul(a,b),c) = sum(a*b+macheps,c) = a*b+c + macheps + macheps
+	// Therefore, when multiplications are chained, the error also depends on the value
+	// of the number, but this does not apply to sums or subtractions.
+	e = macheps*m[3][0]*6 + macheps*m[3][1]*6 + macheps*m[3][2]*6 + macheps*m[3][3]*6 + macheps*47
+	return det, e
+}
+
+// robustIncircle finds the determinant of the matrix using the accurate big/Rat type
```
sbinet: I would rephrase this a bit as `robustIncircle` doesn't really find the determinant.
it computes the determinant and then returns whether it's negative.

fastjet/internal/predicates/orientation.go
```diff
@@ -0,0 +1,93 @@
...
+	// Counterclockwise
+	CCW OrientationKind = iota
+	// Clockwise
+	CW
+	Colinear
+	Indeterminate
+
+	// epsilon determines when the determinant in orientation is too close to 0 to rely on floating point operations
+	// it is 11 * machine epsilon, because there are 11 operations in orientation, with no chained multiplications
+	epsilon = 2.442e-15
```
sbinet: I would replace it with the formula that you describe in the documentation:
```go
epsilon = 11 * macheps
```

also, I would remove `epsilon` from this `const` block and put it on its own (so documentation for `OrientationKind` isn't polluted with `epsilon`.)

actually, I suppose you could move the declaration+definition of `epsilon` directly inside `simpleOrientation` (as it's only used there.)

fastjet/internal/predicates/orientation.go
```diff
@@ -0,0 +1,93 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package predicates
+
+import (
+	"math/big"
+)
+
+type OrientationKind int
```
sbinet: please document `OrientationKind`.

fastjet/internal/predicates/orientation.go
```diff
@@ -0,0 +1,93 @@
...
+func (o OrientationKind) String() string {
+	switch o {
+	case CCW:
+		return "Counterclockwise"
+	case CW:
+		return "Clockwise"
+	case Colinear:
+		return "Colinear"
+	}
+	return "Indeterminate"
```
sbinet: we should catch "impossible" values.

```go
	switch o {
	case CCW:
		return "Counterclockwise"
	case CW:
		return "Clockwise"
	case Colinear:
		return "Colinear"
	case Indeterminate:
		return "Indeterminate"
	default:
		panic(fmt.Errorf("predicates: unknown OrientationKind %d", int(o)))
	}
	panic("unreachable")
```

fastjet/internal/predicates/orientation.go
```diff
@@ -0,0 +1,93 @@
...
+	if o == Indeterminate {
+		// too close to 0 to give a definite answer.
+		// Therefore check with more expansive tests.
+		o = robustOrientation(setBig(x1), setBig(y1), setBig(x2), setBig(y2), setBig(x), setBig(y))
+	}
+	return o
+}
+
+// simpleOrientation finds the orientation using the float64 type.
+// It's accuracy can't be guaranteed, therefore close decisions
```
sbinet: `s/It's/Its/`

fastjet/internal/predicates/orientation.go
```diff
@@ -0,0 +1,93 @@
...
+// testing is necessary
+func simpleOrientation(x1, y1, x2, y2, x, y float64) OrientationKind {
+	if (x1 == x2 && x2 == x) || (y1 == y2 && y2 == y) {
+		// points are horizontally or vertically aligned
+		return Colinear
+	}
+	// Compute the determinant of the matrix
+	//  | x1 y1 1 |
+	//  | x2 y2 1 |
+	//  | x3 y3 1 |
```
sbinet: `s/x3/x/`
`s/y3/y/`

fastjet/internal/predicates/orientation.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	return Indeterminate
+}
+
+// robustOrientation finds the orientation using the accurate big/Rat type.
+func robustOrientation(x1, y1, x2, y2, x, y *big.Rat) OrientationKind {
+	// Compute the determinant of the matrix
+	//  | x1 y1 1 |
+	//  | x2 y2 1 |
+	//  | x3 y3 1 |
```
sbinet: `s/3//g`

sbinet:
it would be great to also add a few tests that show the differences b/w the computation of the determinant w/ `float64` and w/ `big.Rat`.

fastjet/internal/predicates/incircle.go
Bastiantheone: Done

fastjet/internal/predicates/incircle.go
Bastiantheone: Done

fastjet/internal/predicates/incircle.go
Bastiantheone: Done

fastjet/internal/predicates/orientation.go
Bastiantheone: Done

fastjet/internal/predicates/orientation.go
Bastiantheone: Done

fastjet/internal/predicates/orientation.go
Bastiantheone: Done

fastjet/internal/predicates/orientation.go
Bastiantheone: Done

fastjet/internal/predicates/orientation.go
Bastiantheone: Done

fastjet/internal/predicates/orientation.go
Bastiantheone: Done

fastjet/internal/predicates/incircle.go
Bastiantheone: I'm going to go with recasting m, because it seems very complicated to calculate the max error for mat.Det

sbinet: ah, right.

I'd be interested in having a set of tests that compares the values for all these approaches.
(well, at least a few test matrices you happen to stumble upon that have interesting determinant values. then I can do the comparison myself :))

Bastiantheone:
The error value calculation I assumed is not correct. There was a flaw in my logic:
```
func TestSimpleVsRobustOrientation(t *testing.T){
	tests := []struct {
		x1, y1, x2, y2, x, y float64
	}{
		{2.1, 2.1, 1.1, 1.1, 0.1, 0.1}, // works as expected
		{2.1, 2.1, 1.1, 1.1, 100.1, 100.1}, // simple should have been Indeterminate
	}
	for _, test := range tests {
		simpleOr := simpleOrientation(test.x1,test.y1,test.x2,test.y2,test.x,test.y)
		robustOr := robustOrientation(setBig(test.x1),setBig(test.y1),setBig(test.x2),setBig(test.y2),setBig(test.x),setBig(test.y))
		fmt.Printf("x1 = %v, y1 = %v, x2 = %v, y2 = %v, x = %v, y = %v, Simple = %v, Robust = %v \n",test.x1, test.y1, test.x2, test.y2, test.x, test.y, simpleOr,robustOr)
	}
}
```
The higher a number gets the bigger the steps between neighboring floating numbers and therefore bigger error margins. I would have to examine each intermediate result to determine an accurate error margin. I think macheps^exponent of float would be what I have to do. I'm not sure how I would be able to get the exponent of a float. The other options I can see are using robust for every case (big.Float might be good enough), or a relatively high constant chosen to be higher than most or all possible error margins, or instead of macheps I use macheps^maxexponent. I incline to use the option I suggested last. Please let me know if any of this sounds wrong, or if you think another option is better @sbinet

sbinet:
I have tested with your set of points:

```go
func TestSimpleVsRobustOrientation(t *testing.T) {
	tests := []struct {
		x1, y1, x2, y2, x, y float64
		simple               OrientationKind
		robust               OrientationKind
	}{
		{2.1, 2.1, 1.1, 1.1, 0.1, 0.1, Indeterminate, Colinear},
		{2.1, 2.1, 1.1, 1.1, 100.1, 100.1, Indeterminate, Colinear},
	}
	for _, test := range tests {
		simpleOr := simpleOrientation(test.x1, test.y1, test.x2, test.y2, test.x, test.y)
		robustOr := robustOrientation(setBig(test.x1), setBig(test.y1), setBig(test.x2), setBig(test.y2), setBig(test.x), setBig(test.y))
		if simpleOr != test.simple {
			t.Errorf("x1 = %v, y1 = %v, x2 = %v, y2 = %v, x = %v, y = %v, want.Simple = %v. got= %v\n", test.x1, test.y1, test.x2, test.y2, test.x, test.y, simpleOr, test.simple)
		}
		if robustOr != test.robust {
			t.Errorf("x1 = %v, y1 = %v, x2 = %v, y2 = %v, x = %v, y = %v, want.Robust = %v. got= %v\n", test.x1, test.y1, test.x2, test.y2, test.x, test.y, robustOr, test.robust)
		}
		o := matOrientation(test.x1, test.y1, test.x2, test.y2, test.x, test.y)
		if o != test.robust {
			t.Errorf("x1 = %v, y1 = %v, x2 = %v, y2 = %v, x = %v, y = %v, want.Orientation = %v. got= %v\n", test.x1, test.y1, test.x2, test.y2, test.x, test.y, o, test.robust)
		}
	}
}

func matOrientation(x1, y1, x2, y2, x, y float64) OrientationKind {
	if (x1 == x2 && x2 == x) || (y1 == y2 && y2 == y) {
		// points are horizontally or vertically aligned
		return Colinear
	}
	m := mat.NewDense(3, 3, []float64{x1, y1, 1, x2, y2, 1, x, y, 1})
	det := mat.Det(m)
	switch {
	case det > epsilon:
		return CCW
	case det < -epsilon:
		return CW
	case det == 0 || det == -0:
		return Colinear
	}
	return Indeterminate
}
```

it would seem to work.
at least with this set of points.

Bastiantheone:
You have convinced me to use the mat package :) 

I found out that the [geo](https://github.com/golang/geo/blob/master/s2/predicates.go) package has functions that handle predicates. The orientation test is build up similar to the one I build up. Epsilon is calculated the following way:

`detErrorMultiplier = 3.2321 * dblEpsilon`
...
`// The two shortest edges, pointing away from their common point.`
`var e1, e2, op r3.Vector`
...
`maxErr := detErrorMultiplier * math.Sqrt(e1.Norm2()*e2.Norm2())`

It looks to me like this works in general for all dimensions.

Are you aware of something similar for Incirlce, since it has 4 points?

fastjet/internal/predicates/incircle.go
kortschak: In lapack, we express the equivalent as `const dlamchE = 1.0 / (1 << 53)`, here `const macheps = 1.0 / (1 << 52)`.

Bastiantheone:
Orientation shows the same pattern:
```
3000000	       447 ns/op
50000	     32341 ns/op
10000	    125710 ns/op
```

By the way, the test are run on multiple constellation for each operation

Bastiantheone:
I believe the error calculation is correct now. The type float64Pred calculates the error when doing normal float operations. The comment in float64Pred type declaration describes my train of thought.

fastjet/internal/predicates/float64Pred.go
```diff
@@ -0,0 +1,114 @@
...
+	"math"
+)
+
+const (
+	// macheps is the machine epsilon aka unit roundoff
+	// The machine epsilon is an upper bound on the absolute relative true error in
+	// representing a number.
+	// If y is the machine representation of x then |(x-y)/x| <= macheps
+	// https://en.wikipedia.org/wiki/Machine_epsilon
+	// Golangs float64 type has a 52-bit fractional mantissa,
```
sbinet: `s/Golangs/Go's/`

fastjet/internal/predicates/float64Pred.go
```diff
@@ -0,0 +1,114 @@
...
+	// therefore the value 2^-52
+	macheps = 1.0 / (1 << 52)
+)
+
+// float64Pred dynamically updates the potential error
```
sbinet: add a dot at the end of the sentence.

fastjet/internal/predicates/incircle.go
```diff
@@ -5,110 +5,196 @@
...
-	// macheps is the machine epsilon aka unit roundoff
-	// it is the upper bound on relative error due to rounding
-	macheps = 2.22e-16
+	// Inside the circle
+	Inside RelativePosition = iota
+	// On the circle
+	On
+	// Outside the circle
+	Outside
+	IndeterminateP
```
sbinet: `IndeterminatePosition` ?

fastjet/internal/predicates/incircle.go
```diff
@@ -5,110 +5,196 @@
...
 	}
 	return bigSub(
-		bigMul(bigMul(m[0][a], m[1][b]), bigMul(m[2][c], m[3][d])),
-		bigMul(bigMul(m[0][b], m[1][a]), bigMul(m[2][c], m[3][d])),
+		bigMul(bigMul(m[a], m[b+4]), bigMul(m[c+8], m[d+12])),
+		bigMul(bigMul(m[b], m[a+4]), bigMul(m[c+8], m[d+12])),
 	)
 }
+
+// matIncircle computes the determinant of the matrix using the mat package.
```
sbinet: `matIncircle` only seems to be used for the tests/benchmarks.
put it in that file instead of there.

fastjet/internal/predicates/orientation.go
```diff
@@ -16,11 +20,7 @@ const (
 	// Clockwise
 	CW
 	Colinear
-	Indeterminate
-
-	// epsilon determines when the determinant in orientation is too close to 0 to rely on floating point operations
-	// it is 11 * machine epsilon, because there are 11 operations in orientation, with no chained multiplications
-	epsilon = 2.442e-15
+	IndeterminateO
```
sbinet: `IndeterminateOrientation` ?

fastjet/internal/predicates/orientation_test.go
```diff
@@ -30,3 +32,101 @@ func TestOrientation(t *testing.T) {
...
+		o := simpleOrientation(test.x1, test.y1, test.x2, test.y2, test.x, test.y)
+		if o != test.simple {
+			t.Errorf("x1 = %v, y1 = %v, x2 = %v, y2 = %v, x = %v, y = %v, want.Simple = %v. got= %v\n", test.x1, test.y1, test.x2, test.y2, test.x, test.y, test.simple, o)
+		}
+		o = robustOrientation(setBig(test.x1), setBig(test.y1), setBig(test.x2), setBig(test.y2), setBig(test.x), setBig(test.y))
+		if o != test.robust {
+			t.Errorf("x1 = %v, y1 = %v, x2 = %v, y2 = %v, x = %v, y = %v, want.Robust = %v. got= %v\n", test.x1, test.y1, test.x2, test.y2, test.x, test.y, test.robust, o)
+		}
+		o = matOrientation(test.x1, test.y1, test.x2, test.y2, test.x, test.y)
+		if o != test.simple {
```
sbinet: shouldn't you test against `test.robust` instead?
I'd expect the `gonum/mat` routines to be robust :)
(c.f. my comment above)

fastjet/internal/predicates/orientation.go
```diff
@@ -91,3 +104,26 @@ func robustOrientation(x1, y1, x2, y2, x, y *big.Rat) OrientationKind {
...
+	if (x1 == x2 && x2 == x) || (y1 == y2 && y2 == y) {
+		// points are horizontally or vertically aligned
+		return Colinear
+	}
+	m := mat.NewDense(3, 3, []float64{x1, y1, 1, x2, y2, 1, x, y, 1})
+	logDet, sign := mat.LogDet(m)
+	// FIXME determine maxErrorLog
+	// maxErrorLog is the highest possible number that could lead to a wrong sign due to rounding error
+	maxErrorLog := -10.0
+	if logDet < maxErrorLog {
```
sbinet: I am not quite sure of the logic here.
`LogDet` should return `math.Inf(-1)` if the determinant is 0.

if we've convinced ourselves that `gonum/mat` is robust enough, then we should just test whether `math.IsInf(logDet,-1)`, return `Colinear` and be done with it.

I guess what I am aiming at is to find edge cases where `matOrientation` (written as I suggest above) gives the wrong answer compared to the version using `math/big`)

Bastiantheone: That's what I had originally. I'm sorry, I forgot to mention that I found a couple cases, all with higher numbers, where mat returns the wrong result. For example for the test
`{1000,2000, 2000, 3000, 10000, 11000, IndeterminateO, Colinear},`
mat returns Counterclockwise.

fastjet/internal/predicates/float64Pred.go
Bastiantheone: Done

fastjet/internal/predicates/float64Pred.go
Bastiantheone: Done

fastjet/internal/predicates/incircle.go
Bastiantheone: Done

fastjet/internal/predicates/orientation.go
Bastiantheone: Done

fastjet/internal/predicates/orientation.go
Bastiantheone: Should I remove the mat functions, or make them public so that the user can decide what to use?

fastjet/internal/predicates/orientation.go
```diff
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"math/big"
 
+	"math"
```
sbinet: remove empty line 10. `"math"` is a stdlib package, it should be part of the stdlib import block.

Bastiantheone: Done and it is fixed for incircle too

fastjet/internal/predicates/orientation.go
sbinet: just put them in the `_test.go` files for the moment.
and keep them un-exported.

sbinet:
I was curious to see how `numpy` was doing with your set of points:

```python
>>> import numpy as np
>>> np.linalg.det(np.array([1,2,1,2,3,1,10,11,1],dtype="float64").reshape(3,3))
1.9984014443252916e-15
>>> np.linalg.det(np.array([2.1,2.1,1,1.1,1.1,1,0.1,0.1,1]).reshape(3,3))
0.0
```

so it has the same issue (not surprising as the underlying algorithm to compute the determinant is the same: lapack-based)

then I tried [ROOT](https://root.cern.ch) and its `TMatrixD` class:

```
$> root.exe
root [0] std::vector<double> data;
root [1] data = std::vector<double>{1000,2000,1,2000,3000,1,10000,11000,1};
root [2] TMatrixD(3,3,data.data()).Print()

3x3 matrix is as follows

     |      0    |      1    |      2    |
--------------------------------------------
   0 |       1000        2000           1 
   1 |       2000        3000           1 
   2 |      1e+04     1.1e+04           1 

root [3] TMatrixD(3,3,data.data()).Determinant()
(double) 0.000000
root [4] data = std::vector<double>{1,2,1,2,3,1,10,11,1};
root [5] TMatrixD(3,3,data.data()).Print()

3x3 matrix is as follows

     |      0    |      1    |      2    |
--------------------------------------------
   0 |          1           2           1 
   1 |          2           3           1 
   2 |         10          11           1 

root [6] TMatrixD(3,3,data.data()).Determinant()
(double) 0.000000
root [7] data = std::vector<double>{2.1, 2.1, 1.1, 1.1, 1000.1, 1000.1};
root [8] TMatrixD(3,3,data.data()).Print()

3x3 matrix is as follows

     |      0    |      1    |      2    |
--------------------------------------------
   0 |        2.1         2.1         1.1 
   1 |        1.1        1000        1000 
   2 | 1.019e-316  7.164e-322  2.019e-316 

root [9] TMatrixD(3,3,data.data()).Determinant()
(double) -0.000000
```

it seems ROOT uses a different technique to decompose and extract the determinant, which is "Based on Bowler, Martin, Peters and Wilkinson in HACLA"

@btracey do you have any insights about this reference? (my google-fu failed me.)
Here is the code (which is LGPL2, in (lightly templatized) C++):
 - https://github.com/root-project/root/blob/master/math/matrix/src/TMatrixT.cxx#L1357
 - https://github.com/root-project/root/blob/master/math/matrix/src/TDecompLU.cxx#L506
 - https://github.com/root-project/root/blob/master/math/matrix/src/TDecompBase.cxx#L297
 - https://github.com/root-project/root/blob/master/math/matrix/src/TDecompBase.cxx#L209

btracey:
If I understand the issue, it’s that one algorithm outputs literally 0 for the determinant, and the other outputs a very small number. The determinant is an inherently unstable calculation. For this reason, Lapack does not provide any routines for computing the determinant (http://www.netlib.org/lapack/faq.html#_are_there_routines_in_lapack_to_compute_determinants <http://www.netlib.org/lapack/faq.html#_are_there_routines_in_lapack_to_compute_determinants>). It can also be that if the matrix is near-singular, you can get an extremely large determinant rather than a very small one. Somewhere in the gonum/lapack commit history is me running into this issue while trying to write a test. Long story short, if the matrix is (near-)singular, you cannot trust the computation of the determinant (or basically anything else for that matter).

A much better test is to check the condition number (in gonum, mat.Cond). For the first matrix, you’ll see that gonum returns the condition number is 2.1617278211378387e+17. Loosely, you expect to lose log10(condition number) digits of accuracy in computations, and 17 digits is a lot when we only expect 16 digits of precision. Sometimes you’ll get small values of the determinant, but small perturbations can throw it off by a lot.

Our threshold in gonum for signaling singularity is 1e16  (the constant ConditionTolerance). That’s how I think you should test singularity, not looking at the gradient.

sbinet:
ok. so, if I understood correctly, for our use-case (determining whether points are colinear or not), it boils down to:

```go
func matDet(x1, y1, x2, y2, x, y float64) float64 {
   m := mat.NewDense(3,3, []float64{x1, y1, 1, x2, y2, 1, x, y, 1})
   cond := mat.Cond(m, 1)
   det := 0.0
   if cond < mat.ConditionTolerance {
	det = mat.Det(m)
   }
   return det
}
```

sbinet:
interestingly, with this implementation, I get these numbers:

```
BenchmarkSimpleOrientation-4   	 1000000	      1837 ns/op
BenchmarkMatOrientation-4      	   10000	    102737 ns/op
BenchmarkRobustOrientation-4   	    5000	    238879 ns/op
PASS
```

that compare the performances of:
- simple: a "naive" determinant computation
- mat: using `matDet` above (and thus `gonum/mat`)
- robust: using the same "naive" determinant computation but with `math/big` instead of `float64`.

I wasn't expecting such a difference between `simple` and `gonum/mat` (and `gonum/mat` is "only" ~2 times faster than `math/big`...)
I guess I am showing my lack of proper education in numerical recipes :)

btracey:
If you’re doing the Cond in the loop, and not just in a test, you can save a bunch by computing the factorization yourself.

var lu mat.LU
lu.Factorize(m)
cond := lu.Cond()
det := lu.Det()

Or, at least you should be able to. Apparently Cond() isn’t exported for LU like it is for SVD. We should fix that.

For 3x3 matrices the determinant formula reduces to coming simple, and so it’s not surprising to me the speedup is big. LU has to allocate, compute some stuff, compute the condition number, check sizes, all that jazz. Simple doesn’t need to do any of that. You’ll double or triple the simple implementation cost if you also compute the condition number. I don’t know if we’ve gone through and really tried to optimize LU though, so I wouldn’t be surprised if there are gains to be had.

Bastiantheone:
The function Sebastien suggested does still have a significant improvement for Incircle, where we evaluate a 4x4 matrix. So by computing the factorization ourselves the improvement should be even bigger.
```
BenchmarkSimpleIncircle    200000	      6689 ns/op
BenchmarkMatIncircle        20000	     53171 ns/op
BenchmarkRobustIncircle    2000	            651133 ns/op
```

Using the condition number gives correct results for the tests I have, but it is still an approximation. If the matrix is not singular, we also need to determine the sign of the determinant. Is it accurate enough to be used as the robust method?

sbinet:
I think I've convinced myself the robustness we get with `gonum/mat` is good enough for our use case (jets positions in a cylinder).

it's only with:
```go
{1 * fac, 2 * fac, 2 * fac, 3 * fac, 10 * fac, 11*fac + res, IndeterminateOrientation, Colinear},
```

where:
```go
const fac = 1e7
const res = 1
```
or:
```go
const fac = 1
const res = 1e-14
```

that `gonum/mat` doesn't give the same answer than `math/big`.

I can't come up with a situation where we'd have jets that far away.

sbinet:
I think I will accept this PR as is, so, with `math/big` and all the documentation work you poured in.
then in a later PR, move to `gonum/mat` where it makes sense (so, if we later on discover we *really* do need `math/big` it will be relatively simple to bring it back.)

what do you people think?

Bastiantheone:
Should I go ahead and squash the commits?

sbinet:
yes. thanks.

Bastiantheone:
I accidentally deleted the branch earlier, because of a problem with the rebasing and now after force pushing it I can't reopen the issue :( @sbinet

sbinet:
hum. I don't seem to be able to reopen it either...

sbinet:
I have created another PR: go-hep/hep#76.

Bastiantheone:
Thank you! Sorry about that

sbinet:
no worries.
there are 2 kind of people: the ones who screwed up PR, and the ones who haven't yet :)

PR #77, opened 2017-08-07
fastjet/internal/delaunay: new package to implement delaunay triangulation

Bastiantheone:
This PR has the data structures and the frame work for the delaunay implementation

### DIFF
--- a/fastjet/internal/delaunay/point.go
+++ b/fastjet/internal/delaunay/point.go
@@ -1,0 +1,149 @@
+// Copyright 2017 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package delaunay
+
+import (
+	"fmt"
+	"math"
+
+	"go-hep.org/x/hep/fastjet/internal/predicates"
+)
+
+// Point in the X-Y Plane.
+//
+// It holds dynamic information about the
+// adjacent triangles, the nearest neighbor and the distance to that neighbor.
+type Point struct {
+	x, y              float64   // x and y are the coordinates of the point.
+	adjacentTriangles triangles // adjacentTriangles is a list of triangles containing the point.
+	nearest           *Point
+	dist2             float64 // dist2 is the squared distance to the nearest neighbor.
+}
+
+// NewPoint returns Point for the given x,y coordinates
+func NewPoint(x, y float64) *Point {
+	return &Point{
+		x:     x,
+		y:     y,
+		dist2: math.Inf(1),
+	}
+}
+
+// NearestNeighbor returns the nearest neighbor and the distance to that neighbor.
+func (p *Point) NearestNeighbor() (*Point, float64) {
+	return p.nearest, math.Sqrt(p.dist2)
+}
+
+// Coordinates returns the x,y coordinates of a Point.
+func (p *Point) Coordinates() (x, y float64) {
+	return p.x, p.y
+}
+
+func (p *Point) String() string {
+	return fmt.Sprintf("(%f, %f)", p.x, p.y)
+}
+
+// Equals checks whether two points are the same.
+func (p *Point) Equals(v *Point) bool {
+	return p == v || (p.x == v.x && p.y == v.y)
+}
+
+// distance returns the squared distance between two points.
+func (p *Point) distance(v *Point) float64 {
+	dx := p.x - v.x
+	dy := p.y - v.y
+	return dx*dx + dy*dy
+}
+
+// findNearest looks at all adjacent points of p and finds the nearest one.
+// p's nearest neighbor will be updated.
+func (p *Point) findNearest() {
+	var newNearest *Point
+	min := math.Inf(1)
+	for _, t := range p.adjacentTriangles {
+		var dist float64
+		var np *Point
+		// find the point in t that is closest to p, but that is not p.
+		switch {
+		case p.Equals(t.A):
+			distB := p.distance(t.B)
+			distC := p.distance(t.C)
+			if distB <= distC {
+				dist = distB
+				np = t.B
+			} else {
+				dist = distC
+				np = t.C
+			}
+		case p.Equals(t.B):
+			distA := p.distance(t.A)
+			distC := p.distance(t.C)
+			if distA <= distC {
+				dist = distA
+				np = t.A
+			} else {
+				dist = distC
+				np = t.C
+			}
+		case p.Equals(t.C):
+			distA := p.distance(t.A)
+			distC := p.distance(t.C)
+			if distA <= distC {
+				dist = distA
+				np = t.A
+			} else {
+				dist = distC
+				np = t.C
+			}
+		default:
+			panic(fmt.Errorf("delaunay: point P%s not found in T%s", p, t))
+		}
+		// check whether the distance found is smaller than the previous smallest distance.
+		if dist < min {
+			min = dist
+			newNearest = np
+		}
+	}
+	// update p's nearest Neighbor
+	p.dist2 = min
+	p.nearest = newNearest
+}
+
+// inTriangle checks whether the point is in the triangle and whether it is on an edge.
+func (p *Point) inTriangle(t *Triangle) location {
+	o1 := predicates.Orientation(t.A.x, t.A.y, t.B.x, t.B.y, p.x, p.y)
+	o2 := predicates.Orientation(t.B.x, t.B.y, t.C.x, t.C.y, p.x, p.y)
+	o3 := predicates.Orientation(t.C.x, t.C.y, t.A.x, t.A.y, p.x, p.y)
+	if o1 == predicates.CCW && o2 == predicates.CCW && o3 == predicates.CCW {
+		return inside
+	}
+	if o1 == predicates.CW || o2 == predicates.CW || o3 == predicates.CW {
+		return outside
+	}
+	return onEdge
+}
+
+// location is the position of a point relative to a triangle
+type location int
+
+const (
+	inside location = iota
+	onEdge
+	outside
+)
+
+func (l location) String() string {
+	switch l {
+	case inside:
+		return "Inside Triangle"
+	case onEdge:
+		return "On Edge of Triangle"
+	case outside:
+		return "Outside Triangle"
+	default:
+		panic(fmt.Errorf("delaunay: unknown location %d", int(l)))
+	}
+	panic("unreachable")
+}
--- a/fastjet/internal/delaunay/point_test.go
+++ b/fastjet/internal/delaunay/point_test.go
@@ -1,0 +1,100 @@
+// Copyright 2017 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package delaunay
+
+import (
+	"testing"
+
+	"gonum.org/v1/gonum/floats"
+)
+
+func TestPointEquals(t *testing.T) {
+	tests := []struct {
+		x1, y1, x2, y2 float64
+		want           bool
+	}{
+		{5, 5, 5, 5, true},
+		{0.5, 4, 1, 4, false},
+		{-3, 20, 20, -3, false},
+		{2.5, 4, 2.5, 4, true},
+	}
+	for _, test := range tests {
+		p1 := NewPoint(test.x1, test.y1)
+		p2 := NewPoint(test.x2, test.y2)
+		got := p1.Equals(p2)
+		if got != test.want {
+			t.Errorf("%v == %v, got = %v, want = %v", p1, p2, got, test.want)
+		}
+	}
+}
+
+func TestDistance(t *testing.T) {
+	tests := []struct {
+		x1, y1, x2, y2 float64
+		want           float64
+	}{
+		{1, 1, 3, 3, 8},
+		{0, 0, 0, 5, 25},
+	}
+	for _, test := range tests {
+		p1 := NewPoint(test.x1, test.y1)
+		p2 := NewPoint(test.x2, test.y2)
+		got := p1.distance(p2)
+		if got != test.want {
+			t.Errorf("Distance between %v and %v, got = %v, want = %v", p1, p2, got, test.want)
+		}
+	}
+}
+
+func TestFindNearest(t *testing.T) {
+	const tol float64 = 0.001
+	tests := []struct {
+		p                 *Point
+		adjacentTriangles []*Triangle
+		wantDist          float64
+		wantNeighbor      *Point
+	}{
+		{
+			NewPoint(0, 0),
+			[]*Triangle{
+				NewTriangle(NewPoint(0, 0), NewPoint(2, 0), NewPoint(0, 5)),
+				NewTriangle(NewPoint(3, 3), NewPoint(0, 0), NewPoint(0, 5)),
+			},
+			2,
+			NewPoint(2, 0),
+		},
+	}
+	for _, test := range tests {
+		test.p.adjacentTriangles = test.adjacentTriangles
+		test.p.findNearest()
+		gotNeighbor, gotDistance := test.p.NearestNeighbor()
+		if !floats.EqualWithinAbs(gotDistance, test.wantDist, tol) || !test.wantNeighbor.Equals(gotNeighbor) {
+			t.Errorf("Nearest Neighbor for P%v with adjacent triangles %v, \n gotDist = %v, wantDist = %v, gotNeighbor = %v, wantNeighbor = %v",
+				test.p, test.adjacentTriangles, gotDistance, test.wantDist, gotNeighbor, test.wantNeighbor)
+		}
+	}
+}
+
+func TestInTriangle(t *testing.T) {
+	tests := []struct {
+		x1, y1, x2, y2, x3, y3, x, y float64
+		want                         location
+	}{
+		{2, 2, 5, 3, 6, 2, 5, 2.5, inside},
+		{2, 2, 5, 3, 6, 2, 5, 2, onEdge},
+		{2, 2, 5, 3, 6, 2, 4, 3, outside},
+		{2, 2, 5, 3, 6, 2, 10, 2, outside},
+		{5, 3, 2, 2, 6, 1.5, 5, 2.5, inside},
+		{5, 3, 2, 2, 6, 1.5, 6, 2.5, outside},
+	}
+	for _, test := range tests {
+		tri := NewTriangle(NewPoint(test.x1, test.y1), NewPoint(test.x2, test.y2), NewPoint(test.x3, test.y3))
+		p := NewPoint(test.x, test.y)
+		got := p.inTriangle(tri)
+		if got != test.want {
+			t.Fatalf("Point %v in Triangle %v, got = %v, want = %v", p, tri, got, test.want)
+		}
+	}
+}
--- a/fastjet/internal/delaunay/triangle.go
+++ b/fastjet/internal/delaunay/triangle.go
@@ -1,0 +1,161 @@
+// Copyright 2017 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package delaunay
+
+import (
+	"fmt"
+
+	"go-hep.org/x/hep/fastjet/internal/predicates"
+)
+
+// Triangle is a set of three points that make up a triangle. It stores hierarchical information to find triangles.
+type Triangle struct {
+	// children are triangles that lead to the removal of this triangle.
+	// When a point is inserted the triangle that contains the point is found by going down the hierarchical tree.
+	// The tree's root is the root triangle and the children slice contains the children triangles.
+	children []*Triangle
+	// A,B,C are the CCW-oriented points that make up the triangle
+	A, B, C *Point
+	// isInTriangulation holds whether a triangle is part of the triangulation
+	isInTriangulation bool
+}
+
+// NewTriangle returns a triangle formed out of the three given points.
+func NewTriangle(a, b, c *Point) *Triangle {
+	// order the points counter clockwise
+	o := predicates.Orientation(a.x, a.y, b.x, b.y, c.x, c.y)
+	switch o {
+	case predicates.CW:
+		a, b = b, a
+	case predicates.Colinear:
+		panic(fmt.Errorf("delaunay: Can't form triangle, because Points a%v, b%v and c%v are colinear.", a, b, c))
+	}
+	return &Triangle{
+		A: a,
+		B: b,
+		C: c,
+	}
+}
+
+// add is used when the triangle t is added to the delaunay triangulation.
+// It updates the information of the points in t.
+// It returns all points whose nearest neighbor was updated.
+func (t *Triangle) add() []*Point {
+	t.isInTriangulation = true
+	// update the adjacent lists
+	t.A.adjacentTriangles = append(t.A.adjacentTriangles, t)
+	t.B.adjacentTriangles = append(t.B.adjacentTriangles, t)
+	t.C.adjacentTriangles = append(t.C.adjacentTriangles, t)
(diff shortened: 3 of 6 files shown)


fastjet/internal/delaunay/triangle.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+type triangles []*Triangle
+
+// remove removes given triangles from a slice of triangles
+func (t triangles) remove(triangles ...*Triangle) triangles {
+	for i := len(t) - 1; i >= 0; i-- {
+		for j, triangleToRemove := range triangles {
+			if triangleToRemove.Equals(t[i]) {
+				t = append(t[:i], t[i+1:]...)
```
sbinet: do we need to keep the order of the already iterated elements of `t` ?
otherwise we could just use:
```go
n := len(t)-1
t[i], t[n] = t[n], nil
t = t[:n]
```

fastjet/internal/delaunay/triangle.go
```diff
@@ -0,0 +1,147 @@
...
+				triangles = append(triangles[:j], triangles[j+1:]...)
```
sbinet: same remark here *wrt* the order of the already iterated elements of `triangles`.
(but one would need to loop over `triangles` from the end.)

also: what if there is twice the same triangle in the receiver `t triangles` ?
is there some mechanism when creating a list of triangles to ensure it's a unique list of triangles' values ?
if we just assume there should be no duplicate triangle values in `t triangles`, we should document it.

fastjet/internal/delaunay/triangle_test.go
```diff
@@ -0,0 +1,103 @@
...
+		triangles triangles
+		toRemove  []*Triangle
+	}{
+		{
+			triangles{NewTriangle(NewPoint(0, 0), NewPoint(0, 1), NewPoint(1, 0)), NewTriangle(NewPoint(0, 0), NewPoint(5, 1), NewPoint(1, 0)), NewTriangle(NewPoint(0, 5), NewPoint(0, 1), NewPoint(1, 0))},
+			[]*Triangle{NewTriangle(NewPoint(0, 0), NewPoint(0, 1), NewPoint(1, 0)), NewTriangle(NewPoint(0, 5), NewPoint(0, 1), NewPoint(1, 0))},
+		},
+	}
+	for _, test := range tests {
+		triangles := test.triangles.remove(test.toRemove[:]...)
```
sbinet: this is a bit of a tricky and finicky line.
you rely on the fact that a new slice will be created when using the ellipsis operator.

it would be much clearer (and robust) if you rewrite it like so:

```go
toRemove := make([]*Triangle, len(test.toRemove))
copy(toRemove, test.toRemove)
triangles := test.triangles.remove(toRemove...)
for _, tri := range triangles {
    for _, removed := range test.toRemove {
      // ...
```

fastjet/internal/delaunay/triangle_test.go
```diff
@@ -0,0 +1,103 @@
...
+	want := []*Point{
+		points[2], points[2], points[0], points[1], points[0],
+	}
+	var got []*Point
+	for _, p := range points {
+		got = append(got, p.nearest)
+	}
+	for i := range got {
+		if !got[i].Equals(want[i]) {
+			t.Fatalf("After adding triangle nearest neighbor of %v, got = %v, want = %v", points[i], got[i], want[i])
```
sbinet: please add the index `i` value somewhere in the printout.

fastjet/internal/delaunay/triangle_test.go
```diff
@@ -0,0 +1,103 @@
...
+	want := []*Point{
+		points[1], points[0], nil, points[1],
+	}
+	var got []*Point
+	for _, p := range points {
+		got = append(got, p.nearest)
+	}
+	for i := range got {
+		if !got[i].Equals(want[i]) {
+			t.Fatalf("After removing triangle nearest neighbor of %v, got = %v, want = %v", points[i], got[i], want[i])
```
sbinet: please add the index `i` value somewhere in the printout

fastjet/internal/delaunay/triangle_test.go
```diff
@@ -0,0 +1,103 @@
...
+	}{
+		{NewTriangle(NewPoint(3, 2), NewPoint(0, 0), NewPoint(5, 0)), NewTriangle(NewPoint(3, 2), NewPoint(0, 0), NewPoint(5, 0)), true},
+		{NewTriangle(NewPoint(3, 2), NewPoint(0, 0), NewPoint(5, 0)), NewTriangle(NewPoint(5, 0), NewPoint(3, 2), NewPoint(0, 0)), true},
+		{NewTriangle(NewPoint(3, 2), NewPoint(0, 0), NewPoint(5, 0)), NewTriangle(NewPoint(3, 1), NewPoint(0, 0), NewPoint(5, 0)), false},
+		{NewTriangle(NewPoint(3, 2), NewPoint(0, 0), NewPoint(5, 0)), NewTriangle(NewPoint(5, 0), NewPoint(3, 1), NewPoint(0, 0)), false},
+	}
+	for _, test := range tests {
+		got := test.a.Equals(test.b)
+		if got != test.want {
+			t.Fatalf("%v == %v,\n got = %v, want = %v", test.a, test.b, got, test.want)
```
sbinet: use `t.Errorf` instead of `t.Fatalf` so the other test cases can run.
also: please add the index `i` of the test-case being run in the printout.

fastjet/internal/delaunay/point.go
```diff
@@ -0,0 +1,134 @@
...
+//
+// It holds dynamic information about the
+// adjacent triangles, the nearest neighbor and the distance to that neighbor.
+type Point struct {
+	// x and y are the coordinates of the point.
+	x, y float64
+	// adjacentTriangles is a list of triangles containing the point.
+	adjacentTriangles triangles
+	// id is used when points are removed. Copies of the points around the point to
+	// be removed are made. The id is set incremental in counterclockwise order. It identifies
```
sbinet: `s/incremental/incrementally/`

fastjet/internal/delaunay/point.go
```diff
@@ -0,0 +1,134 @@
...
+	// the original the point. The id is then used to determine if a Triangle is inside or outside the
```
sbinet: `the original the point`
there seems to be missing something there (or there's one too many `the`.)

sbinet: `s/determine if/determine whether/`

fastjet/internal/delaunay/point.go
```diff
@@ -0,0 +1,134 @@
...
+	// polygon formed by all those points.
+	id      int
```
sbinet: AFAICT `id` isn't used yet: I would just remove it and introduce it only in a later PR.

sbinet:
@kortschak, @vladimir-ch: do you want to give it a stab?

fastjet/internal/delaunay/point.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"go-hep.org/x/hep/fastjet/internal/predicates"
+)
+
+// Point in the X-Y Plane.
+//
+// It holds dynamic information about the
+// adjacent triangles, the nearest neighbor and the distance to that neighbor.
+type Point struct {
+	// x and y are the coordinates of the point.
```
Bastiantheone: Done

fastjet/internal/delaunay/point.go
```diff
@@ -0,0 +1,134 @@
...
+	x, y float64
+	// adjacentTriangles is a list of triangles containing the point.
+	adjacentTriangles triangles
```
Bastiantheone: Done

fastjet/internal/delaunay/point.go
Bastiantheone: Done

fastjet/internal/delaunay/point.go
Bastiantheone: Done

Bastiantheone: Done

fastjet/internal/delaunay/point.go
Bastiantheone: Done

fastjet/internal/delaunay/point.go
```diff
@@ -0,0 +1,134 @@
...
+	// id is used when points are removed. Copies of the points around the point to
+	// be removed are made. The id is set incremental in counterclockwise order. It identifies
+	// the original the point. The id is then used to determine if a Triangle is inside or outside the
+	// polygon formed by all those points.
+	id      int
+	nearest *Point
+	// dist is the squared distance to the nearest neighbor.
```
Bastiantheone: Done

fastjet/internal/delaunay/point.go
```diff
@@ -0,0 +1,134 @@
...
+	dist float64
```
Bastiantheone: Done

fastjet/internal/delaunay/point.go
```diff
@@ -0,0 +1,134 @@
...
+// NewPoint returns Point for the given x,y coordinates
+func NewPoint(x, y float64) *Point {
+	return &Point{
+		x:    x,
+		y:    y,
+		dist: math.Inf(1),
+	}
+}
+
+// NearestNeighbor returns the nearest Neighbor and the distance to that neighbor.
```
Bastiantheone: Done

fastjet/internal/delaunay/point.go
```diff
@@ -0,0 +1,134 @@
...
+func (p *Point) NearestNeighbor() (*Point, float64) {
+	return p.nearest, math.Sqrt(p.dist)
+}
+
+// Coordinates returns the X,Y Coordinates of a Point.
```
Bastiantheone: Done

fastjet/internal/delaunay/point_test.go
```diff
@@ -0,0 +1,100 @@
+// Copyright 2017 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package delaunay
+
+import (
+	"testing"
+
+	"github.com/gonum/floats"
```
Bastiantheone: Done

fastjet/internal/delaunay/point_test.go
```diff
@@ -0,0 +1,100 @@
...
+		{0.5, 4, 1, 4, false},
+		{-3, 20, 20, -3, false},
+		{2.5, 4, 2.5, 4, true},
+	}
+	for _, test := range tests {
+		p1 := NewPoint(test.x1, test.y1)
+		p2 := NewPoint(test.x2, test.y2)
+		got := p1.Equals(p2)
+		if got != test.want {
+			t.Fatalf("%v == %v, got = %v, want = %v", p1, p2, got, test.want)
```
Bastiantheone: Done

fastjet/internal/delaunay/point_test.go
```diff
@@ -0,0 +1,100 @@
...
+	}{
+		{1, 1, 3, 3, 8},
+		{0, 0, 0, 5, 25},
+	}
+	for _, test := range tests {
+		p1 := NewPoint(test.x1, test.y1)
+		p2 := NewPoint(test.x2, test.y2)
+		got := p1.distance(p2)
+		if got != test.want {
+			t.Fatalf("Distance between %v and %v, got = %v, want = %v", p1, p2, got, test.want)
```
Bastiantheone: Done

fastjet/internal/delaunay/point_test.go
```diff
@@ -0,0 +1,100 @@
...
+			2,
+			NewPoint(2, 0),
+		},
+	}
+	for _, test := range tests {
+		test.p.adjacentTriangles = test.adjacentTriangles
+		test.p.findNearest()
+		gotNeighbor, gotDistance := test.p.NearestNeighbor()
+		if !floats.EqualWithinAbs(gotDistance, test.wantDist, tol) || !test.wantNeighbor.Equals(gotNeighbor) {
+			t.Fatalf("Nearest Neighbor for P%v with adjacent triangles %v, \n gotDist = %v, wantDist = %v, gotNeighbor = %v, wantNeighbor = %v",
```
Bastiantheone: Done

fastjet/internal/delaunay/triangle.go
```diff
@@ -0,0 +1,147 @@
...
+	"go-hep.org/x/hep/fastjet/internal/predicates"
+)
+
+// Triangle is a set of three points that make up a triangle. It stores hierarchical information to find triangles.
+type Triangle struct {
+	// children are triangles that lead to the removal of this triangle.
+	// When a point is inserted the triangle that contains the point is found by going down the hierarchical tree.
+	// The tree's root is the root triangle and the children slice contains the children triangles.
+	children []*Triangle
+	// A,B,C are the points that make up the triangle
```
Bastiantheone: Done

fastjet/internal/delaunay/triangle.go
```diff
@@ -0,0 +1,147 @@
...
+	A, B, C *Point
+	// isInTriangulation holds whether a triangle is part of the triangulation
+	isInTriangulation bool
+}
+
+// NewTriangle returns a triangle formed out of the three given points.
+func NewTriangle(a, b, c *Point) *Triangle {
+	// order the points counter clockwise
+	o := predicates.Orientation(a.x, a.y, b.x, b.y, c.x, c.y)
+	if o == predicates.CW {
```
Bastiantheone: Done

fastjet/internal/delaunay/triangle.go
```diff
@@ -0,0 +1,147 @@
...
+	return &Triangle{
+		A: a,
+		B: b,
+		C: c,
+	}
+}
+
+// add is used when the triangle t is added to the delaunay triangulation.
+// It updates the information of the points in t.
+// It returns all points who's nearest neighbor was updated.
```
Bastiantheone: Done

fastjet/internal/delaunay/triangle.go
```diff
@@ -0,0 +1,147 @@
...
+func (t *Triangle) add() (updated []*Point) {
```
Bastiantheone: Done

fastjet/internal/delaunay/triangle.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	if localMin < t.C.dist {
+		t.C.dist = localMin
+		t.C.nearest = localNearest
+		updated = append(updated, t.C)
+	}
+	return updated
+}
+
+// remove is used when the triangle t is remove from the delaunay triangulation.
```
Bastiantheone: Done

fastjet/internal/delaunay/triangle.go
```diff
@@ -0,0 +1,147 @@
...
+// It updates the information of the points in t.
+// It returns all points who's nearest neighbor was updated.
```
Bastiantheone: Done

fastjet/internal/delaunay/triangle.go
```diff
@@ -0,0 +1,147 @@
...
+func (t *Triangle) remove() (updated []*Point) {
```
Bastiantheone: Done

fastjet/internal/delaunay/triangle.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (t *Triangle) String() string {
+	return fmt.Sprintf("{A%s, B%s, C%s}", t.A, t.B, t.C)
+}
+
+type triangles []*Triangle
+
+// remove removes given triangles from a slice of triangles
+func (t triangles) remove(triangles ...*Triangle) triangles {
```
Bastiantheone: Done

fastjet/internal/delaunay/triangle.go
```diff
@@ -0,0 +1,147 @@
...
+	for i := len(t) - 1; i >= 0; i-- {
+		for j, triangleToRemove := range triangles {
```
Bastiantheone: Done

fastjet/internal/delaunay/triangle.go
Bastiantheone: No the order doesn't matter; it's changed

fastjet/internal/delaunay/triangle.go
Bastiantheone: Done and if you wanted to remove the same triangle twice triangles would have to contain that triangle twice

fastjet/internal/delaunay/triangle_test.go
Bastiantheone: Done

fastjet/internal/delaunay/triangle_test.go
Bastiantheone: Done

fastjet/internal/delaunay/triangle_test.go
Bastiantheone: Done

fastjet/internal/delaunay/triangle_test.go
Bastiantheone: Done

fastjet/internal/delaunay/triangle.go
Bastiantheone: I don't really like the way I worded the comment but I can't think of any good way:
`// To remove a duplicate triangle in ts, triangles would need to contain the triangle twice.`
Do you have a better suggestion?

sbinet: what about:

```go
// remove removes given triangles from a slice of triangles.
//
// remove may modify the content of the input slice of triangles to remove from the receiver.
// remove assumes there are no duplicate triangles in the receiver list of triangles.
// remove will fail to remove the duplicates if that assumption does not hold.
func (...)
```

fastjet/internal/delaunay/triangle_test.go
```diff
@@ -71,7 +73,7 @@ func TestTriangleAdd(t *testing.T) {
 	}
 	for i := range got {
 		if !got[i].Equals(want[i]) {
-			t.Fatalf("After adding triangle nearest neighbor of %v, got = %v, want = %v", points[i], got[i], want[i])
+			t.Errorf("After adding triangle nearest neighbor of %v index %v, got = %v, want = %v", points[i], i, got[i], want[i])
```
sbinet: what about:
```go
t.Errorf("After adding triangle nearest neighbor of points[%v]=%v, got = %v, want = %v", i, points[i], got[i], want[i])
```
instead ?

fastjet/internal/delaunay/triangle_test.go
```diff
@@ -97,7 +99,7 @@ func TestTriangleRemove(t *testing.T) {
 	}
 	for i := range got {
 		if !got[i].Equals(want[i]) {
-			t.Fatalf("After removing triangle nearest neighbor of %v, got = %v, want = %v", points[i], got[i], want[i])
+			t.Errorf("After removing triangle nearest neighbor of %v index %v, got = %v, want = %v", points[i], i, got[i], want[i])
```
sbinet: what about:
```go
t.Errorf("After removing triangle nearest neighbor of points[%d]=%v, got = %v, want = %v", i, points[i], got[i], want[i])
```
instead ?

fastjet/internal/delaunay/triangle.go
Bastiantheone: Done

fastjet/internal/delaunay/triangle_test.go
Bastiantheone: That's better

fastjet/internal/delaunay/triangle_test.go
Bastiantheone: Done

kortschak:
I think this is OK after a brief look. The documentation needs improvement, but it's probably easier to fix that in a PR from one of us than go back and forth fixing it here.

Something that might be worth considering is changing `inTriangle` to return a `type location byte` with `const ( inside location = 1; onEdge location = 2; outside location = 3; )`. The depends on the use, but will likely make the code more readable.

fastjet/internal/delaunay/point.go
```diff
@@ -112,15 +112,38 @@ func (p *Point) findNearest() {
...
+	case inside:
+		return "Inside Triangle"
+	case onEdge:
+		return "On Edge of Triangle"
+	case outside:
+		return "Outside Triangle"
+	default:
+		panic(fmt.Errorf("delaunay: unknown location %d", int(l)))
+	}
+	panic("unreachable")
```
kortschak: This is not necessary (anymore).

fastjet/internal/delaunay/point.go
```diff
@@ -112,15 +112,38 @@ func (p *Point) findNearest() {
...
+		return outside
+	}
+	return onEdge
+}
+
+// location is the position of a point relative to a triangle
+type location int
+
+const (
+	inside location = iota
```
kortschak: Make this `inside location = iota + 1`. This way an uninitialise location is invalid.

fastjet/internal/delaunay/point.go
```diff
@@ -112,15 +112,38 @@ func (p *Point) findNearest() {
...
+	onEdge
+	outside
+)
+
+func (l location) String() string {
+	switch l {
+	case inside:
+		return "Inside Triangle"
```
kortschak: Mixed case in messages is a bit of a no-no. I'd simplify this to "inside", "edge" and "outside", then people can decide how they want to use them.

fastjet/internal/delaunay/point.go
```diff
@@ -112,15 +112,38 @@ func (p *Point) findNearest() {
...
+	case onEdge:
+		return "On Edge of Triangle"
+	case outside:
+		return "Outside Triangle"
+	default:
+		panic(fmt.Errorf("delaunay: unknown location %d", int(l)))
```
kortschak: Does this need to be an error?

fastjet/internal/delaunay/point.go
Bastiantheone: Done

fastjet/internal/delaunay/point.go
Bastiantheone: Done

fastjet/internal/delaunay/point.go
Bastiantheone: Done

fastjet/internal/delaunay/point.go
Bastiantheone: Not necessarily. Instead should I return `return fmt.Sprintf("unknown location %d", int(l))`?

kortschak: Yes, that's fine. Note that with the `%d` verb, you won't need to convert to an `int`. See https://play.golang.org/p/I7m7B8VyvD.

Bastiantheone: got it

sbinet:
(please squash and I'll merge)

fastjet/internal/delaunay/triangle.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+// Triangle is a set of three points that make up a triangle. It stores hierarchical information to find triangles.
+type Triangle struct {
+	// children are triangles that lead to the removal of this triangle.
+	// When a point is inserted the triangle that contains the point is found by going down the hierarchical tree.
+	// The tree's root is the root triangle and the children slice contains the children triangles.
+	children []*Triangle
+	// A,B,C are the CCW-oriented points that make up the triangle
+	A, B, C *Point
```
vladimir-ch: If these points were in `[3]*Point`, I think a lot of the switch casing could be reduced to a loop. This would simplify the code at various places. For example in findNearest.

Bastiantheone:
It is squashed

fastjet/internal/delaunay/triangle.go
sbinet: submitted https://github.com/go-hep/hep/issues/79 so we don't forget.

sbinet:
thanks.

Issue #87, opened 2017-08-15
meta: submit go-hep to JOSS


sbinet:
We should consider submitting a paper about `go-hep` to the journal of open source software:

http://joss.theoj.org/about

sbinet:
@mughetto @decibelCooper @Bastiantheone @ebusato 

could you please add your contact infos in `paper.md` so I can submit this to JOSS?

FYI, orcid is https://orcid.org/

Bastiantheone:
To be clear, I just need to add my school as affiliation?

sbinet:
Your school and create an orcid account if not already (and tell us your orcid.)
Thanks.

sbinet:
thanks all for the prompt reactivity.
I know @ebusato is still on holidays. He should be back by the end of August.

decibelCooper:
Nice work, @sbinet

sbinet:
thanks! (BTW: you'll have to share (at some point) the nice LCIO related plots that you've been creating in your repos. much interested in getting some actual HEP research plots :) Perhaps a blog entry on https://go-hep.org?)

decibelCooper:
@sbinet Sure, if you don't mind them being about EIC studies =) (a little disconnected from the HEP community, I think).  I'll keep this in mind next time I'm working on plots.

PR #88, opened 2017-08-15
joss-paper: first stab

sbinet:
Updates go-hep/hep#87.


sbinet:
BTW, @go-hep/developers feel free to comment, suggest improvements, etc... :)

arfon:
@sbinet - please take a look at the submission guidelines for [what your paper should contain](http://joss.theoj.org/about#paper_structure). The current text is good but it's very terse - a little additional detail would be great.

sbinet:
@arfon PTAL.
did I miss anything?

arfon:
@sbinet - this looks good, thanks. My only concern with what you have now is that there's a lot of content describing the different packages now which you may get a little push back on during review as we usually expect/ask to see content like this is technical documentation.

sbinet:
ok. well, as for code, removing is always easier :)

thanks!

arfon:
> ok. well, as for code, removing is always easier :)

Ha, yes!

PR #103, opened 2017-08-26
fastjet/internal/delaunay: remove root points from updated & set ID incrementally

Bastiantheone:
Added a function to remove the root points from the updatedNearestNeighbor slice from Insert and Remove.
Set the ID incrementally and added a function to access the ID, so that a jet can be mapped from a point.

### DIFF
--- a/fastjet/internal/delaunay/delaunay.go
+++ b/fastjet/internal/delaunay/delaunay.go
@@ -20,6 +20,7 @@ type Delaunay struct {
 	// to locate a point. The variable root's nil-ness also indicates which method to use to locate a point.
 	root *Triangle
 	r    *rand.Rand
+	n    int // n is the number of points inserted. It is used to assign the id to points.
 }
 
 // HierarchicalDelaunay creates a Delaunay Triangulation using the delaunay hierarchy.
@@ -42,6 +43,7 @@ func HierarchicalDelaunay() *Delaunay {
 	root := NewTriangle(a, b, c)
 	return &Delaunay{
 		root: root,
+		n:    0,
 	}
 }
 
@@ -69,18 +71,23 @@ func (d *Delaunay) Triangles() []*Triangle {
 // whose nearest neighbor changed due to the insertion. The slice may contain
 // duplicates.
 func (d *Delaunay) Insert(p *Point) (updatedNearestNeighbor []*Point) {
+	p.id = d.n
+	d.n++
 	if len(p.adjacentTriangles) == 0 {
 		p.adjacentTriangles = make(triangles, 0)
 	}
 	p.adjacentTriangles = p.adjacentTriangles[:0]
 	t, l := d.locatePointHierarchy(p, d.root)
+	var updated points
 	switch l {
 	case inside:
-		return d.insertInside(p, t)
+		updated = d.insertInside(p, t)
 	case onEdge:
-		return d.insertOnEdge(p, t)
+		updated = d.insertOnEdge(p, t)
+	default:
+		panic(fmt.Errorf("delaunay: no triangle containing point %v", p))
 	}
-	panic(fmt.Errorf("delaunay: no triangle containing point %v", p))
+	return updated.remove(d.root.A, d.root.B, d.root.C)
 }
 
 // Remove removes the point from the triangulation. It returns the points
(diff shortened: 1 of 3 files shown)


sbinet:
I am about to embark on my flight back to Europe.
I'll review this on Sunday, Paris time.
Thanks for the PR :) (your work has drawn some interest during my ACAT talk. People were really enthusiast!)

Bastiantheone:
I'm happy to hear that :) Safe flight!

fastjet/internal/delaunay/delaunay.go
```diff
@@ -68,25 +70,29 @@ func (d *Delaunay) Triangles() []*Triangle {
 // Insert inserts the point into the triangulation. It returns the points
 // whose nearest neighbor changed due to the insertion. The slice may contain
 // duplicates.
-func (d *Delaunay) Insert(p *Point) (updatedNearestNeighbor []*Point) {
+func (d *Delaunay) Insert(p *Point) (updatedNearestNeighbor points) {
```
sbinet: having an exported method returning a value of type `points` which isn't itself exported is a bit cumbersome.

either export `points` or re-convert the `points` to `[]*Point` when returning from this method.
Looking at the `points` type, it only has one method, that is not exported.
Thus, I would go with the latter (keep `[]*Point` in the method signature and work with `points` inside the method, re-converting to `[]*Point` when returning.)

fastjet/internal/delaunay/delaunay.go
```diff
@@ -68,25 +70,29 @@ func (d *Delaunay) Triangles() []*Triangle {
...
 	}
-	panic(fmt.Errorf("delaunay: no triangle containing point %v", p))
+	return updatedNearestNeighbor.remove(d.root.A, d.root.B, d.root.C)
 }
 
 // Remove removes the point from the triangulation. It returns the points
 // whose nearest neighbor changed due to the removal. The slice may contain
 // duplicates.
-func (d *Delaunay) Remove(p *Point) (updatedNearestNeighbor []*Point) {
+func (d *Delaunay) Remove(p *Point) (updatedNearestNeighbor points) {
```
sbinet: ditto.

fastjet/internal/delaunay/point.go
```diff
@@ -46,6 +46,10 @@ func (p *Point) Coordinates() (x, y float64) {
 	return p.x, p.y
 }
 
+func (p *Point) ID() int {
```
sbinet: do we need to export this?
in that case, please document that exported method.

fastjet/internal/delaunay/point.go
```diff
@@ -210,3 +214,26 @@ func (l location) String() string {
...
 	}
 }
+
+type points []*Point
+
+// remove removes given points from a slice of points.
+//
+// remove will remove all occurrences of the points.
+func (ps points) remove(pts ...*Point) points {
+	points := make(points, 0, len(ps))
```
sbinet: too many `points` :)
I would call the returned value `out`,  `o` or `opoints`.

fastjet/internal/delaunay/point.go
```diff
@@ -20,10 +20,10 @@ type Point struct {
 	adjacentTriangles triangles // adjacentTriangles is a list of triangles containing the point.
 	nearest           *Point
 	dist2             float64 // dist2 is the squared distance to the nearest neighbor.
-	// id is used when points are removed. Copies of the points around the point to
-	// be removed are made. The ID is set incremental in counterclockwise order. It identifies
-	// the original. It is also used to determine whether a Triangle is inside or outside the
-	// polygon formed by all those points.
+	// id is a unique identifier, that is assigned incrementally to a point on insertion.
```
sbinet: BTW, could you also document in the doc-string of `Point` that one should use the `Equal` method of `Point` to test whether 2 points are equal ?
(and not use the `==` operator. the comment on `id` made me think of that. it's a similar use case (trap, really) to `time.Time`...)

fastjet/internal/delaunay/delaunay.go
Bastiantheone: done

fastjet/internal/delaunay/delaunay.go
Bastiantheone: done

fastjet/internal/delaunay/point.go
Bastiantheone: done

fastjet/internal/delaunay/point.go
Bastiantheone: done

fastjet/internal/delaunay/point.go
Bastiantheone: And yes we need to export it because we need a way of mapping the points back to the jets

fastjet/internal/delaunay/point.go
Bastiantheone: Done

Bastiantheone:
Done

PR #106, opened 2017-08-29
fastjet/internal/delaunay: handle duplicate points

Bastiantheone:
This PR handles duplicate points. When a duplicate point is inserted it only updates the nearest neighbor of itself and the duplicate point. When it is removed nothing happens.

### DIFF
--- a/fastjet/internal/delaunay/delaunay.go
+++ b/fastjet/internal/delaunay/delaunay.go
@@ -36,6 +36,9 @@ type Delaunay struct {
 // All triangles in the current triangulation are leaf triangles.
 // To find the triangle which contains the point, the algorithm follows the graph until
 // a leaf is reached.
+//
+// Duplicate points don't get inserted, but the nearest neighbor is set to the corresponding point.
+// When a duplicate point is removed nothing happens.
 func HierarchicalDelaunay() *Delaunay {
 	a := NewPoint(-1<<30, -1<<30)
 	b := NewPoint(1<<30, -1<<30)
@@ -95,6 +98,10 @@ func (d *Delaunay) Insert(p *Point) (updatedNearestNeighbor []*Point) {
 // duplicates.
 func (d *Delaunay) Remove(p *Point) (updatedNearestNeighbor []*Point) {
 	if len(p.adjacentTriangles) < 3 {
+		if p.dist2 == 0 {
+			// must be a duplicate point, therefore don't panic
+			return updatedNearestNeighbor
+		}
 		panic(fmt.Errorf("delaunay: can't remove point %v, not enough adjacent triangles", p))
 	}
 	var updated points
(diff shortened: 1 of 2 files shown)


fastjet/internal/delaunay/delaunay.go
```diff
@@ -162,6 +169,36 @@ func (d *Delaunay) insertInside(p *Point, t *Triangle) []*Point {
 // insertOnEdge inserts a point on an Edge between two triangles. It returns the points
 // whose nearest neighbor changed.
 func (d *Delaunay) insertOnEdge(p *Point, t *Triangle) []*Point {
+	// Check if p is a duplicate
+	switch {
+	case p.Equals(t.A):
+		if p.id == t.A.id {
+			panic(fmt.Errorf("Point %v was previously inserted", p))
```
sbinet: please prepend a `"delaunay: ` string to the panic message.

fastjet/internal/delaunay/delaunay.go
```diff
@@ -162,6 +169,36 @@ func (d *Delaunay) insertInside(p *Point, t *Triangle) []*Point {
...
+		}
+		p.nearest = t.A
+		p.dist2 = 0
+		t.A.nearest = p
+		t.A.dist2 = 0
+		return []*Point{p, t.A}
+	case p.Equals(t.B):
+		if p.id == t.B.id {
+			panic(fmt.Errorf("Point %v was previously inserted", p))
```
sbinet: please prepend a `"delaunay: ` string to the panic message.

fastjet/internal/delaunay/delaunay.go
```diff
@@ -162,6 +169,36 @@ func (d *Delaunay) insertInside(p *Point, t *Triangle) []*Point {
...
+		}
+		p.nearest = t.B
+		p.dist2 = 0
+		t.B.nearest = p
+		t.B.dist2 = 0
+		return []*Point{p, t.B}
+	case p.Equals(t.C):
+		if p.id == t.C.id {
+			panic(fmt.Errorf("Point %v was previously inserted", p))
```
sbinet: please prepend a `"delaunay: ` string to the panic message.

fastjet/internal/delaunay/delaunay.go
```diff
@@ -162,6 +169,36 @@ func (d *Delaunay) insertInside(p *Point, t *Triangle) []*Point {
 // insertOnEdge inserts a point on an Edge between two triangles. It returns the points
 // whose nearest neighbor changed.
 func (d *Delaunay) insertOnEdge(p *Point, t *Triangle) []*Point {
+	// Check if p is a duplicate
+	switch {
```
sbinet: can `t` be made of 2 or even 3 duplicates?
e.g.: can `p.Equals(t.A)` and `p.Equals(t.B)` be both true?
presumably a `Triangle` can't be made like so, but... as we now have some degeneracy...

Bastiantheone: No, because NewTriangle panics when three points are colinear

sbinet: good :)

fastjet/internal/delaunay/delaunay.go
Bastiantheone: Done

fastjet/internal/delaunay/delaunay.go
Bastiantheone: Done

fastjet/internal/delaunay/delaunay.go
Bastiantheone: Done

fastjet/internal/delaunay/delaunay.go
Bastiantheone: Should I change it to let triangles be composed of duplicates?

sbinet: no, I believe it's fine as is.
we'll cross that bridge when/if we need to.

Bastiantheone:
Done

PR #108, opened 2017-08-31
fastjet/internal/delaunay: add function SecondNearestNeighbor

Bastiantheone:
Add the function SecondNearestNeighbor to handle when a jet's nearest neighbor is its own mirror point.  
Mirror points can be identified in runNlnN, but then the nearest neighbor that is not the mirror points needs to be found. That is when SecondNearestNeighbor comes to use.

### DIFF
--- a/fastjet/internal/delaunay/point.go
+++ b/fastjet/internal/delaunay/point.go
@@ -63,6 +63,64 @@ func (p *Point) Equals(v *Point) bool {
 	return p == v || (p.x == v.x && p.y == v.y)
 }
 
+// SecondNearestNeighbor looks at all adjacent points of p and returns the second nearest one
+// and the distance to that point.
+func (p *Point) SecondNearestNeighbor() (*Point, float64) {
+	var nearest, secondNearest *Point
+	min, secondMin := math.Inf(1), math.Inf(1)
+	if p.dist2 == 0 {
+		// p has a duplicate
+		nearest = p.nearest
+		min = 0
+	}
+	for _, t := range p.adjacentTriangles {
+		var p2, p3 *Point
+		// find the point in t that is not p
+		switch {
+		case p.Equals(t.A):
+			p2 = t.B
+			p3 = t.C
+		case p.Equals(t.B):
+			p2 = t.A
+			p3 = t.C
+		case p.Equals(t.C):
+			p2 = t.A
+			p3 = t.B
+		default:
+			panic(fmt.Errorf("delaunay: point %v not found in %v", p, t))
+		}
+		dist := p.distance(p2)
+		switch {
+		case dist < min:
+			min, secondMin = dist, min
+			nearest, secondNearest = p2, nearest
+			if p2.dist2 == 0 {
+				// p2 has a duplicate
+				secondNearest = p2.nearest
+				secondMin = dist
+			}
+		case dist < secondMin:
+			secondMin = dist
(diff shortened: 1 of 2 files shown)


fastjet/internal/delaunay/point.go
```diff
@@ -63,6 +63,62 @@ func (p *Point) Equals(v *Point) bool {
...
+			p2 = t.A
+			p3 = t.C
+		case p.Equals(t.C):
+			p2 = t.A
+			p3 = t.B
+		default:
+			panic(fmt.Errorf("delaunay: point %v not found in %v", p, t))
+		}
+		dist := p.distance(p2)
+		if dist < min {
```
sbinet: I would use a `switch` instead of that `if-else-if`:

```go
switch {
case dist < min:
case dist < secondMin:
}
```

fastjet/internal/delaunay/point.go
```diff
@@ -63,6 +63,62 @@ func (p *Point) Equals(v *Point) bool {
...
+				// p2 has a duplicate
+				secondNearest = p2.nearest
+				secondMin = dist
+			}
+		} else if dist < secondMin {
+			secondMin = dist
+			secondNearest = p2
+		}
+		dist = p.distance(p3)
+		if dist < min {
```
sbinet: ditto: please use a `switch`.

fastjet/internal/delaunay/point_test.go
```diff
@@ -77,6 +78,34 @@ func TestFindNearest(t *testing.T) {
...
 
+func TestFindSecondNearest(t *testing.T) {
+	const tol float64 = 0.001
+	tests := []struct {
+		p                 *Point
+		adjacentTriangles []*Triangle
+		wantDist          float64
+		wantNeighbor      *Point
+	}{
+		{
```
sbinet: ok.

could you also add:

- a test that triggers the `if p.dist2 == 0 {` of line 71?
I'd like to see how this plays out later on wrt the (now `if-else-if`, later) `switch` which should always take the `dist < secondMin` branch...
- a test that triggers the `if p2.dist2 == 0` of line 96
- and the same for `p3` ?

fastjet/internal/delaunay/point.go
Bastiantheone: done

fastjet/internal/delaunay/point.go
Bastiantheone: done

fastjet/internal/delaunay/point_test.go
Bastiantheone: Done

In the test function:
The first test is without duplicates
The second test will trigger `if p.dist2 == 0{`
The third test will trigger `if p2.dist2 == 0{`
And the fourth test will trigger `if p3.dist2 == 0{`

fastjet/internal/delaunay/point_test.go
```diff
@@ -77,6 +78,73 @@ func TestFindNearest(t *testing.T) {
 	}
 }
 
+func TestFindSecondNearest(t *testing.T) {
+	const tol float64 = 0.001
```
sbinet: the `float64` isn't really needed.

Bastiantheone: done

Bastiantheone:
Done and no problem

Issue #111, opened 2017-09-05
fastjet/internal/delaunay: border points in voronoi diagram


Bastiantheone:
How should I go about points on the outside in the vornoi diagram? I could either return the voronoi diagram where the voronoi points on the outside connect to a user defined border, or I could return it where voronoi point on the outside connect to their neighbor. 
The second option would be easier and probably be more useful, but the first version will definitively look better when plotted. I have planned to return the area, so the area for points on the outside would be infinity and therefore users can identify those easily.
Is there a way to make a line go towards a direction with the gonum plotter? In that case I would like to create a polygon that doesn't connect on all sides. If that's possible it would be a way to use the second option and make it plot nicely.

Bastiantheone:
Or another option would be to have the border vornoi points not connect to anything

Bastiantheone:
What do you think @sbinet ?

sbinet:
humm... right now, I am rather agnostic.
I'd say, start with the easy option and see where that gets us ?

sbinet:
(so, if I recap, between options a), b) and c), I'd go with b), IIUC)

Bastiantheone:
SGTM

PR #114, opened 2017-09-08
fastjet/internal/delaunay: implement voronoi

Bastiantheone:
This PR adds several functions to compute the Voronoi Diagram.

To handle the issue #111 I initially started implementing it where Voronoi points on the outside connect to its neighbor. But then after thinking about it more, I changed it to connect to the Circumcenter of the adjacent root triangle, which is just a dummy triangle. Those points are moved to the first and last position in the slice so that they are easy to identify. They might be useful to get a direction of where the line goes towards.

### DIFF
--- a/fastjet/internal/delaunay/delaunay.go
+++ b/fastjet/internal/delaunay/delaunay.go
@@ -6,6 +6,7 @@ package delaunay
 
 import (
 	"fmt"
+	"math"
 	"math/rand"
 
 	"go-hep.org/x/hep/fastjet/internal/predicates"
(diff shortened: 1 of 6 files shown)


fastjet/internal/delaunay/triangle.go
```diff
@@ -121,6 +122,42 @@ func (t *Triangle) remove() []*Point {
...
+			x = (b2 - b1) / (m1 - m2)
+			y = m1*x + b1
+			if math.IsNaN(y) {
+				// Should never reach this point. To reach this point either all three y or x coordinates
+				// would have to be the same, which is impossible since there are no collinear triangles.
+				// Or two y variables and two x variables would have to be the same. That is impossible,
+				// because of the way this is set up. The variables with the same x coordinates would have to
+				// be the same as the variables with the same y coordinates and triangles are not made up
+				// of duplicates.
+				panic(fmt.Errorf("delaunay: error caluclating the circumcenter of %v", t))
```
sbinet: `s/caluclating/calculating/`

fastjet/internal/delaunay/delaunay.go
```diff
@@ -473,3 +474,49 @@ func areCounterclockwise(a, b, c int) bool {
...
+			}
+		}
+		x, y = t.circumcenter()
+		voronoi[i] = NewPoint(x, y)
+	}
+	if border1 == -1 {
+		area := voronoi.polyArea()
+		return voronoi, area
+	}
+	if border1 != 0 || border2 != len(voronoi)-1 {
```
sbinet: I think this `if-check` could use some explanation.

at this point, `border1` is either `0` or `>= 1`.
but couldn't `border2` be `-1` or `> border1` ?

if `border2=-1`, we might have a problem with the slice-trick below.

there is also probably an issue when `len(voronoi) == 1` (or `len(p.adjacentTriangles) <= 1`)
(as `VoronoiCell` is an exported method, we should check for that pathological case. perhaps a bit around L488)

fastjet/internal/delaunay/triangle.go
```diff
@@ -121,6 +122,42 @@ func (t *Triangle) remove() []*Point {
 	return updated
 }
 
+// circumcenter returns the x,y coordinates of the circumcenter of the triangle
+func (t *Triangle) circumcenter() (x, y float64) {
+	m1 := (t.A.x - t.B.x) / (t.B.y - t.A.y)
```
sbinet: there are a few divisions by the same value (modulo the sign).

not sure whether `circumcenter` will end up in the hot path, but perhaps put a `TODO` to store the inverse of these delta values and use that instead.

fastjet/internal/delaunay/delaunay.go
```diff
@@ -473,3 +474,49 @@ func areCounterclockwise(a, b, c int) bool {
 	}
 	return a < b && c < a
 }
+
+// VoronoiCell returns the Vornoi points of a point in clockwise order
+// and the area those points enclose.
+//
+// If a point is on the border of the Delaunay triangulation the area will be Infinity
+// and the first and last point of the cell will be of a root triangle.
```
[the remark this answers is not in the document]
Bastiantheone: done

fastjet/internal/delaunay/delaunay.go
```diff
@@ -473,3 +474,49 @@ func areCounterclockwise(a, b, c int) bool {
...
+func (d *Delaunay) VoronoiCell(p *Point) ([]*Point, float64) {
+	// border1 is set to the index of the first vornoi point that is part of a root triangle
```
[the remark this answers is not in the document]
Bastiantheone: done

fastjet/internal/delaunay/delaunay.go
```diff
@@ -473,3 +474,49 @@ func areCounterclockwise(a, b, c int) bool {
...
+	border1 := -1
+	// border2 is set to the index of the first vornoi point that is part of a root triangle
```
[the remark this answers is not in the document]
Bastiantheone: done

fastjet/internal/delaunay/delaunay.go
Bastiantheone: done

fastjet/internal/delaunay/delaunay.go
```diff
@@ -473,3 +474,49 @@ func areCounterclockwise(a, b, c int) bool {
...
+		}
+		x, y = t.circumcenter()
+		voronoi[i] = NewPoint(x, y)
+	}
+	if border1 == -1 {
+		area := voronoi.polyArea()
+		return voronoi, area
+	}
+	if border1 != 0 || border2 != len(voronoi)-1 {
+		// reposition the points, so that the border points are the first and last point
```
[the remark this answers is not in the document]
Bastiantheone: done

fastjet/internal/delaunay/triangle.go
Bastiantheone: done

fastjet/internal/delaunay/triangle.go
Bastiantheone: I'm not sure I got it right. 
For example since there is `.../(t.B.y - t.A.y)` and `.../(t.A.y - t.B.y)` we could have `invDeltaAB := 1/(t.B.y - t.A.y)` and use an if statement to check if it is positive or negative.
Did I understand that correctly?

sbinet: leaving as a TODO is fine for now.

fastjet/internal/delaunay/delaunay.go
sbinet: ok. perhaps document that the function will panic if the number of adjacent triangles is <= 3.

Bastiantheone:
done

PR #127, opened 2017-09-17
rootio: have RBuffer.ReadF64 take an input arg

sbinet:


sbinet:
performances of groot-reading are ok, now.

PR #128, opened 2017-09-18
fastjet/internal/plot: plot voronoi diagram

Bastiantheone:
Function to plot the voronoi diagram


fastjet/internal/plot/voronoi.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		p.Add(poly)
+	}
+	s, err := plotter.NewScatter(pts)
+	if err != nil {
+		return err
+	}
+	p.Add(s)
+	// cut plot off to only show the points plus a little extra space.
+	bufSpaceX := (maxX - minX) * .05
```
sbinet: `s/bufSpaceX/xpad/`

fastjet/internal/plot/voronoi.go
```diff
@@ -0,0 +1,74 @@
...
+	bufSpaceY := (maxY - minY) * .05
```
sbinet: `s/bufSpaceY/ypad/`

fastjet/internal/plot/voronoi.go
```diff
@@ -0,0 +1,74 @@
...
+			return err
+		}
+		p.Add(poly)
+	}
+	s, err := plotter.NewScatter(pts)
+	if err != nil {
+		return err
+	}
+	p.Add(s)
+	// cut plot off to only show the points plus a little extra space.
```
sbinet: shouldn't this be taken care of by the polygons themselves ?
ie: the `plot.DataRanger` interface should have correctly detected the correct plot data range to display, and in the case of glyphes, there's the `plot.GlyphBoxer`

- https://github.com/gonum/plot/blob/master/plot.go#L66
- https://github.com/gonum/plot/blob/master/plot.go#L318

see:
- https://github.com/gonum/plot/wiki/Creating-Custom-Plotters:-A-tutorial-on-creating-custom-Plotters

It's not a blocker, I am just curious.

fastjet/internal/plot/voronoi.go
Bastiantheone: done

fastjet/internal/plot/voronoi.go
Bastiantheone: done

fastjet/internal/plot/voronoi.go
Bastiantheone: For points on the outside of the Delaunay triangulation, the first and last point of the Voronoi cell are far outside. They are just there to show the direction the Voronoi cell goes. And they shouldn't be shown

Bastiantheone:
Done :)

Issue #138, opened 2017-11-27
user stories: GDML importer for 3D geometry library


decibelcooper:
The 3D library mentioned in the title is described as a basic concept in #137.

For my work, GDML will be used as a geometry exchange format.  If Go is to be used for HEP/MEP event reconstruction, it will be very useful to be able to accept GDML by constructing an in-memory representation of a detector geometry with an importer.  This could be a part of the library described in #137, but it doesn't need to be.  For the purpose of minimizing code imports for projects that don't need GDML import, I think it makes sense to have this separate.

sbinet:
GDML "specs" are here: http://gdml.web.cern.ch/GDML/doc/GDMLmanual.pdf

decibelcooper:
Ooo!  Much excite!!!

PR #158, opened 2018-04-05
geo/gdml: first import

sbinet:
This CL adds the fundations for a GDML library.

Updates go-hep/hep#138.


sbinet:
@decibelcooper this is just something I did while procrastinating in front of my "GoHEP vision draft article"... :)

PR #163, opened 2018-04-18
Rootio tchain scanner

maloft:
Hello, 
There are many errors in the next proposition, but the idea is to add an argument to the loadEntry method : totalEntries. totalEntries is the total number of entries of the chain. This, i think, will allow to increase the nevbuf (number of entries in the basket) in order to be able to scan a chain containing more than one tree. Is it a way according to you ? 
Thanks

### DIFF
--- a/rootio/tchain_test.go
+++ b/rootio/tchain_test.go
@@ -5,6 +5,9 @@
 package rootio_test
 
 import (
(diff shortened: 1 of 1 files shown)


sbinet:
I only see my own 2 commits :)

Issue #170, opened 2018-04-24
xrootd: implement xrootd protocol


sbinet:
## Specs:

- http://xrootd.org/doc/dev45/XRdv310.htm
- http://xrootd.org/doc/dev49/XRdv400.htm

## Status

- [x] kXR_admin
- [x] kXR_auth (see go-hep/hep#250)
    - [x] kXR_auth + `host`
    - [ ] kXR_auth + `gsi`
    - [x] kXR_auth + `krb5`
    - [ ] kXR_auth + `pwd`
    - [ ] kXR_auth + `sss`
    - [x] kXR_auth + `unix`
- [x] kXR_bind
- [x] kXR_chmod
- [x] kXR_close
- [x] kXR_decrypt
- [x] KXR_dirlist
- [x] KXR_endsess
- [ ] kXR_getfile
- [x] kXR_locate
- [x] kXR_login
- [x] kXR_mkdir
- [x] kXR_mv
- [x] kXR_open
- [x] kXR_ping
- [x] kXR_prepare
- [x] kXR_protocol
- [ ] kXR_putfile
- [x] kXR_query
- [x] kXR_read
- [ ] kXR_readv
- [x] kXR_rm
- [x] kXR_rmdir
- [ ] kXR_set
- [x] kXR_sigver
- [x] kXR_stat
- [x] kXR_statx
- [x] kXR_sync
- [x] kXR_truncate
- [x] kXR_write
- [ ] kXR_writev
- [x] kXR_verifyw

sbinet:
@EgorMatirov this could be broken up into smaller pieces to swallow :)
(_e.g._ the req/rep implementation for everything that pertains to "login" in "readonly", then those that need some basic authentication and basic permissions, etc...)

PR #179, opened 2018-04-26
xrootd: add initial implementation

EgorMatirov:
Updates go-hep/hep#170.


EgorMatirov:
CI failed because tests require a running xrootd server and default address is specified as `0.0.0.0:9001`. I can change it to `ccxrootdgotest.in2p3.fr:9001` for now, but I'm unsure that it's the best solution.

sbinet:
no worries :)

sbinet:
note that I'll be on holidays from 2nd of May to the 7th (with very spotty internet access).

EgorMatirov:
Please take a look at #183. :)
Next PR "add streammanager" is ready to be reviewed but I'm not sure about the process.
Should I publish that PR and later merge (or rebase) changes from #183 or we'll proceed one-by-one without bothering with additional merging\rebasing stuff? I'm fine with both. Basically, I have never made such big PR, so I'm a bit confused with the process. :)

sbinet:
actually, what I meant was to perhaps break this PR into smaller ones :)
it's ok to send a couple of PRs that depend on each others in a somewhat sequential manner so we get an idea of where we're headed.

I'll focus on a few items here for the moment (probably the client as it's the most important thing to get right first and foremost IMHO)

xrootd/LICENSE
```diff
@@ -0,0 +1,23 @@
+Copyright ©2018 The go-hep Authors. All rights reserved.
```
sbinet: this file isn't needed. there's already a `LICENSE` file at the root of the repo.

xrootd/bind.go
```diff
@@ -0,0 +1,28 @@
...
+package xrootd // import "go-hep.org/x/hep/xrootd"
+
+import (
+	"context"
+
+	"go-hep.org/x/hep/xrootd/encoder"
+	"go-hep.org/x/hep/xrootd/requests/bind"
+)
+
+// Bind a socket to pre-existing session
```
sbinet: The Go convention for documenting code would rather recommend to have something like:

```go
// Bind binds the client's socket to a pre-existing session ID.
func (client *Client) Bind(...) {
```

ie: the name of the thing that we are documenting and then a verb that builds a sentence that reads well.
(I am not a native english speaker either so don't worry: I'll make horrible english sentences as well :P)

xrootd/bind.go
```diff
@@ -0,0 +1,28 @@
...
+func (client *Client) Bind(ctx context.Context, sessionID [16]byte) (byte, error) {
+	serverResponse, err := client.call(ctx, bind.RequestID, bind.NewRequest(sessionID))
```
sbinet: `s/serverResponse/resp/` or `response`.
as we are a client, we know we are supposed to talk to a server :)

xrootd/bind.go
```diff
@@ -0,0 +1,28 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	var result = &bind.Response{}
```
sbinet: I would personally write this:
```go
var result bind.Response
```
and pass the address of `result` to `encoder.Unmarshal`.

xrootd/client.go
```diff
@@ -0,0 +1,155 @@
...
+
+	"go-hep.org/x/hep/xrootd/encoder"
+	"go-hep.org/x/hep/xrootd/streammanager"
+)
+
+var logger = log.New(os.Stderr, "xrootd: ", log.LstdFlags)
+
+// A Client to xrootd server
+type Client struct {
+	connection      *net.TCPConn
```
sbinet: `s/connection *net.TCPConn/conn net.Conn/`

`conn` is the usual name for a connection in Go.
also. using a `net.Conn` instead of the concrete underlying type will probably help us during testing (replacing a tcp connection with a `net.Pipe` for example)

xrootd/client.go
```diff
@@ -0,0 +1,155 @@
...
+	sm              *streammanager.StreamManager
```
sbinet: `s/sm/smgr/` ? or just `mgr` ?

xrootd/client.go
```diff
@@ -0,0 +1,155 @@
...
+	"net"
+	"os"
+
+	"go-hep.org/x/hep/xrootd/encoder"
+	"go-hep.org/x/hep/xrootd/streammanager"
+)
+
+var logger = log.New(os.Stderr, "xrootd: ", log.LstdFlags)
+
+// A Client to xrootd server
```
sbinet: `Client` is the main type people will have to deal with.
It could deserve a bit more of documentation :)

xrootd/client.go
```diff
@@ -0,0 +1,155 @@
...
+// New creates a client to xrootd server at address
+func New(ctx context.Context, address string) (*Client, error) {
+	conn, err := createTCPConnection(address)
+	if err != nil {
+		return nil, err
+	}
+
+	client := &Client{conn, streammanager.New(), 0}
+
+	go client.consume()
```
sbinet: this goroutine is leaked: there is no way one can properly close it nor reclaim its resources.
perhaps it should be passed a context value.

naively, when I saw the context value being passed to the `New` function, I had expected it would be bound to the lifetime of the client.

xrootd/client.go
```diff
@@ -0,0 +1,155 @@
...
+
+	err = client.handshake(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return client, nil
+}
+
+func createTCPConnection(address string) (connection *net.TCPConn, err error) {
```
sbinet: this whole function could be removed and replaced with simply:
```go
conn, err := net.Dial("tcp", address)
```

xrootd/client.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		return
+	}
+
+	connection, err = net.DialTCP("tcp", nil, tcpAddr)
+	return
+}
+
+func (client *Client) consume() {
+	for {
```
sbinet: as said above, this `consume` goroutine should take a parent `context` as input and create another context value with a cancel function (that is `defer`-ed)

xrootd/client.go
```diff
@@ -0,0 +1,155 @@
...
+		var header = &responseHeader{}
+
+		var headerBytes = make([]byte, responseHeaderSize)
+		if _, err := io.ReadFull(client.connection, headerBytes); err != nil {
+			logger.Panic(err)
```
sbinet: I am not too keen on having the Client panic like so.
could we devise a way to properly propagate this error back to the user ?

(same remark for all the panics below)

xrootd/client.go
```diff
@@ -0,0 +1,155 @@
...
+		response := &streammanager.ServerResponse{data, nil}
+		if header.Status != 0 {
+			response.Error = extractError(header, data)
+		}
+
+		if err := client.sm.SendData(header.StreamID, response); err != nil {
+			logger.Panic(err)
+		}
+
+		if header.Status != 4000 { // oksofar
```
sbinet: could we loose this magic constant `4000` ? or document it as a real Go `const` ?
or have a method on `Status` or `header` that does the same?

xrootd/client.go
```diff
@@ -0,0 +1,155 @@
...
+
+	connection, err = net.DialTCP("tcp", nil, tcpAddr)
+	return
+}
+
+func (client *Client) consume() {
+	for {
+		var header = &responseHeader{}
+
+		var headerBytes = make([]byte, responseHeaderSize)
```
sbinet: could we put a TODO here to reduce the allocations?
IIUC, `header` and `headerBytes` lifetimes are coupled to the lifetime of a client request.
one can have many client requests in-flight at any given time, so we can't lift these values outside of the `for`-loop.
but these resources could be pooled "by request" and be reused once a request is processed.

xrootd/client.go
```diff
@@ -0,0 +1,155 @@
...
+		if err := encoder.Unmarshal(headerBytes, header); err != nil {
+			logger.Panic(err)
+		}
+
+		data := make([]byte, header.DataLength)
+		if _, err := io.ReadFull(client.connection, data); err != nil {
+			logger.Panic(err)
+		}
+
+		response := &streammanager.ServerResponse{data, nil}
```
sbinet: `s/response/resp/`

xrootd/client.go
```diff
@@ -0,0 +1,155 @@
...
+	if header.Status == 4003 {
+		code := int32(binary.BigEndian.Uint32(data[0:4]))
+		message := string(data[4 : len(data)-1]) // Skip \0 character at the end
+
+		return serverError{code, message}
+	}
+	return nil
+}
+
+func (client *Client) callWithBytesAndResponseChannel(ctx context.Context, responseChannel streammanager.DataReceiveChannel, requestData []byte) (responseBytes []byte, err error) {
```
sbinet: `s/responseBytes/data/`

xrootd/client.go
EgorMatirov: I have added a bit more of documentation, but I'm unsure if that's enough and what else I can add.

xrootd/client.go
EgorMatirov: ```header``` and ```headerBytes``` lifetimes are coupled to the lifetime of a server response, so it's ok to reuse them. There is only one response being processed inside ```consume``` per time and ```header``` and ```headerBytes``` are not passed anywhere.

Issue #177, opened 2018-04-24
xrootd: benchmark+improve client performances


sbinet:

EgorMatirov:
- [ ] Preallocate reasonable amount of bytes in ```MarshalRequest``` https://github.com/go-hep/hep/pull/183#discussion_r185445598.
- [ ]  Create code generator for marshalling to be used instead of reflection https://github.com/go-hep/hep/pull/183#discussion_r185294377.

PR #183, opened 2018-04-29
xrootd: add encoder

EgorMatirov:
Add encoder package which (un)marshals requests/responses to a Go struct.
Specific structs will be added as part of the corresponding request implementation.

Is part of #170, #179.

### DIFF
--- a/xrootd/encoder/encoder.go
+++ b/xrootd/encoder/encoder.go
@@ -1,0 +1,144 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package encoder // import "go-hep.org/x/hep/xrootd/encoder"
+
+import (
+	"encoding/binary"
+	"reflect"
+
+	"github.com/pkg/errors"
+	"go-hep.org/x/hep/xrootd/protocol"
+)
+
+// MarshalRequest encodes request body alongside with request and stream ids
+func MarshalRequest(requestID uint16, streamID protocol.StreamID, requestBody interface{}) ([]byte, error) {
+	requestHeader := make([]byte, 4)
+
+	copy(requestHeader, streamID[:])
+	binary.BigEndian.PutUint16(requestHeader[2:], requestID)
+
+	b, err := Marshal(requestBody)
+	if err != nil {
+		return nil, err
+	}
+
+	return append(requestHeader, b...), nil
+}
+
+// Marshal encodes structure to the bytes
+func Marshal(x interface{}) ([]byte, error) {
+	v := reflect.ValueOf(x)
+	if v.Kind() == reflect.Ptr {
+		v = v.Elem()
+	}
+
+	dataSize, err := calculateSizeForMarshaling(v)
+	if err != nil {
+		return nil, err
+	}
+
+	data := make([]byte, dataSize)
+	pos := 0
+	for i := 0; i < v.NumField() && err == nil; i++ {
+		field := v.Field(i)
+		fieldSize := 0
+		switch field.Kind() {
+		case reflect.Uint8:
+			fieldSize = 1
+			data[pos] = uint8(field.Uint())
+		case reflect.Uint16:
+			fieldSize = 2
+			binary.BigEndian.PutUint16(data[pos:pos+fieldSize], uint16(field.Uint()))
+		case reflect.Int32:
+			fieldSize = 4
+			binary.BigEndian.PutUint32(data[pos:pos+fieldSize], uint32(field.Int()))
+		case reflect.Int64:
+			fieldSize = 8
+			binary.BigEndian.PutUint64(data[pos:pos+fieldSize], uint64(field.Int()))
+		case reflect.Slice:
+			fieldSize = field.Len()
+			reflect.Copy(reflect.ValueOf(data[pos:pos+fieldSize]), field)
+		case reflect.Array:
+			fieldSize = field.Len()
+			reflect.Copy(reflect.ValueOf(data[pos:pos+fieldSize]), field)
+
+		default:
+			err = errors.Errorf("Cannot encode kind %s", field.Kind())
+		}
+		pos += fieldSize
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	return data, nil
+}
+
+// Unmarshal decodes data from byte slice
+func Unmarshal(data []byte, x interface{}) (err error) {
+	v := reflect.ValueOf(x)
+
+	if v.Kind() == reflect.Ptr {
+		v = v.Elem()
+	}
+	pos := 0
+
+	for i := 0; i < v.NumField() && err == nil; i++ {
+		field := v.Field(i)
+		fieldSize := 0
+		switch field.Kind() {
+		case reflect.Uint8:
+			fieldSize = 1
+			field.SetUint(uint64(data[pos]))
+		case reflect.Uint16:
+			fieldSize = 2
+			var value = binary.BigEndian.Uint16(data[pos : pos+fieldSize])
+			field.SetUint(uint64(value))
+		case reflect.Int32:
+			fieldSize = 4
+			var value = int32(binary.BigEndian.Uint32(data[pos : pos+fieldSize]))
+			field.SetInt(int64(value))
+		case reflect.Int64:
+			fieldSize = 8
+			var value = int64(binary.BigEndian.Uint64(data[pos : pos+fieldSize]))
+			field.SetInt(value)
+		case reflect.Slice:
+			bytes := data[pos:]
+			fieldSize = len(bytes)
+			field.SetBytes(bytes)
+		case reflect.Array:
+			fieldSize = field.Len()
+			reflect.Copy(field, reflect.ValueOf(data[pos:pos+fieldSize]))
+		default:
+			err = errors.Errorf("Cannot decode kind %s", field.Kind())
+		}
+		pos += fieldSize
+	}
+	return
+}
+
+func calculateSizeForMarshaling(v reflect.Value) (size int, err error) {
+	for i := 0; i < v.NumField() && err == nil; i++ {
+		field := v.Field(i)
+		switch field.Kind() {
+		case reflect.Uint8:
+			size++
+		case reflect.Uint16:
+			size += 2
+		case reflect.Int32:
+			size += 4
+		case reflect.Int64:
+			size += 8
+		case reflect.Array:
+			size += field.Len()
+		case reflect.Slice:
+			size += field.Len()
+		default:
+			err = errors.Errorf("Cannot decode kind %s", field.Kind())
+		}
+	}
+	return
+}
--- a/xrootd/encoder/encoder_test.go
+++ b/xrootd/encoder/encoder_test.go
@@ -1,0 +1,107 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package encoder // import "go-hep.org/x/hep/xrootd/encoder"
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go-hep.org/x/hep/xrootd/protocol"
+)
+
+type request struct {
+	Z int64
+	X int32
+	A uint8
+	C uint16
+	D [2]byte
+	Y []byte
+}
+
+type benchmarkRequest struct {
+	X   int32
+	A   uint8
+	A1  uint8
+	A2  uint8
+	A3  uint8
+	A4  int32
+	A5  int32
+	A6  int32
+	A7  int32
+	A8  int32
+	A9  int32
+	A10 int32
+	A11 int32
+	A12 int32
+	A13 int32
+	A14 int32
+	A16 int32
+	C   uint16
+	D   [10]byte
+	Z   [10]byte
+}
+
+type undecodable struct {
+	A float64
+}
+
+func TestMarshalRequest(t *testing.T) {
+	var requestID uint16 = 1337
+	var streamID = protocol.StreamID{42, 37}
+	expected := []byte{42, 37, 5, 57, 0, 0, 0, 0, 0, 0, 0, 7, 0, 0, 0, 1, 2, 0, 3, 6, 7, 11, 13}
+
+	actual, err := MarshalRequest(requestID, streamID, request{7, 1, 2, 3, protocol.StreamID{6, 7}, []byte{11, 13}})
+
+	assert.NoError(t, err)
+	assert.Equal(t, expected, actual)
+}
+
+func TestUnmarshal(t *testing.T) {
+	var expected = request{7, 1, 2, 3, protocol.StreamID{6, 7}, []byte{11, 13}}
+
+	var actual = &request{}
+	err := Unmarshal([]byte{0, 0, 0, 0, 0, 0, 0, 7, 0, 0, 0, 1, 2, 0, 3, 6, 7, 11, 13}, actual)
+
+	assert.Equal(t, expected, *actual)
+	assert.NoError(t, err)
+}
+
+func TestMarshalRequest_Undecodable(t *testing.T) {
+	var requestID uint16 = 1337
+	var streamID = protocol.StreamID{42, 37}
+
+	_, err := MarshalRequest(requestID, streamID, undecodable{1})
+
+	assert.Error(t, err)
+}
+
+func TestUnmarshal_Undecodable(t *testing.T) {
+	var actual = &undecodable{}
+
+	err := Unmarshal([]byte{0, 0, 0, 1, 2, 0, 3, 6, 7, 11, 13}, actual)
+
+	assert.Error(t, err)
+}
+
+func BenchmarkMarshal(b *testing.B) {
+	br := benchmarkRequest{}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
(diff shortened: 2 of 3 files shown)


xrootd/encoder/encoder.go
```diff
@@ -0,0 +1,144 @@
...
+	b, err := Marshal(requestBody)
+	if err != nil {
+		return nil, err
+	}
+
+	return append(requestHeader, b...), nil
+}
+
+// Marshal encodes structure to the bytes
+func Marshal(x interface{}) ([]byte, error) {
```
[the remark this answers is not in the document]
EgorMatirov: Thanks! I think I'll add it to https://github.com/go-hep/hep/issues/177 for further optimization. Is it ok?

xrootd/encoder/encoder.go
```diff
@@ -0,0 +1,144 @@
...
+	"encoding/binary"
+	"reflect"
+
+	"github.com/pkg/errors"
+	"go-hep.org/x/hep/xrootd/protocol"
+)
+
+// MarshalRequest encodes request body alongside with request and stream ids
+func MarshalRequest(requestID uint16, streamID protocol.StreamID, requestBody interface{}) ([]byte, error) {
+	requestHeader := make([]byte, 4)
```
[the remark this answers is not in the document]
EgorMatirov: Sorry, I'm not sure that I totally understand. Are you suggesting to preallocate an array for the whole request and not just header?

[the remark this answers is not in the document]
sbinet: Yes, it would be great to have something like:

```go
req := make([]byte, 4, n)
```

Where n isn't too big to not preallocate too much but big enough so most calls to MarshalRequest wouldn't need to copy the whole buffer over in the call to append.

This could be marked as a TODO :)

xrootd/encoder/encoder.go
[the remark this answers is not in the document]
sbinet: Sure.

xrootd/encoder/encoder.go
```diff
@@ -0,0 +1,178 @@
...
+		case reflect.Int32:
+			fieldSize = 4
+			binary.BigEndian.PutUint32(data[pos:pos+fieldSize], uint32(field.Int()))
+		case reflect.Int64:
+			fieldSize = 8
+			binary.BigEndian.PutUint64(data[pos:pos+fieldSize], uint64(field.Int()))
+		case reflect.Slice:
+			elementKind := field.Type().Elem().Kind()
+			if elementKind != reflect.Uint8 {
+				return nil, errors.Errorf("Cannot marshal slice of %s, only slices of uint8 are supported", elementKind)
```
sbinet: please modify that error message along the lines of my previous comment above.

xrootd/encoder/encoder.go
```diff
@@ -0,0 +1,178 @@
...
+			fieldSize = field.Len()
+			reflect.Copy(reflect.ValueOf(data[pos:pos+fieldSize]), field)
+		case reflect.Array:
+			elementKind := field.Type().Elem().Kind()
+			if elementKind != reflect.Uint8 {
+				return nil, errors.Errorf("Cannot marshal array of %s, only arrays of uint8 are supported", elementKind)
+			}
+
+			fieldSize = field.Len()
+			reflect.Copy(reflect.ValueOf(data[pos:pos+fieldSize]), field)
```
sbinet: ditto.
except that would probably look like:
```go
v := field.Slice(0, field.Len()).Interface().([]byte)
fieldSize = copy(data[pos:], v)
```
(untested)

xrootd/encoder/encoder.go
```diff
@@ -0,0 +1,178 @@
...
+// Supported types are: uint8, uint16, int32, int64, slices and arrays of uint8.
+func Unmarshal(data []byte, x interface{}) (err error) {
+	v := reflect.ValueOf(x)
+
+	if v.Kind() == reflect.Ptr {
+		v = v.Elem()
+	}
+
+	if v.Kind() != reflect.Struct {
+		return errors.Errorf("Cannot unmarshal %s, struct is expected", v.Kind())
```
sbinet: please update the error message.
also, what is really expected is a *pointer* to a struct.

xrootd/encoder/encoder.go
```diff
@@ -0,0 +1,178 @@
...
+		case reflect.Uint8:
+			fieldSize = 1
+			field.SetUint(uint64(data[pos]))
+		case reflect.Uint16:
+			fieldSize = 2
+			var value = binary.BigEndian.Uint16(data[pos : pos+fieldSize])
+			field.SetUint(uint64(value))
+		case reflect.Int32:
+			fieldSize = 4
+			var value = int32(binary.BigEndian.Uint32(data[pos : pos+fieldSize]))
```
sbinet: why do you add this `uint32 -> int32` conversion when you convert `value` to an `int64` just the line after?
is it really needed?

xrootd/encoder/encoder.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	data := make([]byte, dataSize)
+	pos := 0
+	for i := 0; i < v.NumField(); i++ {
+		field := v.Field(i)
+		fieldSize := 0
+		switch field.Kind() {
```
sbinet: instead of switching on the field `Kind`, I would directly `type-switch` on its interface value:

```go
switch fv := field.Interface().(type) {
case uint8:
    fieldSize = 1
    data[pos] = fv
case uint16:
    fieldSize = 2
    binary.BigEndian.PutUint16(data[pos:pos+fieldSize], fv))
...
case []byte:
    fieldSize = copy(data[pos:], fv)
default:
    if field.Kind() == reflect.Array {
        // handle it
    } else {
        errors.Errorf("encoder: can not Marshal invalid field type %T", fv)
    }
}
```

le

xrootd/encoder/encoder.go
```diff
@@ -0,0 +1,178 @@
...
+			fieldSize = 8
+			var value = int64(binary.BigEndian.Uint64(data[pos : pos+fieldSize]))
+			field.SetInt(value)
+		case reflect.Slice:
+			elementKind := field.Type().Elem().Kind()
+			if elementKind != reflect.Uint8 {
+				return errors.Errorf("Cannot unmarshal slice of %s, only slices of uint8 are supported", elementKind)
+			}
+
+			bytes := data[pos:]
```
sbinet: hum... if I am not mistaken...
this assumes that a slice is always the last field of a struct.
this also assumes we'll never have 2 fields that are slices.

is this really the case?
naively, I would have assumed that the following C++ payload:
```c
struct Request {
    int32_t msgLen;
    char *msg;
};
```

would translate into:
```go
type Request struct {
    Msg []byte
}
```

with the following layout on the wire: `[4 bytes: msgLen][msgLen bytes: msg]`.

so the `Marshal` function would - when encountering a `[]byte` - first write the length of the slice on the wire, and then copy the whole slice.
similarly, the `Unmarshal` function would - when encountering a `[]byte` field - first read a `uint32` (the length) and read as many bytes.

but maybe I am missing something?

xrootd/protocol/streamID.go
```diff
@@ -0,0 +1,7 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+package protocol // import "go-hep.org/x/hep/xrootd/protocol"
+
+// StreamID is the binary identifier associated with request stream
```
sbinet: `s/with request/with a request/`

xrootd/encoder/encoder_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+func TestUnmarshal_Undecodable(t *testing.T) {
+	var actual = &undecodable{}
+
+	err := Unmarshal([]byte{0, 0, 0, 1, 2, 0, 3, 6, 7, 11, 13}, actual)
+
+	assert.Error(t, err)
+}
+
+func BenchmarkMarshal(b *testing.B) {
```
sbinet: could you add benchmark numbers as part of the commit message?
just to get an idea.

xrootd/encoder/encoder_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"go-hep.org/x/hep/xrootd/protocol"
+)
+
+type request struct {
+	Z int64
+	X int32
+	A uint8
+	C uint16
+	D [2]byte
+	Y []byte
```
sbinet: I believe `Marshal` and `Unmarshal` would break if you were to add another field of type `[]byte` here.

xrootd/encoder/encoder_test.go
```diff
@@ -0,0 +1,107 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package encoder // import "go-hep.org/x/hep/xrootd/encoder"
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
sbinet: I must admit I am not a huge fan of `testify/assert`.
could we just use `testing` and "inline" code to express what we expect?

xrootd/encoder/encoder_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+func TestMarshalRequest(t *testing.T) {
+	var requestID uint16 = 1337
+	var streamID = protocol.StreamID{42, 37}
+	expected := []byte{42, 37, 5, 57, 0, 0, 0, 0, 0, 0, 0, 7, 0, 0, 0, 1, 2, 0, 3, 6, 7, 11, 13}
+
+	actual, err := MarshalRequest(requestID, streamID, request{7, 1, 2, 3, protocol.StreamID{6, 7}, []byte{11, 13}})
+
+	assert.NoError(t, err)
+	assert.Equal(t, expected, actual)
```
sbinet: please replace with:
```go
if !reflect.DeepEqual(got, want) {
    t.Fatalf("invalid marshal request\ngot = %v\nwant= %v\n", got, want)
}
```

xrootd/encoder/encoder_test.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+type undecodable struct {
+	A float64
+}
+
+func TestMarshalRequest(t *testing.T) {
+	var requestID uint16 = 1337
+	var streamID = protocol.StreamID{42, 37}
+	expected := []byte{42, 37, 5, 57, 0, 0, 0, 0, 0, 0, 0, 7, 0, 0, 0, 1, 2, 0, 3, 6, 7, 11, 13}
```
sbinet: `s/expected/want/`

xrootd/encoder/encoder_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+	actual, err := MarshalRequest(requestID, streamID, request{7, 1, 2, 3, protocol.StreamID{6, 7}, []byte{11, 13}})
```
sbinet: `s/actual/got/`

xrootd/encoder/encoder_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+	assert.NoError(t, err)
```
sbinet: please replace with:

```go
if err != nil {
    t.Fatalf("could not marshal request: %v", err)
}
```

xrootd/encoder/encoder.go
```diff
@@ -0,0 +1,178 @@
...
+	pos := 0
+
+	for i := 0; i < v.NumField() && err == nil; i++ {
+		field := v.Field(i)
+		fieldSize := 0
+		switch field.Kind() {
+		case reflect.Uint8:
+			fieldSize = 1
+			field.SetUint(uint64(data[pos]))
+		case reflect.Uint16:
```
sbinet: what about `Int8`, `Int16`, `Uint32` and `Uint64` ?

xrootd/encoder/encoder.go
```diff
@@ -0,0 +1,178 @@
...
+			reflect.Copy(field, reflect.ValueOf(data[pos:pos+fieldSize]))
+		default:
+			err = errors.Errorf("Cannot unmarshal kind %s", field.Kind())
+		}
+		pos += fieldSize
+	}
+	return
+}
+
+func calculateSizeForMarshaling(v reflect.Value) (size int, err error) {
```
sbinet: I would just name it `sizeOf`

xrootd/encoder/encoder.go
EgorMatirov: If I have done it correctly, it is 1.5x - 2x slower: 
1130 ns/op vs 1888 ns/op for Marshal and 
733 ns/op vs 1780 ns/op for Unmarshal. 

https://github.com/EgorMatirov/hep/commit/104d3630761afefc267392a1c9fefa70ef2fd1b4

xrootd/encoder/encoder.go
EgorMatirov: Most of time it's so, but on some requests / responses length is not specified before bytes.
For example, read with readahead, see http://xrootd.org/doc/dev45/XRdv310.pdf (page 99) - rvec's length is passed indirectly via alen.

Also, in responses, for example read response, such scheme exists:
// Start of the header
kXR_char streamid[2]
kXR_unt16 status
kXR_int32 dlen  // length of the response
// End of the header
// Start of data that will be passed to the unmarshal
kXR_char data[dlen]
// End of data that will be passed to the unmarshal

However, I don't remember any requests/responses that have 2 slices of bytes. I'll look into specification once more and check if it's truth.

Any suggestions are welcome! :)

xrootd/encoder/encoder.go
EgorMatirov: Only uint8, uint16, int32 and int64 are supported by specification: http://xrootd.org/doc/dev45/XRdv310.pdf (page 6)

xrootd/encoder/encoder.go
EgorMatirov: I have got `panic: reflect.Value.Slice: slice of unaddressable array` with this snippet. Reduced duplication anyway, thanks.

sbinet: ah... right.
something like this could work:
```go
case reflect.Array:
    // Create slice from array
    var n = reflect.New(v.Type())
    n.Elem().Set(v)
    v = reflect.Indirect(n).Slice(0, v.Len())
    fallthrough
case reflect.Slice:
```
(but it copies the whole array...)

xrootd/encoder/encoder.go
```diff
@@ -0,0 +1,169 @@
...
+
+	b, err := Marshal(requestBody)
+	if err != nil {
+		return nil, err
+	}
+
+	return append(requestHeader, b...), nil
+}
+
+// Marshal encodes structure to the bytes following the XRootd protocol specification.
```
sbinet: `s/Marshal encodes structure to the bytes/Marshal encodes a structure value into a slice of bytes/`

xrootd/encoder/encoder.go
```diff
@@ -0,0 +1,169 @@
...
+// Fields are encoded in the same order as they are specified in the struct definition.
+// Each field is encoded in network byte order (BigEndian) without any alignment or padding.
+// Slices and arrays of uint8 are binary copied without further encoding.
+// Supported types are: uint8, uint16, int32, int64, slices and arrays of uint8.
+func Marshal(x interface{}) ([]byte, error) {
+	v := reflect.ValueOf(x)
+	if v.Kind() == reflect.Ptr {
```
sbinet: this could be written using `reflect.Indirect`:
```go
v = reflect.Indirect(v)
```

xrootd/encoder/encoder.go
sbinet: ok. that's interesting...
let's use `reflect.Value.Kind` then (it should be a moot point once we get per-type (un)marshaling)

xrootd/encoder/encoder.go
sbinet: ok.

sbinet:
ah, one more thing:
once you've applied the 2 little modifications I've suggested in my latest review, could you squash everything into 1 commit (w/o forgetting to add the benchmark numbers to that one commit message) ?

thanks a lot again.

EgorMatirov:
Got failed CI due to:
```
=== RUN   TestH1D
--- PASS: TestH1D (0.15s)
=== RUN   TestH1DtoPDF
--- FAIL: TestH1DtoPDF (2.87s)
	hplot_test.go:47: error: "testdata/h1d_plot.pdf" differ with reference file
=== RUN   TestH2D
--- PASS: TestH2D (0.19s)
=== RUN   TestH2DABCD
--- PASS: TestH2DABCD (0.11s)
=== RUN   TestSubPlot
--- PASS: TestSubPlot (0.17s)
=== RUN   TestDiffPlot
--- PASS: TestDiffPlot (0.14s)
=== RUN   TestLatexPlot
--- PASS: TestLatexPlot (0.00s)
=== RUN   TestS2D
--- PASS: TestS2D (0.13s)
=== RUN   TestScatter2DWithErrorBars
--- PASS: TestScatter2DWithErrorBars (0.11s)
=== RUN   TestTiledPlot
--- PASS: TestTiledPlot (0.20s)
FAIL
FAIL	go-hep.org/x/hep/hplot	4.094s
```
which wasn't changed by this commit...

sbinet:
yes, `gonum/plot` was updated to get a new PDF engine and it produces slightly different PDF files.
just rebase on top of the latest `master` of `go-hep/hep` (ie: commit 117a8d7) and it should be good.

sbinet:
thanks!

PR #184, opened 2018-05-05
xrootd: add mux

EgorMatirov:
Add mux package which manages channels by their ids.
Is part of #170, #179 and is built on top of #183.

I'm still not sure how to make several PRs and keep them "stacked". I'm sorry if I messed it up.

### DIFF
--- a/xrootd/mux/mux.go
+++ b/xrootd/mux/mux.go
@@ -1,0 +1,126 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mux // import "go-hep.org/x/hep/xrootd/mux"
+
+import (
+	"math"
+	"sync"
+
+	"github.com/pkg/errors"
+	"go-hep.org/x/hep/xrootd/protocol"
+)
+
+// ServerResponse contains slice of bytes Data representing data from
+// XRootD server response (see XRootD protocol specification) and
+// Err representing error received from server or occurred
+// during response decoding.
+type ServerResponse struct {
+	Data []byte
+	Err  error
+}
+
+type dataSendChan chan<- ServerResponse
+type DataRecvChan <-chan ServerResponse
+
+const streamIDPartSize = math.MaxUint8
+const streamIDPoolSize = streamIDPartSize * streamIDPartSize
+
+// Mux manages channels by their ids.
+// Basically, it's a map[StreamID] chan<-ServerResponse
+// with methods to claim, free and pass data to a specific channel by id.
+type Mux struct {
+	mu          sync.Mutex
+	dataWaiters map[protocol.StreamID]dataSendChan
+	freeIDs     chan uint16
+	quit        chan struct{}
+}
+
+// New creates a new Mux
+func New() *Mux {
+	const freeIDsBufferSize = 32 // 32 is completely arbitrary ATM and should be refined based on real use cases.
+
+	m := Mux{
+		dataWaiters: make(map[protocol.StreamID]dataSendChan),
+		freeIDs:     make(chan uint16, freeIDsBufferSize),
+		quit:        make(chan struct{}),
+	}
+
+	go func() {
+		var i uint16 = 0
+		for {
+			select {
+			case m.freeIDs <- i:
+				i = (i + 1) % streamIDPoolSize
+			case <-m.quit:
+				return
+			}
+		}
+	}()
+
+	return &m
+}
+
+// Close closes the Mux
+func (m *Mux) Close() {
+	close(m.quit)
+}
+
+// Claim searches for unclaimed id and returns corresponding channel
+func (m *Mux) Claim() (protocol.StreamID, DataRecvChan, error) {
+	ch := make(chan ServerResponse)
+
+	for {
+		id := <-m.freeIDs
+		streamId := protocol.StreamID{byte(id >> 8), byte(id)}
+
+		m.mu.Lock()
+		if _, claimed := m.dataWaiters[streamId]; claimed { // Skip id if it was already claimed manually via ClaimWithID
+			m.mu.Unlock()
+			continue
+		}
+
+		m.dataWaiters[streamId] = ch
+		m.mu.Unlock()
+		return streamId, ch, nil
+	}
+}
+
+// ClaimWithID checks if id is unclaimed and returns the corresponding channel in case of success
+func (m *Mux) ClaimWithID(id protocol.StreamID) (DataRecvChan, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	ch := make(chan ServerResponse, 1)
+
+	if _, claimed := m.dataWaiters[id]; claimed {
+		return nil, errors.Errorf("mux: channel with id %s is already claimed", id)
+	}
+
+	m.dataWaiters[id] = ch
+
+	return ch, nil
+}
+
+// Unclaim marks channel with specified id as unclaimed
+func (m *Mux) Unclaim(id protocol.StreamID) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	close(m.dataWaiters[id])
+	delete(m.dataWaiters, id)
+}
+
+// SendData sends data to channel with specific id
+func (m *Mux) SendData(id protocol.StreamID, data ServerResponse) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if _, ok := m.dataWaiters[id]; !ok {
+		return errors.Errorf("mux: cannot find data waiter for id %s", id)
+	}
+
+	m.dataWaiters[id] <- data
+
+	return nil
+}
--- a/xrootd/mux/mux_test.go
+++ b/xrootd/mux/mux_test.go
@@ -1,0 +1,198 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mux // import "go-hep.org/x/hep/xrootd/mux"
+
+import (
+	"reflect"
+	"testing"
+
+	"go-hep.org/x/hep/xrootd/protocol"
+)
+
+func TestMux_Claim(t *testing.T) {
+	m := New()
+	defer m.Close()
+	claimedIds := map[protocol.StreamID]bool{}
+
+	for i := 0; i < streamIDPoolSize; i++ {
+		id, channel, err := m.Claim()
+
+		if err != nil {
+			t.Fatalf("could not Claim streamID: %v", err)
+		}
+
+		if channel == nil {
+			t.Fatal("channel should not be nil")
+		}
+
+		if claimedIds[id] {
+			t.Fatalf("should not claim id %s, which was already claimed", id)
+		}
+
+		claimedIds[id] = true
+	}
+}
+
+func TestMux_Claim_AfterUnclaim(t *testing.T) {
+	m := New()
+	defer m.Close()
+	claimedIds := map[protocol.StreamID]bool{}
+
+	for i := 0; i < streamIDPoolSize; i++ {
+		id, _, _ := m.Claim()
+		claimedIds[id] = true
+	}
+
+	wantID := protocol.StreamID{13, 14}
+	m.Unclaim(wantID)
+
+	gotID, channel, err := m.Claim()
+
+	if err != nil {
+		t.Fatalf("could not Claim streamID: %v", err)
+	}
+
+	if channel == nil {
+		t.Fatal("channel should not be nil")
+	}
+
+	if !reflect.DeepEqual(gotID, wantID) {
+		t.Fatalf("invalid claim\ngot = %v\nwant = %v", gotID, wantID)
+	}
+}
+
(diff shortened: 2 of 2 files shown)


sbinet:
that's alright, github's interface isn't clever enough to deal with stacked PRs...
that's the best we can do for now :)

xrootd/streammanager/streammanager.go
```diff
@@ -0,0 +1,113 @@
...
+	"math"
+	"sync"
+
+	"github.com/pkg/errors"
+	"go-hep.org/x/hep/xrootd/protocol"
+)
+
+type ServerResponse struct {
+	Data  []byte
+	Error error
```
sbinet: please call it `Err` instead of `Error` (so we can eventually make it implement the `error` interface)

xrootd/streammanager/streammanager.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+type dataSendChannel chan<- *ServerResponse
+type DataReceiveChannel <-chan *ServerResponse
```
sbinet: `s/DataReceiveChannel/DataRecvChan/`

xrootd/streammanager/streammanager.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/pkg/errors"
+	"go-hep.org/x/hep/xrootd/protocol"
+)
+
+type ServerResponse struct {
+	Data  []byte
+	Error error
+}
+
+type dataSendChannel chan<- *ServerResponse
```
sbinet: `s/dataSendChannel/dataSendChan/`

xrootd/streammanager/streammanager.go
```diff
@@ -0,0 +1,113 @@
...
+	freeIds     chan protocol.StreamID
+	mutex       sync.Mutex
+}
+
+// New creates new StreamManager
+func New() *StreamManager {
+	sm := StreamManager{
+		make(map[protocol.StreamID]dataSendChannel),
+		make(chan protocol.StreamID, streamIDPoolSize),
+		sync.Mutex{},
```
sbinet: no need for this `sync.Mutex{}`, we can use the default value.

xrootd/streammanager/streammanager.go
```diff
@@ -0,0 +1,113 @@
...
+
+const streamIDPartSize = math.MaxUint8
+const streamIDPoolSize = streamIDPartSize * streamIDPartSize
+
+// StreamManager manages channels by their ids.
+// Basically, it's a map[StreamID] chan<-*ServerResponse
+// with methods to claim, free and pass data to specific channel by id.
+type StreamManager struct {
+	dataWaiters map[protocol.StreamID]dataSendChannel
+	freeIds     chan protocol.StreamID
```
sbinet: `go vet` will probably complain that it should be named `freeIDs`.

xrootd/streammanager/streammanager.go
```diff
@@ -0,0 +1,113 @@
...
+
+// New creates new StreamManager
+func New() *StreamManager {
+	sm := StreamManager{
+		make(map[protocol.StreamID]dataSendChannel),
+		make(chan protocol.StreamID, streamIDPoolSize),
+		sync.Mutex{},
+	}
+
+	for firstByte := uint16(0); firstByte < streamIDPartSize; firstByte++ {
```
sbinet: wouldn't it be more efficient to allocate a global `[streamIDPoolSize]protocol.StreamID` array and launch a goroutine that would iterate over that.
the goroutine would send the current index to a non-buffered channel.
the stream-manager would draw from that non-buffered channel as needed.

this goroutine would need to be properly collected when we are done with the StreamManager.

xrootd/streammanager/streammanager.go
```diff
@@ -0,0 +1,113 @@
...
+		select {
+		case id = <-sm.freeIds:
+			if _, claimed := sm.dataWaiters[id]; claimed { // Skip id if it was already claimed manually via ClaimWithID
+				continue
+			}
+
+			sm.dataWaiters[id] = bidirectionalChannel
+			channel = bidirectionalChannel
+		default:
+			err = errors.New("Cannot claim free channel")
```
sbinet: please modify the error message to follow conventions.

xrootd/streammanager/streammanager.go
```diff
@@ -0,0 +1,113 @@
...
+			sm.freeIds <- protocol.StreamID{byte(firstByte), byte(secondByte)}
+		}
+	}
+
+	return &sm
+}
+
+// Claim searches for unclaimed id and returns corresponding channel
+func (sm *StreamManager) Claim() (id protocol.StreamID, channel DataReceiveChannel, err error) {
+	bidirectionalChannel := make(chan *ServerResponse, 1)
```
sbinet: `s/bidirectionalChannel/ch/`
or `srvCh`, or `respCh` ?

xrootd/streammanager/streammanager.go
```diff
@@ -0,0 +1,113 @@
...
+	for firstByte := uint16(0); firstByte < streamIDPartSize; firstByte++ {
+		for secondByte := uint16(0); secondByte < streamIDPartSize; secondByte++ {
+			sm.freeIds <- protocol.StreamID{byte(firstByte), byte(secondByte)}
+		}
+	}
+
+	return &sm
+}
+
+// Claim searches for unclaimed id and returns corresponding channel
```
sbinet: `s/returns/returns the/`

xrootd/streammanager/streammanager.go
sbinet: just curious but: why is this channel buffered?

xrootd/streammanager/streammanager.go
```diff
@@ -0,0 +1,113 @@
...
+func (sm *StreamManager) Claim() (id protocol.StreamID, channel DataReceiveChannel, err error) {
+	bidirectionalChannel := make(chan *ServerResponse, 1)
+	sm.mutex.Lock()
```
sbinet: IIUC, the mutex protects `dataWaiters`.
perhaps we could move the lock closer to the modification/use of `dataWaiters` to reduce the "critical section"

xrootd/streammanager/streammanager.go
```diff
@@ -0,0 +1,113 @@
...
+const streamIDPartSize = math.MaxUint8
+const streamIDPoolSize = streamIDPartSize * streamIDPartSize
+
+// StreamManager manages channels by their ids.
+// Basically, it's a map[StreamID] chan<-*ServerResponse
+// with methods to claim, free and pass data to specific channel by id.
+type StreamManager struct {
+	dataWaiters map[protocol.StreamID]dataSendChannel
+	freeIds     chan protocol.StreamID
+	mutex       sync.Mutex
```
sbinet: the convention in Go for mutexes is to:
- name them just `mu`
- put them above the field(s) they are supposed to guard

xrootd/streammanager/streammanager.go
```diff
@@ -0,0 +1,113 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package streammanager // import "go-hep.org/x/hep/xrootd/streammanager"
```
sbinet: I am not too fond about the name of the package.

trying to come up with a (IMO) better name... basically, the main type of that package "routes" server responses, right?
something akin to [Gorilla's router/multiplexer](https://godoc.org/github.com/gorilla/mux) ?

I'd be tempted to rename `streammanager` into `mux`, `router` or `dispatch`.
what do you think?

xrootd/streammanager/streammanager.go
EgorMatirov: I thought about the case when `SendData` will be deadlocked due to unreceived data from channel. But that's highly unlikely now because all reads from this channel will be done in one method of `Client`. Anyway, we can't specify buffer size (if we are not going to use some huge number for that) because server response can be split in many messages and client doesn't know how many responses expect. I'll make it unbuffered.

xrootd/streammanager/streammanager.go
EgorMatirov: Can you please explain how can I collect goroutine?

Btw, when I make a goroutine like this:
```
go func() {
    i := 0
    for {
        sm.freeIDs <- i
        i = (i + 1) % streamIDPoolSize
    }
}()
```
where `freeIDs` is `make(chan int, streamIDPoolSize)`
it gives ~530 ns/op for Claim and Unclaim in benchmark comparing to  440 ns/op for current implementation.

sbinet: also, I am bit concerned with memory usage of this `make(chan protocol.StreamID, streamIDPoolSize)`
where: `streamIDPoolSize = 255*255 = 65025`
adding the size of each element of that buffered channel, we get: `65025 * 2 = 130050` bytes.

and that is per `StreamManager` value.

couldn't we find a way where we don't have to materialize these big structures?
(just working off simple integers `int` (or `int16`) and automatically create `protocol.StreamID` from/to these? a bit of `<<` and `>>` magic sounds like it would do.)

at the very least, we should mark all of these big structures with a clear `TODO` (+ an issue) so we don't forget about this.

xrootd/mux/mux.go
```diff
@@ -0,0 +1,113 @@
...
+		default:
+			err = errors.New("streammanager: cannot Claim free channel")
+		}
+	}
+
+	return
+}
+
+// ClaimWithID checks if id is unclaimed and returns the corresponding channel in case of success
+func (m *Mux) ClaimWithID(id protocol.StreamID) (channel DataRecvChan, err error) {
```
sbinet: ditto, wrt named returned values.

IMHO, this reads as well (if not better :P):
```go
func (m *Mux) ClaimWithID(id protocol.StreamID) (DataRecvChan, error) {
    m.mu.Lock()
    defer m.mu.Unlock()
    ch := make(chan *ServerResponse, 1)

    if _, claimed := m.dataWaiters[id]; claimed {
        return nil, errors.Errorf(...)
    }

    m.dataWaiters[id] = ch
    return ch, nil
}
```

xrootd/mux/mux.go
```diff
@@ -0,0 +1,124 @@
...
+const streamIDPartSize = math.MaxUint8
+const streamIDPoolSize = streamIDPartSize * streamIDPartSize
+
+// Mux manages channels by their ids.
+// Basically, it's a map[StreamID] chan<-*ServerResponse
+// with methods to claim, free and pass data to a specific channel by id.
+type Mux struct {
+	mu          sync.Mutex
+	dataWaiters map[protocol.StreamID]dataSendChan
+	freeIDs     chan uint16
```
EgorMatirov: If freeIDs is unbuffered, then benchmark result is ~700 ns /op vs ~350 ns /op before.
If it's buffered (with buffer at least 20 or 30, for example) then benchmark result is close to ~350 ns/op.

Should I make it buffered? If so, which value to choose?

sbinet: let's use the magic number of `32` then (with a comment saying it's completely arbitrary at the moment and that value should be refined based on real use cases.)

xrootd/mux/mux.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+type dataSendChan chan<- ServerResponse
+type DataRecvChan <-chan ServerResponse
+
+const streamIDPartSize = math.MaxUint8
+const streamIDPoolSize = streamIDPartSize * streamIDPartSize
+
+// Mux manages channels by their ids.
+// Basically, it's a map[StreamID] chan<-*ServerResponse
```
sbinet: `s/*ServerResponse/ServerResponse/`

xrootd/mux/mux_test.go
```diff
@@ -0,0 +1,195 @@
...
+		for {
+			select {
+			case <-ch:
+			case <-done:
+				return
+			}
+		}
+	}()
+
+	for i := 0; i < b.N; i++ {
```
sbinet: perhaps add `b.ResetTimer()` just before this for-loop so we really just benchmark `SendData` ?

xrootd/mux/mux_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+	if err == nil {
+		t.Fatal("should not be able to SenData when id is unclaimed")
+	}
+}
+
+func BenchmarkMux_Claim(b *testing.B) {
+	m := New()
+	defer m.Close()
+	for i := 0; i < b.N; i++ {
```
sbinet: perhaps add `b.ResetTimer()` just before this for-loop so we really just benchmark `Claim` ?

xrootd/mux/mux_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+func TestMux_ClaimWithID_WhenIDIsTakenByClaim(t *testing.T) {
+	m := New()
+	defer m.Close()
+	id, _, _ := m.Claim()
+
+	_, err := m.ClaimWithID(id)
+
+	if err == nil {
+		t.Fatal("should not be able to ClaimWithId when that id is already claimed")
```
sbinet: `s/ClaimWithId/ClaimWithID/`

sbinet:
ah! I didn't get a notification that you did squash everything :) (bad github! usually, feel free to write "PTAL" to notify people :))

PR #189, opened 2018-05-15
xrootd/client: first import

EgorMatirov:
Add client package which allows to send requests and receive responses from XRootD server.
Add handshake request which performs an initial handshake.

Benchmark result is 150 ms/op which equals 2 * ping to the XRootD server.

Is part of https://github.com/go-hep/hep/issues/170.
I have extracted the client-related part from https://github.com/go-hep/hep/pull/179 to make it easier to review core component.

Specific requests will come as separated PR if it's ok and makes the review process easier. :)

### DIFF
--- a/xrootd/client/client.go
+++ b/xrootd/client/client.go
@@ -1,0 +1,175 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+/*
+Package client implements the XRootD client following protocol from http://xrootd.org.
+
+The NewClient function connects to a server:
+
+	ctx := context.Background()
+
+	client, err := NewClient(ctx, *Addr)
+	if err != nil {
+		// handle error
+	}
+
+	// ...
+
+	if err := client.Close(); err != nil {
+		// handle error
+	}
+*/
+package client // import "go-hep.org/x/hep/xrootd/client"
+
+import (
+	"context"
+	"io"
+	"net"
+
+	"go-hep.org/x/hep/xrootd/encoder"
+	"go-hep.org/x/hep/xrootd/mux"
+	"go-hep.org/x/hep/xrootd/protocol"
+)
+
+// A Client to xrootd server which allows to send requests and receive responses.
+// Concurrent requests are supported.
+// Zero value is invalid, Client should be instantiated using NewClient.
+type Client struct {
+	cancel          context.CancelFunc
+	conn            net.Conn
+	mux             *mux.Mux
+	protocolVersion int32
+}
+
+// NewClient creates a new xrootd client that connects to the given address.
+// When the context expires, a response handling is stopped, however, it is
+// necessary to call Cancel to correctly free resources.
+func NewClient(ctx context.Context, address string) (*Client, error) {
+	ctx, cancel := context.WithCancel(ctx)
+
+	var d net.Dialer
+	conn, err := d.DialContext(ctx, "tcp", address)
+	if err != nil {
+		cancel()
+		return nil, err
+	}
+
+	client := &Client{cancel, conn, mux.New(), 0}
+
+	go client.consume(ctx)
+
+	if err := client.handshake(ctx); err != nil {
+		client.Close()
+		return nil, err
+	}
+
+	return client, nil
+}
+
+// Close closes the connection. Any blocked operation will be unblocked and return error.
+func (client *Client) Close() error {
+	client.cancel()
+
+	client.mux.Close()
+	return client.conn.Close()
+}
+
+func (client *Client) consume(ctx context.Context) {
+	var header protocol.ResponseHeader
+	var headerBytes = make([]byte, protocol.ResponseHeaderLength)
+
+	for {
+		select {
+		case <-ctx.Done():
+			// TODO: Should wait for active requests to be completed?
+			return
+		default:
+			if _, err := io.ReadFull(client.conn, headerBytes); err != nil {
+				if ctx.Err() != nil {
+					// something happened to the context.
+					// ignore this error.
+					continue
+				}
+				panic(err)
+				// TODO: handle EOF by redirection as specified at http://xrootd.org/doc/dev45/XRdv310.pdf, page 11
+			}
+
+			if err := encoder.Unmarshal(headerBytes, &header); err != nil {
+				if ctx.Err() != nil {
+					// something happened to the context.
+					// ignore this error.
+					continue
+				}
+				panic(err)
+				// TODO: should redirect in case if is not possible to decode a header as well?
+			}
+
+			resp := mux.ServerResponse{Data: make([]byte, header.DataLength)}
+			if _, err := io.ReadFull(client.conn, resp.Data); err != nil {
+				if ctx.Err() != nil {
+					// something happened to the context.
+					// ignore this error.
+					continue
+				}
+				resp.Err = err
+			} else if header.Status != protocol.Ok {
+				resp.Err = header.Error(resp.Data)
+			}
+
+			if err := client.mux.SendData(header.StreamID, resp); err != nil {
+				if ctx.Err() != nil {
+					// something happened to the context.
+					// ignore this error.
+					continue
+				}
+				panic(err)
+				// TODO: should we just ignore responses to unclaimed stream IDs?
+			}
+
+			if header.Status != protocol.OkSoFar {
+				client.mux.Unclaim(header.StreamID)
+			}
+		}
+	}
+}
+
+func (client *Client) send(ctx context.Context, responseChannel mux.DataRecvChan, request []byte) ([]byte, error) {
+	if _, err := client.conn.Write(request); err != nil {
+		return nil, err
+	}
+
+	more := true
+	var resp mux.ServerResponse
+	var data []byte
+
+	for more {
+		select {
+		case resp, more = <-responseChannel:
+			data = append(data, resp.Data...)
+			if resp.Err != nil {
+				return nil, resp.Err
+			}
+		case <-ctx.Done():
+			if err := ctx.Err(); err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	return data, nil
+}
+
+func (client *Client) call(ctx context.Context, requestID uint16, request interface{}) ([]byte, error) {
+	streamID, responseChannel, err := client.mux.Claim()
+	if err != nil {
+		return nil, err
+	}
+
+	data, err := encoder.MarshalRequest(requestID, streamID, request)
+	if err != nil {
+		return nil, err
+	}
+
+	return client.send(ctx, responseChannel, data)
+}
--- a/xrootd/client/client_test.go
+++ b/xrootd/client/client_test.go
@@ -1,0 +1,57 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build xrootd_test_with_server
+
+package client // import "go-hep.org/x/hep/xrootd/client"
+
+import (
+	"context"
+	"testing"
+)
+
+var testClientAddrs = []string{"0.0.0.0:9001"}
+
+func TestNewClient(t *testing.T) {
+	for _, addr := range testClientAddrs {
+		t.Run(addr, func(t *testing.T) {
+			testNewClient(t, addr)
+		})
+	}
+}
+
+func testNewClient(t *testing.T, addr string) {
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	client, err := NewClient(ctx, addr)
+	if err != nil {
+		t.Fatalf("could not create client: %v", err)
+	}
+
+	if err := client.Close(); err != nil {
+		t.Fatalf("could not close client: %v", err)
+	}
+}
+
+func BenchmarkNewClient(b *testing.B) {
+	for _, addr := range testClientAddrs {
+		b.Run(addr, func(b *testing.B) {
+			benchmarkNewClient(b, addr)
+		})
+	}
+}
+
+func benchmarkNewClient(b *testing.B, addr string) {
+	for i := 0; i < b.N; i++ {
+		client, err := NewClient(context.Background(), addr)
+		if err != nil {
+			b.Fatalf("could not create client: %v", err)
+		}
+		if err := client.Close(); err != nil {
+			b.Fatalf("could not close client: %v", err)
+		}
+	}
+}
--- a/xrootd/protocol/handshake/handshake.go
+++ b/xrootd/protocol/handshake/handshake.go
@@ -1,0 +1,39 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package handshake contains the structures describing request and response
+// for handshake request (see XRootD specification).
+package handshake // import "go-hep.org/x/hep/xrootd/protocol/handshake"
+
+// ServerType is the general server type kept for compatibility
+// with 2.0 protocol version (see xrootd protocol specification v3.1.0, p. 5).
+type ServerType int32
+
+const (
+	// LoadBalancingServer indicates whether this is a load-balancing server.
+	LoadBalancingServer ServerType = iota
+	// DataServer indicates whether this is a data server.
+	DataServer ServerType = iota
+)
+
+// Response is a response for the handshake request,
+// which contains protocol version and server type.
+type Response struct {
+	ProtocolVersion int32
+	ServerType      ServerType
+}
+
+// Request holds the handshake request parameters.
+type Request struct {
+	Reserved1 int32
+	Reserved2 int32
+	Reserved3 int32
+	Reserved4 int32
+	Reserved5 int32
+}
+
+// NewRequest forms a Request that comply with the XRootD protocol v3.1.0.
+func NewRequest() Request {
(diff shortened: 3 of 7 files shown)


xrootd/client/client.go
```diff
@@ -0,0 +1,141 @@
...
+			// TODO: wait for active requests to be completed and handle errors from client.conn.Close().
+			client.conn.Close()
+			client.mux.Close()
+			return
+		default:
+		}
+	}
+}
+
+func extractError(header *responseHeader, data []byte) error {
```
sbinet: perhaps this could be a method of `responseHeader` ?
```go
func (hdr responseHeader) error(data []byte) error { ... }
```

the decoding of the error from the `[]byte` could even (perhaps) be put inside a method on `protocol.ServerError`.

xrootd/client/client.go
```diff
@@ -0,0 +1,141 @@
...
+	if header.Status == protocol.Error {
+		code := int32(binary.BigEndian.Uint32(data[0:4]))
+		message := string(data[4 : len(data)-1]) // Skip \0 character at the end
+
+		return protocol.ServerError{code, message}
+	}
+	return nil
+}
+
+func (client *Client) callWithBytesAndResponseChannel(ctx context.Context, responseChannel mux.DataRecvChan, request []byte) ([]byte, error) {
```
sbinet: not fond of this super long method name :)
what about `send` ?

xrootd/client/client.go
```diff
@@ -0,0 +1,141 @@
...
+	if _, err := client.conn.Write(request); err != nil {
+		return nil, err
+	}
+
+	more := true
+	var serverResponse mux.ServerResponse
```
sbinet: in the context of the client, I don't think there's an ambiguity as to from which/whom a response is (ie: the server)
I'd be tempted to just call this `resp` or `response`:
```go
var resp mux.ServerResponse
```

xrootd/client/client.go
```diff
@@ -0,0 +1,141 @@
...
+	return client, nil
+}
+
+func (client *Client) consume(ctx context.Context) {
+	var header = &responseHeader{}
+	var headerBytes = make([]byte, responseHeaderSize)
+
+	for {
+		if _, err := io.ReadFull(client.conn, headerBytes); err != nil {
+			// TODO: handle EOF by redirection as specified at http://xrootd.org/doc/dev45/XRdv310.pdf, page 11
```
sbinet: add a panic here? so we don't swallow errors?

xrootd/client/client.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+
+		if err := encoder.Unmarshal(headerBytes, header); err != nil {
+			// TODO: should redirect in case if is not possible to decode a header as well?
```
sbinet: here as well? add a panic here? so we don't swallow errors?

xrootd/client/client.go
```diff
@@ -0,0 +1,141 @@
...
+type responseHeader struct {
+	StreamID   protocol.StreamID
+	Status     uint16
+	DataLength int32
+}
+
+// NewClient creates a client to xrootd server at address.
+// ctx defines a lifetime of the client, once it is done response handling is stopped.
+func NewClient(ctx context.Context, address string) (*Client, error) {
+	conn, err := net.Dial("tcp", address)
```
sbinet: perhaps weave in the incoming `ctx`?
ie: use `net.Dialer.DialContext` ?

sbinet:
one thing that struck me: perhaps we could just merge the `protocol` and `encoder` packages into one package, `protocol`.
the encoding/decoding to/from raw bytes eminently pertains to the minute details of the XRootD protocol.
so it makes kind of sense to me.
it would also read:
```go
err := protocol.Unmarshal(data, &resp)
```
which looks slightly less non-sensical than:
```go
err := encoder.Unmarshal(...) // what? an encoder that decodes ? :)
```

what do you think?
(this obviously can be done in a separate PR)

xrootd/client/client.go
```diff
@@ -0,0 +1,141 @@
...
+			// TODO: should we just ignore responses to unclaimed stream IDs?
+		}
+
+		if header.Status != protocol.OkSoFar {
+			client.mux.Unclaim(header.StreamID)
+		}
+
+		select {
+		case <-ctx.Done():
+			// TODO: wait for active requests to be completed and handle errors from client.conn.Close().
```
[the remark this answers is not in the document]
EgorMatirov: First of all, thank you for the review!
It really helps to sort things out and I highly appreciate it.

I like the idea of having `Close` function, but I see a problem in proposed approach with cancellation logic inside `Cancel`:
Since `context` is passed to the client, it is expected that cancellation/deadline of that context will affect the client.
So, this workflow seems legit:
```go
ctx, cancel := context.WithCancel(ctx)
defer cancel()
client, err := client.NewClient(ctx,  "address")
// some logic and client usage
// ...
// cancel invokation
```

But thus the 
```go
    if err := cli.conn.Close(); err != nil { ... }
    if err := cli.mux.Close(); err != nil { ... }
```
lines inside `Cancel` will never be called.

I'm kinda new to Go and I'm not familiar with all of the conventions, and yes, it is highly likely that if there is `Cancel` user of the client will call it, especially if it's used in examples/documentation... But it still looks a bit confusing IMHO.

Maybe keep cancellation logic in `consume` and in `Cancel` simply call `context`'s `close`? But then it requires another goroutine or smth else to capture errors in `consume` and return them from `Cancel`.

EgorMatirov:
> one thing that struck me: perhaps we could just merge the protocol and encoder packages into one package, protocol.

Makes sense to me too. I'll do that in a separate PR. Thanks. :)

xrootd/client/client.go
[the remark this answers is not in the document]
sbinet: I am not claiming this is a completely 100% correct solution:
- https://play.golang.org/p/YTwxZRYYnzO

but it's the way I was imagining it :)
one can play with the condition at line 36 to enable an early-cancel by the user.

it seems to work (even under the `-race` detector :P)

you're right that, also in my example above, one needs some other medium (+possibly a goroutine) to communicate potential errors from the `consume` goroutine back to the `Close()` method.
perhaps a simple:
```go
type Client struct {
    // ... other fields elided ...
   mu sync.Mutex
   err error
}

func (cli *Client) consume() {
    for {
        select {
        default:
            if err := cli.conn.Close(); err != nil {
                cli.mu.Lock()
                if cli.err == nil { cli.err = err }
                cli.mu.Unlock()
            }
        }
    }
}
```

and return this `cli.err` error (guarded by the same mutex) in `Client.Close()`.

sbinet:
oops, sorry for close+reopen.

I didn't notice you have been sending PR against the `xrootd` branch instead of just `master`.
for the next PR, could you send it against `master` ? thanks.

xrootd/client/client.go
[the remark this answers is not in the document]
EgorMatirov: Line 144 (`log.Printf("existing loop...")`) is *almost* never reached in your suggestion as far as I understand. :-/
Loop blocks on `io.ReadFull` and then returns due to:
```go
if ctx.Err() != nil {
    // something happened to the context.
    // ignore this error.
    return
}
```

I have changed `return` to `continue`, so loop successfully continues to the `case <-ctx.Done()` part.
However, it's still impossible to have all of the connection closing logic here because `io.ReadFull` will not unblock until the connection is not closed. 
Probably, it's possible to extract reading to another goroutine and `select` with `case <-ctx.Done()` and `case <- serverResponse~, but it looks a bit overcomplicated.

I see another 2 possibilities: 
1) Use context just for connecting, like `DialContext` does:
```
// If the context expires before
// the connection is complete, an error is returned. Once successfully
// connected, any expiration of the context will not affect the
// connection.
```

2) Make another goroutine for calling `Cancel` when the context is cancelled:
```go
go func() {
    <-ctx.Done()
    client.Cancel()
}
```
and remove `select` from `consume`, leaving only `for`-loop which returns on context error:
```go
func (client *Client) consume(ctx context.Context) {
    for {
        // ...
       if ctx.Err() != nil {
	    // something happened to the context.
	    // ignore this error.
            return
	}
    }
}
```

Looks like awaiting active requests should be done in the same place where `client.conn.Close()` is called, because after that `any blocked Read or Write operations will be unblocked and return errors`.

EgorMatirov:
I have rebased PR against `master`. :)

xrootd/protocol/protocol.go
```diff
@@ -0,0 +1,49 @@
...
+
+// Package protocol contains the XRootD protocol specific types.
+package protocol // import "go-hep.org/x/hep/xrootd/protocol"
+
+import (
+	"encoding/binary"
+	"fmt"
+)
+
+const Ok = 0
```
sbinet: please document these exported constants.

xrootd/protocol/handshake/handshake.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package handshake contains the structures describing request and response
+// for handshake request (see XRootD specification).
+package handshake // import "go-hep.org/x/hep/xrootd/protocol/handshake"
+
+type ServerType int32
```
sbinet: please document this type.
it can be just some very short introductory text + a link to the XRootD specs.

xrootd/protocol/handshake/handshake.go
```diff
@@ -0,0 +1,31 @@
...
+
+const (
+	LoadBalancingServer ServerType = iota
```
sbinet: please document these consts.
it can be just some very short introductory text + a link to the XRootD specs.

xrootd/client/main_test.go
```diff
@@ -0,0 +1,11 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build xrootd_test_with_server
+
+package client // import "go-hep.org/x/hep/xrootd/client"
```
sbinet: is this file neeeded?
it doesn't seem to be needed at present.

xrootd/client/client.go
```diff
@@ -0,0 +1,179 @@
...
+// Close closes the connection. Any blocked operation will be unblocked and return error.
+func (client *Client) Close() error {
+	client.cancel()
+
+	err := client.conn.Close()
+	if err != nil {
+		return err
+	}
+
+	client.mux.Close()
```
sbinet: shouldn't we always close the multiplexer, even if `client.conn.Close` failed ?
ie:

```go
err := client.conn.Close()
client.mux.Close()
return err
```

or, even:
```go
client.mux.Close()
return client.conn.Close()
```

xrootd/client/client.go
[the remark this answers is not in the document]
sbinet: I don't think it's an issue if the `"exiting loop..."` case isn't selected, as long as the goroutine is exited.
your modification of issueing `continue` in case of the read-header request will indeed make sure do go through that `<-ctx.Done()` case (and then exit, thus releasing resources associated with the goroutine).

having all the "cleanup/closing" of resources in one place (_ie:_ in `func (*Client) Close() error`) (and not in the `<-ctx.Done()` case) is easier to reason about, I think.

xrootd/client/main_test.go
EgorMatirov: It defines an `Addr` flag, which is used by tests to connect to the remote XRootD server.  (locally I am running tests with `-tags xrootd_test_with_server --addr=ccxrootdgotest.in2p3.fr:9001`). That definition is placed in `main_test.go` because later there will be `protocol_test.go`, etc, and I'm not sure it will not be confusing to have that flag somewhere in one of these files. Anyway, if you think that it's ok to put it in `client_test.go`, I'll do that. :)

xrootd/client/client.go
EgorMatirov: My bad. Thank you!

xrootd/protocol/handshake/handshake.go
```diff
@@ -6,18 +6,25 @@
 // for handshake request (see XRootD specification).
 package handshake // import "go-hep.org/x/hep/xrootd/protocol/handshake"
 
+// ServerType is the general server type kept for compatibility
+// with 2.0 protocol version (see xrootd protocol specification v3.1.0, p. 5).
 type ServerType int32
 
 const (
+	// LoadBalancingServer indicates that server is a load-balancing server.
```
sbinet: `s/that server is a/whether this is a/`

xrootd/protocol/handshake/handshake.go
```diff
@@ -6,18 +6,25 @@
...
 	LoadBalancingServer ServerType = iota
-	DataServer          ServerType = iota
+	// DataServer indicates that server is a data server.
```
sbinet: `s/that server is a/whether this is a/`

xrootd/protocol/handshake/handshake.go
```diff
@@ -6,18 +6,25 @@
...
 )
 
+// Response is a response for the handshake request,
+// which contains protocol version and server type.
 type Response struct {
 	ProtocolVersion int32
 	ServerType      ServerType
 }
 
+// Request is the struct that holds handshake request parameters.
```
sbinet: `s/is the struct that holds/holds the/`

xrootd/protocol/handshake/handshake.go
```diff
@@ -26,6 +33,7 @@ type Request struct {
 	Reserved5 int32
 }
 
+// NewRequest forms a Request with predefined parameters.
```
sbinet: `with predefined parameters` is a bit cryptic and not very much helpful :)
instead, we could perhaps say that a `Request` is created to comply with the XRootD protocol version XXX.

xrootd/client/main_test.go
sbinet: ok. let's remove this file from this PR (and only add it as is in the PR that makes use of that `Addr` flag.)
I think I'll probably suggest another way to perform that test :)

EgorMatirov: > (and only add it as is in the PR that makes use of that Addr flag.)

It's used in `client_test.go` already (see line 19): `client, err := NewClient(ctx, *Addr)`

sbinet: Ah. Sorry. That passed me.

xrootd/client/client.go
```diff
@@ -0,0 +1,174 @@
...
+// NewClient creates a new xrootd client that connects to the given address.
+// When the context expires, a response handling is stopped, however, it is
+// necessary to call Cancel to correctly free resources.
+func NewClient(ctx context.Context, address string) (*Client, error) {
+	ctx, cancel := context.WithCancel(ctx)
+
+	var d net.Dialer
+	conn, err := d.DialContext(ctx, "tcp", address)
+	if err != nil {
+		return nil, err
```
sbinet: perhaps add a call to `cancel()` here?
I am not sure whether `d.DialContext` does it in case of an error...

xrootd/client/client.go
```diff
@@ -0,0 +1,174 @@
...
+				if ctx.Err() != nil {
+					// something happened to the context.
+					// ignore this error.
+					continue
+				}
+				panic(err)
+				// TODO: should redirect in case if is not possible to decode a header as well?
+			}
+
+			resp := mux.ServerResponse{make([]byte, header.DataLength), nil}
```
sbinet: please add struct field names.

xrootd/client/client_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+	"testing"
+)
+
+func TestNewClient(t *testing.T) {
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	client, err := NewClient(ctx, *Addr)
```
sbinet: ok.
instead of this `Addr` flag value I would rather prefer something like this:

```go
var testClientAddrs = []string{"0.0.0.0:9001"}

func TestNewClient(t *testing.T) {
    for _, addr := range testClientAddrs {
        t.Run(addr, func(t *testing.T) {
            testNewClient(t, addr)
        })
    }
}

func testNewClient(t *testing.T, addr string) { ... }
```

then, in some other `_test.go` file, perhaps protected with a specific build tag:
```go
// +build cxx-xrootd

package client

func init() {
    // add a C++ XRootD test server hosted in CC-Lyon.
    testClientAddrs = append(testClientAddrs, "ccxrootdgotest.in2p3.fr:9001")
}
```

what do you think ?

the advantage is that it's more explicit and the expected instructions on how to run the test(s) are kind of "documented in the code" (instead of some arcane way to run the test.)

xrootd/protocol/protocol.go
```diff
@@ -0,0 +1,59 @@
...
+type ResponseHeader struct {
+	StreamID   StreamID
+	Status     ResponseStatus
+	DataLength int32
+}
+
+// Error returns an error received from the server or nil if request hasn't failed.
+func (hdr ResponseHeader) Error(data []byte) error {
+	if hdr.Status == Error {
+		code := int32(binary.BigEndian.Uint32(data[0:4]))
```
sbinet: are we sure the input `data` slice will always be of size at least 4?
perhaps better to add a check for that?

what do you think?

xrootd/client/client.go
EgorMatirov: From documentation it looks like context is used to abort long-running connection primarily.

```
// The provided Context must be non-nil. If the context expires before
// the connection is complete, an error is returned. Once successfully
// connected, any expiration of the context will not affect the
// connection.
```
Anyway, I'll add a call just to be sure. :)

EgorMatirov:
Just to be sure, that you were notified... PTAL, @sbinet. :)

PR #195, opened 2018-05-21
xrootd/{client,protocol}: add protocol request

EgorMatirov:
Add Protocol request which queries for such info as:
- Protocol version
- Server type
- Security version
- Security options
- Security level
- Security overrides, that customize predefined security level

Information about security level and security overrides will further be used for signing requests when it's required by the server.

Client clean-up in the mock test is a bit awful due to not decided (yet) way of closing.
Response parsing is a bit awful too, since it copies data twice - from slice to the structs and then assembles these structs in more complicated response.

I want to make a generator of `Marshal`\\`Unmarshal` functions and handle bit-wise encoding and support of nested structs, slice length and so on.
Something like: 

```go
type ProtocolInfo struct {
	BinaryProtocolVersion int32 `order 0`
	ServerType            protocol.ServerType // Haven't come up with idea for enums yet.
	IsManager             bool `order 4 bit 0x00000002`
	IsServer              bool `order 4 bit 0x00000001`
	IsMeta                bool `order 4 bit 0x00000100`
	IsProxy               bool `order 4 bit 0x00000200`
	IsSupervisor          bool `order 4 bit 0x00000400`
	SecurityVersion       byte `order 10 optional` // Note that it is 10 due to 2 preceeding fields of int32 (4+4) and two reserved fields of byte (1+1)
	ForceSecurity         bool `order 11 bit 0x02 optional`
	SecurityLevel         byte `order 12 optional`
	SecurityOverridesSize byte `order 13 optional`
	SecurityOverrides     []protocolRequest.SecurityOverride `order 14 size SecurityOverridesSize`
}
```
Since I'm unsure that it is worth keeping a runtime-based Marshal\Unmarshal after generator will be written, I haven't added these features in it as part of that PR (or before that PR).

Regarding implementing the generator - I would like to deal with few complicated request before that to have better idea about possible features, if it is ok.

### DIFF
--- a/xrootd/client/client.go
+++ b/xrootd/client/client.go
@@ -35,10 +35,11 @@ import (
 // Concurrent requests are supported.
 // Zero value is invalid, Client should be instantiated using NewClient.
 type Client struct {
-	cancel          context.CancelFunc
-	conn            net.Conn
-	mux             *mux.Mux
-	protocolVersion int32
+	cancel           context.CancelFunc
+	conn             net.Conn
+	mux              *mux.Mux
+	protocolVersion  int32
+	signRequirements protocol.SignRequirements
 }
 
 // NewClient creates a new xrootd client that connects to the given address.
@@ -54,7 +55,7 @@ func NewClient(ctx context.Context, address string) (*Client, error) {
 		return nil, err
 	}
 
-	client := &Client{cancel, conn, mux.New(), 0}
+	client := &Client{cancel: cancel, conn: conn, mux: mux.New()}
 
 	go client.consume(ctx)
 
@@ -62,6 +63,14 @@ func NewClient(ctx context.Context, address string) (*Client, error) {
 		client.Close()
 		return nil, err
 	}
+
+	protocolInfo, err := client.Protocol(ctx)
+	if err != nil {
+		client.Close()
+		return nil, err
+	}
+
+	client.signRequirements = protocol.NewSignRequirements(protocolInfo.SecurityLevel, protocolInfo.SecurityOverrides)
 
 	return client, nil
 }
--- a/xrootd/client/main_mock_test.go
+++ b/xrootd/client/main_mock_test.go
@@ -1,0 +1,91 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package client // import "go-hep.org/x/hep/xrootd/client"
+
+import (
+	"context"
+	"encoding/binary"
+	"io"
+	"net"
+
+	"github.com/pkg/errors"
+	"go-hep.org/x/hep/xrootd/internal/mux"
+	"go-hep.org/x/hep/xrootd/protocol"
+)
+
+func testClientWithMockServer(serverFunc func(cancel func(), conn net.Conn), clientFunc func(cancel func(), client *Client)) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	server, conn := net.Pipe()
+	defer server.Close()
+	defer conn.Close()
+
+	client := &Client{cancel: cancel, conn: conn, mux: mux.New(), signRequirements: protocol.DefaultSignRequirements()}
+	defer client.Close()
+
+	go serverFunc(func() { client.Close() }, server)
+	go client.consume(ctx)
+
+	clientFunc(cancel, client)
+}
+
+func readRequest(conn net.Conn) ([]byte, error) {
+	// 16 is for the request options and 4 is for the data length
+	const requestSize = protocol.RequestHeaderLength + 16 + 4
+	var request = make([]byte, requestSize)
+	if _, err := io.ReadFull(conn, request); err != nil {
+		return nil, err
+	}
+
+	dataLength := binary.BigEndian.Uint32(request[protocol.RequestHeaderLength+16:])
+	if dataLength == 0 {
+		return request, nil
+	}
+
+	var data = make([]byte, dataLength)
+	if _, err := io.ReadFull(conn, data); err != nil {
+		return nil, err
+	}
+
+	return append(request, data...), nil
+}
+
+func writeResponse(conn net.Conn, data []byte) error {
+	n, err := conn.Write(data)
+	if err != nil {
+		return err
+	}
+	if n != len(data) {
+		return errors.Errorf("could not write all %d bytes: wrote %d", len(data), n)
+	}
+	return nil
+}
+
+// TODO: move marshalResponse outside of main_mock_test.go and use it for server implementation.
+func marshalResponse(responseParts ...interface{}) ([]byte, error) {
+	var data []byte
+	for _, p := range responseParts {
+		pData, err := protocol.Marshal(p)
+		if err != nil {
+			return nil, err
+		}
+		data = append(data, pData...)
+	}
+	return data, nil
+}
+
+// TODO: move unmarshalRequest outside of main_mock_test.go and use it for server implementation.
+func unmarshalRequest(data []byte, request interface{}) (protocol.RequestHeader, error) {
+	var header protocol.RequestHeader
+	if err := protocol.Unmarshal(data[:protocol.RequestHeaderLength], &header); err != nil {
+		return protocol.RequestHeader{}, err
+	}
+	if err := protocol.Unmarshal(data[protocol.RequestHeaderLength:], request); err != nil {
+		return protocol.RequestHeader{}, err
+	}
+
+	return header, nil
+}
--- a/xrootd/client/protocol.go
+++ b/xrootd/client/protocol.go
@@ -1,0 +1,93 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package client // import "go-hep.org/x/hep/xrootd/client"
+
+import (
+	"context"
+
+	xrdproto "go-hep.org/x/hep/xrootd/protocol"
+	"go-hep.org/x/hep/xrootd/protocol/protocol"
+)
+
+// ProtocolInfo is a response for the `Protocol` request. See details in the xrootd protocol specification.
+type ProtocolInfo struct {
+	BinaryProtocolVersion int32
+	ServerType            xrdproto.ServerType
+	IsManager             bool
+	IsServer              bool
+	IsMeta                bool
+	IsProxy               bool
+	IsSupervisor          bool
+	SecurityVersion       byte
+	ForceSecurity         bool
+	SecurityLevel         protocol.SecurityLevel
+	SecurityOverrides     []protocol.SecurityOverride
+}
+
+// Protocol obtains the protocol version number, type of the server and security information, such as:
+// the security version, the security options, the security level, and the list of alterations
+// needed to the specified predefined security level.
+func (client *Client) Protocol(ctx context.Context) (ProtocolInfo, error) {
+	resp, err := client.call(ctx, protocol.RequestID, protocol.NewRequest(client.protocolVersion, true))
+	if err != nil {
+		return ProtocolInfo{}, err
+	}
+
+	var generalResp protocol.GeneralResponse
+
+	if err = xrdproto.Unmarshal(resp, &generalResp); err != nil {
+		return ProtocolInfo{}, err
+	}
+
+	var securityInfo *protocol.SecurityInfo
+	if len(resp) > protocol.GeneralResponseLength {
+		securityInfo = &protocol.SecurityInfo{}
+		err = xrdproto.Unmarshal(resp[protocol.GeneralResponseLength:], securityInfo)
+		if err != nil {
+			return ProtocolInfo{}, err
+		}
+	}
+
+	var info = ProtocolInfo{
+		BinaryProtocolVersion: generalResp.BinaryProtocolVersion,
+		ServerType:            extractServerType(generalResp.Flags),
+
+		// TODO: implement bit-encoded fields support in Unmarshal.
+		IsManager:    generalResp.Flags&protocol.IsManager != 0,
+		IsServer:     generalResp.Flags&protocol.IsServer != 0,
+		IsMeta:       generalResp.Flags&protocol.IsMeta != 0,
+		IsProxy:      generalResp.Flags&protocol.IsProxy != 0,
+		IsSupervisor: generalResp.Flags&protocol.IsSupervisor != 0,
+	}
+
+	if securityInfo != nil {
+		info.SecurityVersion = securityInfo.SecurityVersion
+		info.ForceSecurity = securityInfo.SecurityOptions&protocol.ForceSecurity != 0
+		info.SecurityLevel = securityInfo.SecurityLevel
+
+		if securityInfo.SecurityOverridesSize > 0 {
+			info.SecurityOverrides = make([]protocol.SecurityOverride, securityInfo.SecurityOverridesSize)
+
+			const offset = protocol.GeneralResponseLength + protocol.SecurityInfoLength
+			const elementSize = protocol.SecurityOverrideLength
+
+			for i := byte(0); i < securityInfo.SecurityOverridesSize; i++ {
+				err = xrdproto.Unmarshal(resp[offset+elementSize*int(i):], &info.SecurityOverrides[i])
+				if err != nil {
+					return ProtocolInfo{}, err
+				}
+			}
+		}
+	}
+
+	return info, nil
+}
+
+func extractServerType(flags protocol.Flags) xrdproto.ServerType {
+	if int32(flags)&int32(xrdproto.DataServer) != 0 {
+		return xrdproto.DataServer
+	}
+	return xrdproto.LoadBalancingServer
+}
--- a/xrootd/client/protocol_mock_test.go
+++ b/xrootd/client/protocol_mock_test.go
@@ -1,0 +1,93 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package client
+
+import (
+	"context"
+	"net"
+	"reflect"
+	"testing"
+
+	xrdproto "go-hep.org/x/hep/xrootd/protocol"
+	"go-hep.org/x/hep/xrootd/protocol/protocol"
+)
+
+func TestClient_Protocol_WithSecurityInfo(t *testing.T) {
+	var protocolVersion int32 = 0x310
+
+	serverFunc := func(cancel func(), conn net.Conn) {
+		defer cancel()
+
+		data, err := readRequest(conn)
+		if err != nil {
+			t.Fatalf("could not read request: %v", err)
+		}
+
+		var gotRequest protocol.Request
+		gotHeader, err := unmarshalRequest(data, &gotRequest)
+		if err != nil {
+			t.Fatalf("could not unmarshal request: %v", err)
+		}
+
+		if gotHeader.RequestID != protocol.RequestID {
+			t.Fatalf("invalid request id was specified:\nwant = %d\ngot = %d\n", protocol.RequestID, gotHeader.RequestID)
+		}
+
+		if gotRequest.ClientProtocolVersion != protocolVersion {
+			t.Fatalf("invalid client protocol version was specified:\nwant = %d\ngot = %d\n", protocolVersion, gotRequest.ClientProtocolVersion)
+		}
+
+		flags := protocol.IsManager | protocol.IsServer | protocol.IsMeta | protocol.IsProxy | protocol.IsSupervisor
+
+		responseHeader := xrdproto.ResponseHeader{
+			StreamID:   gotHeader.StreamID,
+			DataLength: protocol.GeneralResponseLength + protocol.SecurityInfoLength + protocol.SecurityOverrideLength,
+		}
+
+		protocolResponse := protocol.GeneralResponse{protocolVersion, flags}
+
+		protocolSecurityInfo := protocol.SecurityInfo{
+			SecurityOptions:       protocol.None,
+			SecurityLevel:         protocol.Pedantic,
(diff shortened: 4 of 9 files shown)


xrootd/client/protocol_mock_test.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		if err := writeResponse(conn, response); err != nil {
+			cancel()
+			t.Errorf("invalid write: %s", err)
+		}
+	}
+
+	var want = ProtocolInfo{
+		protocolVersion,
```
sbinet: please add struct fields names here:
```go
want := ProtocolInfo{
    BinaryProtocolVersion: protocolVersion,
    ...
}
```

xrootd/client/protocol.go
```diff
@@ -0,0 +1,92 @@
...
+// license that can be found in the LICENSE file.
+
+package client // import "go-hep.org/x/hep/xrootd/client"
+
+import (
+	"context"
+
+	"go-hep.org/x/hep/xrootd/encoder"
+	"go-hep.org/x/hep/xrootd/protocol"
+	protocolRequest "go-hep.org/x/hep/xrootd/protocol/protocol"
```
sbinet: package names are usually not "camel-cased".
could we think about a better name in this unfortunate name collision?

`xproto` ? `protoreq` ?

xrootd/client/protocol.go
```diff
@@ -0,0 +1,92 @@
...
+	var generalResp protocolRequest.GeneralResponse
+
+	if err = encoder.Unmarshal(resp, &generalResp); err != nil {
+		return ProtocolInfo{}, err
+	}
+
+	var securityInfo *protocolRequest.SecurityInfo
+	if len(resp) > protocolRequest.GeneralResponseLength {
+		securityInfo = &protocolRequest.SecurityInfo{}
+		err = encoder.Unmarshal(resp[protocolRequest.GeneralResponseLength:], securityInfo)
```
sbinet: it seems we are dropping this error on the floor.
shouldn't we return it here ?

xrootd/client/protocol.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	var info = ProtocolInfo{}
```
sbinet: it seems we could create a `ProtocolInfo` value with much of its fields already initialized to the correct values, like so:

```go
info := ProtocolInfo{
    BinaryProtocolVersion: generalResp.BinaryProtocolVersion,
    ServerType: extractServerType(generalResp.Flags),
    ...
}
```

xrootd/client/protocol_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"context"
+	"reflect"
+	"testing"
+
+	"go-hep.org/x/hep/xrootd/protocol"
+)
+
+func TestClient_Protocol(t *testing.T) {
+	var want = ProtocolInfo{
+		784,
```
sbinet: please use struct field names here as well.

xrootd/protocol/protocol/protocol.go
```diff
@@ -0,0 +1,121 @@
...
+2) SecurityInfo - a response part that is added to general response
+if `ReturnSecurityRequirements` is provided and server supports it.
+It contains security version, security options, security level,
+and number of following security overrides, if any.
+
+3) A list of SecurityOverride - alterations needed to the specified predefined security level.
+*/
+package protocol // import "go-hep.org/x/hep/xrootd/protocol/protocol"
+
+// RequestID is the id of the request, it is sent as part of message. See xrootd protocol specification for details.
```
sbinet: > See xrootd protocol specification for details.

please add either a link and/or the name of the section that deals with these details.

xrootd/protocol/protocol/protocol.go
```diff
@@ -0,0 +1,121 @@
...
+	RequestOptionsNone RequestOptions = 0
+	// ReturnSecurityRequirements specifies that security requirements should be returned
+	// if that's supported by the server.
+	ReturnSecurityRequirements RequestOptions = 1
+)
+
+// Request is the struct that holds protocol request parameters.
+type Request struct {
+	ClientProtocolVersion int32
+	Reserved1             [11]byte
```
sbinet: if it works with the current `reflect`-based code of the marshaling/unmarshaling protocol, we should rename these `ReservedX` fields into just `_`.

if this doesn't work with the `reflect`-based code, then leave it like that but add a FIXME stating we should rename these into `_` when the automatically generated (un)marshaling code is in place.

xrootd/protocol/protocol/protocol.go
```diff
@@ -0,0 +1,121 @@
...
+	Reserved2             int32
+}
+
+// NewRequest forms a Request according to provided parameters.
+func NewRequest(protocolVersion int32, withSecurityRequirements bool) Request {
+	var options = RequestOptionsNone
+	if withSecurityRequirements {
+		options |= ReturnSecurityRequirements
+	}
+	return Request{protocolVersion, [11]byte{}, options, 0}
```
sbinet: please use struct field names here.
you can drop the `ReservedX` fields:

```go
return Request{ClientProtocolVersion: protocolVersion, Options:options}
```

xrootd/client/protocol.go
sbinet: or name the other import like:
```go
import (
    xrdproto "go-hep.org/x/hep/xrootd/protocol"
    "go-hep.org/x/hep/xrootd/protocol/protocol"
)
```
?

EgorMatirov:
@sbinet, PTAL.

The problem with that request was in the difference between specification and actual implementation:
```go
// NOTE: That request differs from one specified in http://xrootd.org/doc/dev45/XRdv310.pdf, p. 71:
// order of Options and Reserved fields differs in the protocol specification and
// actual implementation which can be found at https://github.com/xrootd/xrootd/blob/master/src/XProtocol/XProtocol.hh
// (see ClientProtocolRequest definition).
type Request struct {
	ClientProtocolVersion int32
	Options               RequestOptions
	// FIXME: Rename Reserved* fields to _ when automatically generated (un)marshalling will be available.
	Reserved1 [11]byte
	Reserved2 int32
}
```

Right now, I used the order of implementation and added a `NOTE:`.
Should I do something else? (Probably create an issue for xrootd?)

sbinet:
Yes I guess we should file an issue against xrootd. (Probably to change the specs to match the implementation :P)

xrootd/protocol/signRequirements.go
```diff
@@ -0,0 +1,59 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
```
sbinet: by convention, we don't use very much camelCase names for files.
I would either rename `signRequirements.go` into `signing.go` or `signed_requirements.go`.

or, perhaps, just merge the content of that file at the bottom of `protocol.go` ?

xrootd/protocol/signRequirements.go
```diff
@@ -0,0 +1,59 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package protocol contains the XRootD protocol specific types.
```
sbinet: please remove this package-level comment as it may override the much nicer one that already exists in `protocol.go`.

xrootd/protocol/signRequirements.go
```diff
@@ -0,0 +1,59 @@
...
+		return false
+	}
+	if v == protocol.SignLikely && !modifies {
+		return false
+	}
+	return true
+}
+
+// NewSignRequirementsDefault creates a default SignRequirements with "None" security level.
+func NewSignRequirementsDefault() SignRequirements {
```
sbinet: to follow suit of `http.DefaultClient` or `gonum.org/v1/gonum/optimize.DefaultSettings`, I would:
`s/NewSignRequirementsDefault/DefaultSignRequirements/`

xrootd/protocol/signRequirements.go
```diff
@@ -0,0 +1,59 @@
...
+	"go-hep.org/x/hep/xrootd/protocol/protocol"
+)
+
+// SignRequirements implements a way to check if request should be signed
+// according to XRootD protocol specification v. 3.1.0, p.75-76.
+type SignRequirements struct {
+	requirements map[uint16]protocol.RequestLevel
+}
+
+// Needed checks if the request should be signed.
```
sbinet: `s/Needed checks if the request should be signed./Needed returns whether the request should be signed./`

`modifies` should also probably be documented.

xrootd/client/main_mock_test.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+		data = append(data, pData...)
+	}
+	return data, nil
+}
+
+// TODO: move unmarshalRequest outside of main_mock_test.go and use it for server implementation.
+func unmarshalRequest(data []byte, request interface{}) (protocol.RequestHeader, error) {
+	var header protocol.RequestHeader
+	if err := encoder.Unmarshal(data[0:protocol.RequestHeaderLength], &header); err != nil {
```
sbinet: `s/[0:protocol.RequestHeaderLength]/[:protocol.RequestHeaderLength]/`

xrootd/client/protocol.go
```diff
@@ -0,0 +1,96 @@
...
+	IsProxy               bool
+	IsSupervisor          bool
+	SecurityVersion       byte
+	ForceSecurity         bool
+	SecurityLevel         protocol.SecurityLevel
+	SecurityOverrides     []protocol.SecurityOverride
+}
+
+/*
+Protocol obtains the protocol version number, type of the server and security information, such as:
```
sbinet: please use `//` comments instead of `/* .. */`. (just for consistency's sake)

xrootd/protocol/protocol.go
```diff
@@ -63,3 +73,14 @@ func (hdr ResponseHeader) Error(data []byte) error {
...
 	return nil
 }
+
+// ServerType is the general server type kept for compatibility
+// with 2.0 protocol version (see xrootd protocol specification v3.1.0, p. 5).
+type ServerType int32
+
+const (
+	// LoadBalancingServer indicates whether this is a load-balancing server.
+	LoadBalancingServer ServerType = iota
```
sbinet: I would write it like so:
```go
const (
	// LoadBalancingServer indicates whether this is a load-balancing server.
	LoadBalancingServer ServerType = iota

	DataServer
)
```

xrootd/protocol/protocol/protocol.go
```diff
@@ -0,0 +1,153 @@
...
+
+/*
+Package protocol contains the structures describing request and response
+for protocol request (see XRootD specification).
+
+Response consist of 3 parts:
+
+1) GeneralResponse - general response that is always returned and specifies protocol version and flags describing server type.
+
+2) SecurityInfo - a response part that is added to general response
```
sbinet: `s/added to general/added to the general/`

xrootd/protocol/protocol/protocol.go
```diff
@@ -0,0 +1,153 @@
...
+if `ReturnSecurityRequirements` is provided and server supports it.
+It contains security version, security options, security level,
```
sbinet: `s/security version/the security version/`
`s/security options/the security options/`
`s/security level/the security level/`

xrootd/protocol/protocol/protocol.go
```diff
@@ -0,0 +1,153 @@
...
+and number of following security overrides, if any.
```
sbinet: `s/and number/and the number/`

xrootd/protocol/protocol/protocol.go
```diff
@@ -0,0 +1,153 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+/*
```
sbinet: please use `// ` instead of `/* ... */` (for consistency's sake.)

xrootd/protocol/protocol/protocol.go
```diff
@@ -0,0 +1,153 @@
...
+type GeneralResponse struct {
+	BinaryProtocolVersion int32
+	Flags                 Flags
+}
+
+// Flags are the flags that define xrootd server type. See xrootd protocol specification for further info.
+type Flags int32
+
+const (
+	// IsServer indicates whether this server has server role.
```
sbinet: this whole block could be rewritten to look like:
```go
const (
    IsServer  Flags = 0x00000001 // IsServer indicates whether this server has server role.
    IsManager Flags = 0x00000002 // ...
    ...
)
```

xrootd/protocol/protocol/protocol.go
```diff
@@ -0,0 +1,153 @@
...
+	// SignLikely indicates that the request must be signed if it modifies data.
+	SignLikely RequestLevel = 1
+	// SignNeeded indicates that the request mush be signed.
+	SignNeeded RequestLevel = 2
+)
+
+// SecurityOverrideLength is the length of SecurityOverride in bytes.
+const SecurityOverrideLength = 2
+
+// SecurityOverride is a alteration needed to the specified predefined security level.
```
sbinet: `s/is a/is an/`

xrootd/protocol/protocol/protocol.go
```diff
@@ -0,0 +1,153 @@
...
+// It consist of the request index and the security requirement that associated request is to have.
```
sbinet: `s/It consist/It consists/`
`s/that associated/the associated/`
`s/is to have/should have/`

sbinet:
I've filed https://github.com/xrootd/xrootd/issues/717

EgorMatirov:
Thank you very much for the review!

I have rebased on top of the `master` and added suggested changes.

xrootd/protocol/signRequirements.go
EgorMatirov: I would like not to put it into `protocol.go` because I expect a lot of "configuration" code to be placed here to initialize the predefined levels. I'm fine with `signing.go`. :)

xrootd/client/protocol.go
```diff
@@ -0,0 +1,93 @@
...
+	SecurityVersion       byte
+	ForceSecurity         bool
+	SecurityLevel         protocol.SecurityLevel
+	SecurityOverrides     []protocol.SecurityOverride
+}
+
+// Protocol obtains the protocol version number, type of the server and security information, such as:
+// the security version, the security options, the security level, and the list of alterations
+// needed to the specified predefined security level.
+func (client *Client) Protocol(ctx context.Context) (ProtocolInfo, error) {
```
sbinet: I am wondering whether `Protocol` should be a method on `xrootd/client.Client` or a function of `xrootd/protocol/protocol`...

if we keep `Protocol` a method of `client.Client`, then I suppose we'll have to have every single request/response as a method of `client.Client` as well (_ie_ 32 methods)

I am thinking out loud here (so it may just be a completely bunker idea), but it seems we could just get away with 1 or 2 methods in `client.Client` and keep the rest in each of the `xyz` protocol packages.
it *seems* to me this would introduce less coupling between `client.Client` and each request/response.

for example:

- `xrootd/protocol/protocol`:
```go
type Request { ... }
func NewRequest(...) Request { ... }

type Response { ... } // essentially client.ProtocolInfo

func Send(ctx context.Context, cli *xrdclient.Client, req Request) (Response, error) {
    raw, err := cli.Call(ctx, req.ID(), req.Marshal(cli.Version()))
    // ...
    var resp Response
    err = xrdproto.Unmarshal(&resp)
    return resp, err
}
```

I thought that was what you were aiming at when you put each request-type into its own package.

that said, I am not completely happy with the `Send` function as it is.
Perhaps, `Send` should actually be a method of `xrdclient.Client`, but it would take a (yet to be defined) `xrdproto.Request` interface as argument:

```go
func (cli *Client) Send(ctx context.Context, resp Response, req Request) error {
    raw, err := cli.marshal(req)
    recv, err := cli.call(ctx, req.ID(), raw)
    err = cli.unmarshal(raw, resp)
    return err
}
```
and used like so:
```go
var resp protocol.Response
err := cli.Send(ctx, &resp, protocol.NewRequest(...))
```

from a user perspective, it would look better like so:
```go
resp, err := cli.Send(ctx, protocol.NewRequest(...))
```
but one would have to type-assert to get the `protocol.Response` type...

I am still on the fence.
happy to get your input :)

EgorMatirov: It sounds reasonable to me. 
In fact, most of these `Client.*` request methods will be pretty straightforward - create request, send it, parse and return response.

However, from the user perspective it is:
```go
dirs, err := client.Dirlist(ctx, "\tmp")
```
vs
```go
var resp dirlist.Response
err := client.Send(ctx, dirlist.NewRequest("\tmp"))
dirs := resp.Dirs
// (or without predeclaration but with type-assertion)
```
So we have a lot more boilerplate code from user perspective IMHO.
Also, it's much easier to type "client." and get a list of code completions than be like "What was the name of that request again?".

Actually, I think, we can implement both ways:
- After implementing a few more requests and the marshalling code generator, implement routine like this:
```go
 raw, err := client.marshal(req)
 recv, err := client.call(...)
 err = client.unmarshal(...)
 return err
```
- Make small helper methods of `Client`:
```go
 func (client *Client) Dirlist(ctx context.Context, path string) ([]string, err) {
     var resp dirlist.Response
     err := client.Send(ctx, &resp, dirlist.NewRequest(path))
     if err != nil {
         return nil, err
     }
     return resp.Dirs, nil
 }
```

What do you think about it?

sbinet:
(moving the discussion to the main thread)

I agree it's more verbose from the `Client` user.
and I agree we should devise something a bit more easy on the keyboard as a user-facing API.
I also agree having everything exploded into many small-ish packages (for the req/resp) doesn't help discoverability (although, that can be somewhat mitigated with examples).

but, perhaps `Client` isn't the best place to hang `Dirlist`?
what about a few helper types that regroup functionalities by "themes"?

`Dirlist` is clearly something related to a filesystem.
`kXR_rm`, `kXR_rmdir`, `kXR_chmod`, (to name a few) as well.
while `kXR_read`, `kXR_readv` and `kXR_write` are more related to a (remote) file.

what about, then:
```go
type File struct {
    c *xrdclient.Client
    name string
}

func (f *File) Readdirnames(n int) ([]string, error) {
    var resp dirlist.Response
    err := f.c.Send(ctx, &resp, dirlist.NewRequest(f.name))
    return resp.Dirs, err
}
```

EgorMatirov:
I like that idea.

There is a number of specific requests like:
- kXR_admin
- kXR_auth
- kXR_bind 
- kXR_login
- kXR_ping
- kXR_protocol
- kXR_query
- kXR_set

that are more related to a "session" and they probably can reside inside `Client`.
I'm not sure, but probably most of them are so specific that they will be called only during `Client` connection routine (protocol, login, auth, bind) and at some point we can decide even unexport them.

Something like that for example:
```go

type Client struct {
    // connection-wise info
}

func (c *Client) Ping(ctx context.Context) error {
	// ...
}

type Filesystem struct {
    c *Client
}

func (f *Filesystem) Dirlist(ctx context.Context) ([]string, error) {
    // ...
}

type File struct {
    c *Client
    path string
}

func NewFile(client *Client, path string) File {
    return File{client, path}
}

func (f *File) Read(ctx context.Context, offset int64, length int32) ([]byte, error) {
    // ...
}
```

However, `Read` bothers me because of "readahead list" (see [protocol specification, p. 99-100](http://xrootd.org/doc/dev45/XRdv310.htm#_Toc464248841)). It looks a bit awful to call `Read` on the specific file and pass a list of readaheads to be done for other files.
Do you know how wide this feature is used (just in case)? I mean, if it is more "specific" scenario, it is ok to leave it to more "low-level" routine:
```go
var resp read.Response
err := client.Send(ctx, &resp, read.NewRequest(...))
```

Anyway, even with that additional argument in `Read` call I really like that approach.

sbinet:
_wrt_ `kXR_admin`, ... `kXR_set`: agreed. they probably are bound to either `Client` or `Session`.

(just nitpicking but:
```go
func (fs *Filesystem) Open(path string) (*File, error) { ... }
func (fs *Filesystem) Create(path string) (*File, error) { ... }
func (fs *Filesystem) OpenFile(path string, flag int, perm os.FileMode) (*File, error) { ... }
```
(ie: I wouldn't have `NewFile` as a function)

I think for the readahead list, we could as a first pass, just discard that feature.
eventually, we would send the read request to the client for that file, and have the client gather all these requests and merge into a single one with all the readaheads for all these files.

but we can table this for now :)

xrootd/client/client.go
```diff
@@ -54,7 +55,7 @@ func NewClient(ctx context.Context, address string) (*Client, error) {
 		return nil, err
 	}
 
-	client := &Client{cancel, conn, mux.New(), 0}
+	client := &Client{cancel: cancel, conn: conn, mux: mux.New(), signRequirements: protocol.DefaultSignRequirements()}
```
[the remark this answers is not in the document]
EgorMatirov: Yep, you are right.

EgorMatirov:
@sbinet, sorry, what about merging this one? It looks like that we agree that it can be placed in `Client` for now.

I have updated PR with the fixes of other issues (somehow github managed to put that commit *above* our conversation, so I'm unsure if you noticed it). Commit messsage now includes "Updates go-hep/hep#170" too.

xrootd/protocol/protocol/protocol.go
```diff
@@ -0,0 +1,144 @@
...
+const (
+	// RequestOptionsNone specifies that only general response should be returned.
+	RequestOptionsNone RequestOptions = 0
+	// ReturnSecurityRequirements specifies that security requirements should be returned
+	// if that's supported by the server.
+	ReturnSecurityRequirements RequestOptions = 1
+)
+
+// Request holds protocol request parameters.
+// NOTE: That request differs from one specified in http://xrootd.org/doc/dev45/XRdv310.pdf, p. 71:
```
sbinet: the specs have been "fixed" to follow what's in the code :)
please remove.

sbinet:
FYI, the test failure in `master` is coming from:
- https://github.com/golang/go/issues/25596
- https://github.com/golang/go/issues/25615

PR #201, opened 2018-05-25
xrootd/client: fix 'more' usage in channel reading

EgorMatirov:
It is set to false when the channel is already closed.

### DIFF
--- a/xrootd/client/client.go
+++ b/xrootd/client/client.go
@@ -139,25 +139,27 @@ func (client *Client) send(ctx context.Context, responseChannel mux.DataRecvChan
 		return nil, err
 	}
 
-	more := true
-	var resp mux.ServerResponse
 	var data []byte
 
-	for more {
+	for {
 		select {
-		case resp, more = <-responseChannel:
-			data = append(data, resp.Data...)
+		case resp, more := <-responseChannel:
+			if !more {
+				return data, nil
+			}
+
 			if resp.Err != nil {
 				return nil, resp.Err
 			}
+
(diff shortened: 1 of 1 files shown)


xrootd/client/client.go
```diff
@@ -146,9 +146,11 @@ func (client *Client) send(ctx context.Context, responseChannel mux.DataRecvChan
 	for more {
 		select {
 		case resp, more = <-responseChannel:
-			data = append(data, resp.Data...)
-			if resp.Err != nil {
-				return nil, resp.Err
+			if more {
```
sbinet: it would read better like so:
```go
if !more {
    return data, nil
}
data = append(data, resp.Data...)
if resp.Err != nil {
    return nil, resp.Err
}
```

(always try to minimize indentation)

this actually begs the question of the real need for the `for more { ... }` loop.
wouldn't it work as well like so:
```go
for {
    select {
    case resp, more := <-responseChannel:
        if !more { return ... }
        // ...
    case <-ctx.Done():
        // ...
    }
}
panic("unreachable")
```

EgorMatirov: Thanks!
Can you please explain what that `panic("unreachable")` is for? (I mean, why put it there).
I'm just curious. :)

sbinet: just an old habit inherited from Go <= 1.6 (I think)
when the compiler wasn't clever enough to deduce there were no code-path leading to this position.
also, a safeguard for when I haven't drunk enough coffee and there actually *is* a valid path to this position :)

PR #202, opened 2018-05-25
xrootd/internal/mux: close all channels with error message after Close

EgorMatirov:
So all active request calls will unblock and receive an error message stating that Close was called.

### DIFF
--- a/xrootd/internal/mux/mux.go
+++ b/xrootd/internal/mux/mux.go
@@ -87,6 +87,7 @@ func New() *Mux {
 			case m.freeIDs <- i:
 				i = (i + 1) % streamIDPoolSize
 			case <-m.quit:
+				close(m.freeIDs)
 				return
 			}
 		}
@@ -105,6 +106,12 @@ func (m *Mux) Close() {
 	m.closed = true
 	m.mu.Unlock()
 	close(m.quit)
+
+	response := ServerResponse{Err: errors.New("xrootd: close was called before response was fully received")}
+	for streamID := range m.dataWaiters {
+		m.SendData(streamID, response)
+		m.Unclaim(streamID)
+	}
 }
 
 // Claim searches for unclaimed id and returns corresponding channel.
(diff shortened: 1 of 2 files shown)


EgorMatirov:
Already noticed, that CI failed, my bad. :(

sbinet:
I'll wait for go-hep/hep#203 to go in before going back to this one :)

sbinet:
this needs to be rebased (to capture changes of go-hep/hep#203).

xrootd/internal/mux/mux.go
```diff
@@ -105,6 +105,12 @@ func (m *Mux) Close() {
 	m.closed = true
 	m.mu.Unlock()
 	close(m.quit)
+
+	response := ServerResponse{Err: errors.New("xrootd: close was called before response was fully received")}
+	for streamID := range m.dataWaiters {
```
sbinet: there's a possible race here.
I think we should create a slice of ids, lock, loop over the dataWaiters keys, populating the slice with those, unlock, and then iterate over the slice.

```go
m.closed = true
close(m.quit)

ids := make([]protocol.StreamID, 0, len(m.dataWaiters))
for id := range m.dataWaiters {
    ids = append(ids, id)
}
m.mu.Unlock()

for _, id := range ids {
    m.SendData(id, response)
    m.Unclaim(id)
}
```

EgorMatirov: You are right.

If Claim is called during Close, then race occurs. But isn't it better to check if `mux` is in closed state inside `Claim` and `ClaimWithID`?
In your snippet, if `Claim` is called, then after unlocking mutex it is possible that another id will be claimed and it will not be closed by the next loop.

What do you think about something like that in `Claim` and `ClaimWithID` instead?
```go
m.mu.Lock()
if m.closed {
	m.mu.Unlock()
	return protocol.StreamID{}, nil, errors.New("mux: Claim was called on closed Mux")
}
if _, claimed := m.dataWaiters[streamId]; claimed { // Skip id if it was already claimed manually via ClaimWithID
	m.mu.Unlock()
	continue
}

m.dataWaiters[streamId] = ch
m.mu.Unlock()
```

sbinet: sounds good.

PR #204, opened 2018-05-25
all: first stab at vgo support

sbinet:


sbinet:
Abandonned in favor of go-hep/hep#209.

PR #208, opened 2018-05-30
xrootd: add login request

EgorMatirov:
Login request should be issued prior to the most of other requests
(see http://xrootd.org/doc/dev45/XRdv310.pdf, p.10).
As part of the response, SecurityInformation may be provided,
indicating that auth request is required. SecurityInformation
defines available authentication protocols and additional parameters.

Updates go-hep/hep#170.


xrootd/protocol/login/login.go
```diff
@@ -0,0 +1,59 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package login contains the structures describing request and response for login request.
+// Login request should be issued prior to the most of other
```
sbinet: `s/to the most of other/to most of the other/`

xrootd/protocol/login/login.go
```diff
@@ -0,0 +1,59 @@
...
+// requests (see http://xrootd.org/doc/dev45/XRdv310.pdf, p.10).
+// As part of the response, SecurityInformation may be provided,
+// indicating that auth request is required. SecurityInformation
```
sbinet: `s/that auth/that an auth/`

xrootd/protocol/login/login.go
```diff
@@ -0,0 +1,59 @@
...
+// defines available authentication protocols and additional parameters.
```
sbinet: `s/defines/defines the/`
`s/and additional/together with some additional/`

xrootd/protocol/login/login.go
```diff
@@ -0,0 +1,59 @@
...
+// See XRootD protocol specification, page 127 for the further info
```
sbinet: `s/for the further info/for further informations/`

xrootd/protocol/login/login.go
```diff
@@ -0,0 +1,59 @@
...
+	Role         byte    // Role is the role being assumed for this login: administrator or regular user.
+	TokenLength  int32   // TokenLength is the length of the following token.
+	Token        []byte  // Token is the token supplied by the previous redirection response, plus optional elements.
+}
+
+// Capabilities for v3.1.0 client without async support.
+const clientCapabilities byte = 4
+
+// NewRequest forms a Request according to provided parameters.
+func NewRequest(username string, token string) Request {
```
sbinet: `s/username string, token string/username, token string/`

xrootd/client/client.go
```diff
@@ -9,7 +9,7 @@ The NewClient function connects to a server:
 
 	ctx := context.Background()
 
-	client, err := NewClient(ctx, *Addr)
+	client, err := NewClient(ctx, *Addr, username)
```
sbinet: `s/*Addr/addr/`

xrootd/client/client.go
```diff
@@ -42,10 +42,10 @@ type Client struct {
 	signRequirements protocol.SignRequirements
 }
 
-// NewClient creates a new xrootd client that connects to the given address.
+// NewClient creates a new xrootd client that connects to the given address using username.
 // When the context expires, a response handling is stopped, however, it is
 // necessary to call Cancel to correctly free resources.
-func NewClient(ctx context.Context, address string) (*Client, error) {
+func NewClient(ctx context.Context, address string, username string) (*Client, error) {
```
sbinet: I am wondering whether we shouldn't/couldn't use "functional options" here (so the number of parameters can be limited in simple cases, instead of passing `NewClient(ctx, "example.com:9999", "")` when identification, authentication and authorization aren't required)
although, in the real world, identification+authentication *is* required...

let's leave it like that for the moment.

PR #211, opened 2018-05-30
xrootd: add dirlist request

EgorMatirov:
Updates go-hep/hep#170.

I'm not sure about testing via `fs.Dirlist(context.Background(), "/tmp/.xrootd")`, since response may change. Probably we need to create some permanent directory on the remote server for test purposes?

### DIFF
--- a/xrootd/client/filesystem.go
+++ b/xrootd/client/filesystem.go
@@ -1,0 +1,36 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package client // import "go-hep.org/x/hep/xrootd/client"
+
+import (
+	"context"
+
+	"go-hep.org/x/hep/xrootd/xrdfs"
+	"go-hep.org/x/hep/xrootd/xrdproto/dirlist"
+)
+
+// FS returns a xrdfs.FileSystem which uses this client to make requests.
+func (cli *Client) FS() xrdfs.FileSystem {
+	return &fileSystem{cli}
+}
+
+// fileSystem contains filesystem-related methods of the XRootD protocol.
+type fileSystem struct {
+	c *Client
+}
+
+// Dirlist returns the contents of a directory together with the stat information.
+func (fs *fileSystem) Dirlist(ctx context.Context, path string) ([]xrdfs.EntryStat, error) {
+	var resp dirlist.Response
+	err := fs.c.Send(ctx, &resp, dirlist.NewRequest(path))
+	if err != nil {
+		return nil, err
+	}
+	return resp.Entries, err
+}
+
+var (
+	_ xrdfs.FileSystem = (*fileSystem)(nil)
+)
--- a/xrootd/client/filesystem_test.go
+++ b/xrootd/client/filesystem_test.go
@@ -1,0 +1,50 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package client
+
+import (
+	"context"
+	"reflect"
+	"testing"
+
+	"go-hep.org/x/hep/xrootd/xrdfs"
+)
+
+func testFileSystem_Dirlist(t *testing.T, addr string) {
+	var want = []xrdfs.EntryStat{
+		{
+			EntryName:   "file1.txt",
+			HasStatInfo: true,
+			ID:          60129606914,
+			EntrySize:   17,
+			Mtime:       1527753618,
+			Flags:       xrdfs.StatIsReadable | xrdfs.StatIsWritable,
+		},
+	}
+
+	client, err := NewClient(context.Background(), addr, "gopher")
+	if err != nil {
+		t.Fatalf("could not create client: %v", err)
+	}
+	defer client.Close()
+
+	fs := client.FS()
+
+	got, err := fs.Dirlist(context.Background(), "/tmp/dir1")
+	if err != nil {
+		t.Fatalf("invalid protocol call: %v", err)
+	}
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("FileSystem.Dirlist()\ngot = %v\nwant = %v", got, want)
+	}
+}
+
+func TestFileSystem_Dirlist(t *testing.T) {
+	for _, addr := range testClientAddrs {
+		t.Run(addr, func(t *testing.T) {
+			testFileSystem_Dirlist(t, addr)
+		})
+	}
+}
--- a/xrootd/xrdfs/stat.go
+++ b/xrootd/xrdfs/stat.go
@@ -1,0 +1,180 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package xrdfs // import "go-hep.org/x/hep/xrootd/xrdfs"
+
+import (
+	"bytes"
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/pkg/errors"
+	"go-hep.org/x/hep/xrootd/internal/xrdenc"
+)
+
+// StatFlags identifies the entry's attributes.
+type StatFlags int32
+
+const (
+	// StatIsExecutable indicates that entry is either an executable file or a searchable directory.
+	StatIsExecutable StatFlags = 1
+	// StatIsDir indicates that entry is a directory.
+	StatIsDir StatFlags = 2
+	// StatIsOther indicates that entry is neither a file nor a directory.
+	StatIsOther StatFlags = 4
+	// StatIsOffline indicates that the file is not online (i. e., on disk).
+	StatIsOffline StatFlags = 8
+	// StatIsReadable indicates that read access to that entry is allowed.
+	StatIsReadable StatFlags = 16
+	// StatIsWritable indicates that write access to that entry is allowed.
+	StatIsWritable StatFlags = 32
+	// StatIsPOSCPending indicates that the file was created with kXR_posc and has not yet been successfully closed.
+	// kXR_posc is an option of open request indicating that the "Persist On Successful Close" processing is enabled and
+	// the file will be persisted only when it has been explicitly closed.
+	StatIsPOSCPending StatFlags = 64
+)
+
+// EntryStat holds the entry name and the entry stat information.
+type EntryStat struct {
+	EntryName   string    // EntryName is the name of entry.
+	HasStatInfo bool      // HasStatInfo indicates if the following stat information is valid.
+	ID          int64     // ID is the OS-dependent identifier assigned to this entry.
+	EntrySize   int64     // EntrySize is the decimal size of the entry.
+	Flags       StatFlags // Flags identifies the entry's attributes.
+	Mtime       int64     // Mtime is the last modification time in Unix time units.
+}
+
+// Name implements os.FileInfo.
+func (es EntryStat) Name() string {
+	return es.EntryName
+}
+
+// Size implements os.FileInfo.
+func (es EntryStat) Size() int64 {
+	return es.EntrySize
+}
+
+// ModTime implements os.FileInfo.
+func (es EntryStat) ModTime() time.Time {
+	return time.Unix(es.Mtime, 0)
+}
+
+// Sys implements os.FileInfo.
+func (es EntryStat) Sys() interface{} {
+	return nil
+}
+
+// Mode implements os.FileInfo.
+func (es EntryStat) Mode() os.FileMode {
+	var mode os.FileMode
+	if es.IsDir() {
+		mode |= os.ModeDir
+	}
+	if es.IsWritable() {
+		mode |= 0222
+	}
+	if es.IsReadable() {
+		mode |= 0444
+	}
+	return mode
+}
+
+// IsExecutable indicates whether this entry is either an executable file or a searchable directory.
+func (es EntryStat) IsExecutable() bool {
+	return es.Flags&StatIsExecutable != 0
+}
+
+// IsDir indicates whether this entry is a directory.
+func (es EntryStat) IsDir() bool {
+	return es.Flags&StatIsDir != 0
+}
+
+// IsOther indicates whether this entry is neither a file nor a directory.
+func (es EntryStat) IsOther() bool {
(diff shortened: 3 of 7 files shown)


xrootd/protocol/protocol.go
```diff
@@ -85,3 +87,67 @@ const (
...
+		IsReadable:    flags&16 > 0,
+		IsWritable:    flags&32 > 0,
+		IsPoscPending: flags&64 > 0,
+	}
+
+	return result, nil
+}
+
+// NewEntryStatWithoutInfo creates a new EntryStat according to the name of the entry and without any stat info.
+func NewEntryStatWithoutInfo(name string) EntryStat {
```
sbinet: this seems to be equivalent to:
```go
es := EntryStat{Name:name}
```
perhaps just drop this function?

xrootd/protocol/protocol.go
```diff
@@ -85,3 +87,67 @@ const (
...
+	IsPoscPending bool   // IsPoscPending indicates that the file was created with kXR_posc and has not yet been successfully closed.
+	Mtime         int64  // Mtime is the last modification time in Unix time units.
+}
+
+// NewEntryStat creates a new EntryStat according to the name of the entry
+// and statinfo following XRootD protocol specification (see page 114).
+func NewEntryStat(name, statinfo string) (EntryStat, error) {
+	stats := strings.Split(statinfo, " ")
+
+	id, err := strconv.Atoi(stats[0])
```
sbinet: we should probably test that `stats` has the expected length instead of (implicitly) panicking otherwise.

xrootd/client/filesystem.go
```diff
@@ -0,0 +1,69 @@
...
+
+	result.Data = bytes.TrimRight(result.Data, "\x00")
+	linesBytes := bytes.Split(result.Data, []byte{'\n'})
+	lines := make([]string, len(linesBytes))
+
+	for i, v := range linesBytes {
+		lines[i] = string(v)
+	}
+
+	if len(lines) < 2 || lines[0] != "." || lines[1] != "0 0 0 0" {
```
sbinet: perhaps use:
```go
if !bytes.HasPrefix(linesBytes, []byte(".\n0 0 0 0\n")) { ... }
```
?

sbinet:
wrt `"/tmp/.xrootd"`: I've created a little fs hierarchy over at `ccxrootdgotest.in2p3.fr`:

```
$> tree /tmp
/tmp
├── dir1
│   └── file1.txt
├── file1.txt
```

xrootd/client/filesystem.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"bytes"
+	"context"
+
+	"github.com/pkg/errors"
+	"go-hep.org/x/hep/xrootd/protocol"
+	"go-hep.org/x/hep/xrootd/protocol/dirlist"
+)
+
+// Filesystem contains filesystem-related methods of the XRootD protocol.
```
sbinet: to piggy-back on the stdlib (and on `go-fuse`):
```
$> go doc net/http.FileSystem 
type FileSystem interface {
	Open(name string) (File, error)
}
    A FileSystem implements access to a collection of named files. The elements
    in a file path are separated by slash ('/', U+002F) characters, regardless
    of host operating system convention.

```

I would `s/Filesystem/FileSystem/`

xrootd/protocol/dirlist/dirlist.go
```diff
@@ -0,0 +1,54 @@
...
+package dirlist // import "go-hep.org/x/hep/xrootd/protocol/dirlist"
+
+// RequestID is the id of the request, it is sent as part of message.
+// See xrootd protocol specification for details: http://xrootd.org/doc/dev45/XRdv310.pdf, 2.3 Client Request Format.
+const RequestID uint16 = 3004
+
+// Response is a response for the dirlist request,
+// which contains a byte array containing encoded response.
+// The format (if stat information is supported by the server) is:
+// ".\n"
```
sbinet: please indent these by a couple of spaces so they are properly formatted in `godoc`:

```go
// The format (if stat information is supported by the server) is:
//     ".\n"
//     "0 0 0 0\n"
//     "dirname\n"
//     "id size flags modtime\n"
//     ...
//     0
// 
// In case that the server doesn't support returning the stat information, the format is:
//     "dirname\n"
//     ...
//     0
// 
// See xrootd protocol specification, page 45 for further details.
```

xrootd/protocol/dirlist/dirlist.go
```diff
@@ -0,0 +1,54 @@
...
+// for dirlist request used to obtain the contents of a directory.
+package dirlist // import "go-hep.org/x/hep/xrootd/protocol/dirlist"
+
+// RequestID is the id of the request, it is sent as part of message.
+// See xrootd protocol specification for details: http://xrootd.org/doc/dev45/XRdv310.pdf, 2.3 Client Request Format.
+const RequestID uint16 = 3004
+
+// Response is a response for the dirlist request,
+// which contains a byte array containing encoded response.
+// The format (if stat information is supported by the server) is:
```
sbinet: `s/The format (if stat information is supported by the server) is/When stat information is supported by the server, the format is:/`

xrootd/protocol/dirlist/dirlist.go
```diff
@@ -0,0 +1,54 @@
...
+// ".\n"
+// "0 0 0 0\n"
+// "dirname\n"
+// "id size flags modtime\n"
+// ...
+// 0
+// In case that the server doesn't support returning the stat information, the format is:
```
sbinet: `s/In case that the server doesn't support returning the stat information, the format is:/Otherwise, the format is the following:/`

xrootd/protocol/dirlist/dirlist.go
sbinet: that said, it's perhaps a bit too low-level (or nitty-gritty details.)

Ideally, the `Response` should come out of the pipe with a slice of `EntryStat`...
(that should probably be done by `func (resp *Response) UnmarshalXrd(data []byte) error`)

xrootd/protocol/protocol.go
```diff
@@ -85,3 +87,63 @@ const (
 	// DataServer indicates whether this is a data server.
 	DataServer
 )
+
+// EntryStat holds the entry name and the entry stat information.
+type EntryStat struct {
```
sbinet: it would be great if we could have `EntryStat` implement `os.FileInfo`:
```
$> go doc os.FileInfo
type FileInfo interface {
	Name() string       // base name of the file
	Size() int64        // length in bytes for regular files; system-dependent for others
	Mode() FileMode     // file mode bits
	ModTime() time.Time // modification time
	IsDir() bool        // abbreviation for Mode().IsDir()
	Sys() interface{}   // underlying data source (can return nil)
}
    A FileInfo describes a file and is returned by Stat and Lstat.

```
or at least devise some way for it to interoperate with `os.FileInfo`.

xrootd/protocol/protocol.go
```diff
@@ -85,3 +87,63 @@ const (
...
+	Name          string // Name is the name of entry.
+	HasStatInfo   bool   // HasStatInfo indicates if the following stat information is valid.
+	Id            int64  // Id is the OS-dependent identifier assigned to this entry.
```
sbinet: `s/Id/ID/`

xrootd/protocol/protocol.go
```diff
@@ -85,3 +87,63 @@ const (
...
+		return EntryStat{}, err
+	}
+
+	result := EntryStat{
+		Name:          name,
+		HasStatInfo:   true,
+		Id:            int64(id),
+		Size:          uint64(size),
+		Mtime:         int64(mtime),
+		IsExecutable:  flags&1 > 0,
```
sbinet: `s/>/!=/` ?
at least, that's usually the canonical way how we test flags.

EgorMatirov:
> I think it's maybe time we implement the Xrd(Un)Marshaler stuff.
> it would consolidate greatly this last PR :)
> WDYT?

Yep, I'll start working on it then. I have partially implemented `Open` request and it looks like it would benefit from the generator too.

I'll use interfaces from `xrootd-marshaller` and code from the `brio-gen` as a base.

xrootd/protocol/dirlist/dirlist.go
```diff
@@ -0,0 +1,54 @@
...
+type RequestOptions byte
+
+const (
+	None         RequestOptions = 0 // None specifies that no addition information except entry names is required.
+	WithStatInfo RequestOptions = 2 // WithStatInfo specifies that stat information for every entry is required.
+)
+
+// NewRequest forms a Request according to provided path.
+func NewRequest(path string) Request {
+	var pathBytes = make([]byte, len(path))
```
sbinet: this could be simply rewritten as:
```go
pathBytes := []byte(path)
```

EgorMatirov:
I have updated PR with the version built on top of #214.

xrootd/client/filesystem_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	got, err := fs.Dirlist(context.Background(), "/tmp/dir1")
+	if err != nil {
+		t.Fatalf("invalid protocol call: %v", err)
+	}
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("FileSystem.Dirlist()\ngot = %v\nwant = %v", got, want)
+	}
+
+	client.Close()
```
sbinet: couldn't this be moved up to line 34 as a `defer client.Close()` ?

xrootd/protocol/protocol.go
```diff
@@ -102,3 +106,97 @@ const (
...
+		return EntryStat{}, err
+	}
+
+	result := EntryStat{
+		EntryName:     name,
+		HasStatInfo:   true,
+		ID:            int64(id),
+		EntrySize:     int64(size),
+		Mtime:         int64(mtime),
+		IsExecutable:  flags&1 != 0,
```
sbinet: perhaps store `flags` as an `EntryStat` field and convert all these `IsXyz` fields into methods on `EntryStat`?

xrootd/protocol/protocol.go
```diff
@@ -102,3 +106,97 @@ const (
...
+	HasStatInfo   bool   // HasStatInfo indicates if the following stat information is valid.
+	ID            int64  // ID is the OS-dependent identifier assigned to this entry.
+	EntrySize     int64  // EntrySize is the decimal size of the entry.
+	IsExecutable  bool   // IsExecutable indicates that entry is either an executable file or a searchable directory.
+	IsDirectory   bool   // IsDirectory indicates that entry is a directory.
+	IsOther       bool   // IsOther indicates that entry is neither a file nor a directory.
+	IsOffline     bool   // IsOffline indicates that the file is not online (i. e., on disk).
+	IsReadable    bool   // IsReadable indicates that read access to that entry is allowed.
+	IsWritable    bool   // IsWritable indicates that write access to that entry is allowed.
+	IsPoscPending bool   // IsPoscPending indicates that the file was created with kXR_posc and has not yet been successfully closed.
```
sbinet: the meaning of a file opened with `kXR_posc` could be perhaps explained here (or on the method, if that field is converted into a method)

sbinet:
breaking an import cycle can be done by either:
a) merge a few of the involved packages together (so no more cycle to speak of)
b) introduce another package to break the cycle

I went with `b)` when I introduced `internal/xrdenc` because:
- `client` imports `protocol`
- `client` import `protocol/{handshake,login,protocol}`
- `protocol/{handshake,login,protocol}` import `protocol` for the marshaling.
I broke the cycle by lifting the marshaling (encoder/decoder) types into a new package.

usually, an import cycle means the granularity of the packages is too fine or, the packages' boundaries are the wrong ones.

perhaps the `protocol/{handshake,login,protocol}` packages are too special and too tightly related to `protocol` (a.k.a `xrdproto`) and ought to be merged into it?
perhaps we should have `xrdproto` and `client` merged into one package? (but leave each req/resp into its own package)

I don't know.

(also: could you please rebase this PR ? thanks.)

sbinet:
perhaps put the `EntryStat`, `FileSystem` and `FileXyz` types into a new `xrdfs` package?
that would break the import cycle.

sbinet:
in the meantime, I'll create a rebased `xrootd-dirlist` branch, on top of #216 .

sbinet:
done.

EgorMatirov:
Hello.
Thank you for your help!

I have had exams again. :( (However, I successfully passed them).

I have updated the `xrootd-dirlist` branch to follow your suggestions, however, I want to ask you if I can merge `xrootd-xrdfs` into this branch? If so, I'll merge them and enable correct signing for `dirlist`.

I'll be working on https://github.com/go-hep/hep/pull/215 in the meantime, rebasing it on top of `xrootd-xrdfs`.

sbinet:
> I have had exams again. :( (However, I successfully passed them). 

that's certainly good news :)

> however, I want to ask you if I can merge `xrootd-xrdfs` into this branch?

sure. like last time, I wanted to see what it would look like and whether there would be any roadblock ahead :)

xrootd/client/filesystem_test.go
```diff
@@ -0,0 +1,52 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build xrootd_test_with_server
```
sbinet: let's remove this build tag to ensure Travis is testing against the C++ server :)

xrootd/xrdfs/stat.go
```diff
@@ -0,0 +1,180 @@
...
+// IsPOSCPending indicates whether this the file was created with kXR_posc and has not yet been successfully closed.
+// kXR_posc is an option of open request indicating that the "Persist On Successful Close" processing is enabled and
+// the file will be persisted only when it has been explicitly closed.
+func (es EntryStat) IsPOSCPending() bool {
+	return es.Flags&StatIsPOSCPending != 0
+}
+
+// MarshalXrd implements xrdproto.Marshaler.
+func (o EntryStat) MarshalXrd(wBuffer *xrdenc.WBuffer) error {
+	if !o.HasStatInfo {
```
sbinet: shouldn't `EntryName` be written out?

EgorMatirov: It should be written out in the `dirlist` response, yes. However, in `open` and `stat` responses entry name isn't specified. So, I thought that we can leave that to the specific response Marshaler.

PR #210, opened 2018-05-30
Xrootd marshaler

sbinet:
it's still a WIP.
but gives a general idea of what I think ATM it should look like.
(not sure yet on the `type RequestID uint16` part)

PR #214, opened 2018-06-01
xrootd: introduce a XRootD (un)marshal interface and generator

EgorMatirov:
This helps to generate boilerplate code which is used for encoding
and decoding of the requests and the responses.

It's a bit involved version of the #210.
@sbinet, thank you very much for your help!


xrootd/internal/xrdenc/xrdenc.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (enc *Encoder) WriteI64(v int64) {
+	var buf [8]byte
+	binary.BigEndian.PutUint64(buf[:], uint64(v))
+	enc.buf = append(enc.buf, buf[:]...)
+}
+
+func (enc *Encoder) WriteBool(v bool) {
+	if v {
+		enc.buf = append(enc.buf, 1)
```
sbinet: to facilitate inlining, this probably should be written like:
```go
if v {
    enc.buf = append(enc.buf, 1)
    return
}
enc.buf = append(enc.buf, 0)
```

xrootd/internal/xrdenc/xrdenc.go
```diff
@@ -0,0 +1,143 @@
...
+func (enc *Encoder) WriteBytes(vs []byte) {
+	enc.buf = append(enc.buf, vs...)
+}
+
+func (enc *Encoder) WriteStr(str string) {
+	enc.WriteLen(len(str))
+	enc.WriteBytes([]byte(str))
+}
+
+func (enc *Encoder) WriteReserved(n int) {
```
sbinet: looking back at this, it seems I wasn't completely super inspired when coming up with `WriteReserved` as a name.

Perhaps `WritePadding` ?
(not completely happy with that either... what about you?)

xrootd/internal/xrdenc/xrdenc.go
```diff
@@ -0,0 +1,143 @@
...
+	return &Decoder{buf: data}
+}
+
+func (dec *Decoder) ReadU8() uint8 {
+	o := dec.buf[dec.pos]
+	dec.pos++
+	return o
+}
+
+func (dec *Decoder) Available() int {
```
sbinet: to mimic other Go's stdlib APIs, I would rename this `Len` or `Size`:

```
$> go doc bytes.Buffer.Len
func (b *Buffer) Len() int
    Len returns the number of bytes of the unread portion of the buffer; b.Len()
    == len(b.Bytes()).

$> go doc bytes.Reader.Len
func (r *Reader) Len() int
    Len returns the number of bytes of the unread portion of the slice.

$> go doc bytes.Reader.Size 
func (r *Reader) Size() int64
    Size returns the original length of the underlying byte slice. Size is the
    number of bytes available for reading via ReadAt. The returned value is
    always the same and is not affected by calls to any other method.

```

(so.. perhaps `Len` fits better)

xrootd/internal/xrdenc/xrdenc.go
```diff
@@ -0,0 +1,143 @@
...
+	beg := dec.pos
+	end := dec.pos + 8
+	dec.pos += 8
+	o := binary.BigEndian.Uint64(dec.buf[beg:end])
+	return int64(o)
+}
+
+func (dec *Decoder) ReadBool() bool {
+	dec.pos += 1
+	return dec.buf[dec.pos] == 1
```
sbinet: `s/== 1/!= 0/`

the result is the same (d'oh!) but it's more in line with what old C brains would expect :)

xrootd/protocol/login/login.go
```diff
@@ -29,31 +31,76 @@ type Response struct {
...
+	o.SecurityInformation = append(o.SecurityInformation, dec.Bytes()...)
+	return dec.Pos(), err
+}
+
 // Request holds the login request parameters.
 type Request struct {
 	Pid          int32   // Pid is the process number associated with this connection.
 	Username     [8]byte // Username is the unauthenticated name of the user to be associated with the connection.
-	Reserved     byte    // Reserved is an area reserved for future use.
+	_            byte    // Reserved is an area reserved for future use.
```
sbinet: `s/Reserved is an area reserved for future use./Reserved for future use./`

xrootd/protocol/marshaling.go
```diff
@@ -87,6 +117,11 @@ func Marshal(x interface{}) ([]byte, error) {
 // Since the length of the slice is unknown, all bytes to the end of data is copied to it.
 // Supported types are: uint8, uint16, int32, int64, slices and arrays of uint8.
 func Unmarshal(data []byte, x interface{}) (err error) {
+	if x, ok := x.(Unmarshaler); ok {
+		_, err := x.UnmarshalXrd(data)
```
sbinet: `s/:=/=/`
(to prevent any unnecessary variable shadowing...)

xrootd/protocol/protocol/protocol.go
```diff
@@ -7,134 +7,234 @@
...
-	// as well as certain non-destructive requests need to be signed.
-	Standard SecurityLevel = 2
-	// Intense indicates that request that may reveal metadata or modify data need to be signed.
-	Intense SecurityLevel = 3
-	// Pedantic indicates that all requests need to be signed.
-	Pedantic SecurityLevel = 4
-)
+// SetIsManager sets whether this server has manager role.
+func (resp *Response) SetIsManager(v bool) {
+	if v {
```
sbinet: please use an early `return` to help the compiler inline calls of this method.

xrootd/protocol/protocol/protocol.go
```diff
@@ -7,134 +7,234 @@
...
+}
 
-const (
-	SignNone   RequestLevel = 0 // SignNone indicates that the request need not to be signed.
-	SignLikely RequestLevel = 1 // SignLikely indicates that the request must be signed if it modifies data.
-	SignNeeded RequestLevel = 2 // SignNeeded indicates that the request mush be signed.
-)
+// SetIsServer sets whether this server has server role.
+func (resp *Response) SetIsServer(v bool) {
+	if v {
```
sbinet: please use an early `return` to help the compiler inline calls of this method.

xrootd/protocol/protocol/protocol.go
```diff
@@ -7,134 +7,234 @@
...
-// It consists of the request index and the security requirement the associated request should have.
-// Request index is calculated as:
-//     (request code) - (request code of Auth request)
-// according to xrootd protocol specification.
-type SecurityOverride struct {
-	RequestIndex byte
-	RequestLevel RequestLevel
+// SetIsMeta sets whether this server has meta attribute.
+func (resp *Response) SetIsMeta(v bool) {
+	if v {
```
sbinet: please use an early `return` to help the compiler inline calls of this method.

xrootd/protocol/protocol/protocol.go
```diff
@@ -7,134 +7,234 @@
...
-const (
-	// RequestOptionsNone specifies that only general response should be returned.
-	RequestOptionsNone RequestOptions = 0
-	// ReturnSecurityRequirements specifies that security requirements should be returned
-	// if that's supported by the server.
-	ReturnSecurityRequirements RequestOptions = 1
-)
+// SetIsProxy sets whether this server has proxy attribute.
+func (resp *Response) SetIsProxy(v bool) {
+	if v {
```
sbinet: please use an early `return` to help the compiler inline calls of this method.

xrootd/protocol/protocol/protocol.go
```diff
@@ -7,134 +7,234 @@
...
 }
 
-// NewRequest forms a Request according to provided parameters.
-func NewRequest(protocolVersion int32, withSecurityRequirements bool) Request {
-	var options = RequestOptionsNone
-	if withSecurityRequirements {
-		options |= ReturnSecurityRequirements
+// SetIsSupervisor sets whether this server has supervisor attribute.
+func (resp *Response) SetIsSupervisor(v bool) {
+	if v {
```
sbinet: please use an early `return` to help the compiler inline calls of this method.

xrootd/protocol/protocol/protocol.go
```diff
@@ -7,134 +7,234 @@
...
+
+// ForceSecurity indicates whether signing is required even if the authentication
+// protocol does not support generic encryption.
+func (resp *Response) ForceSecurity() bool {
+	return resp.SecurityOptions&forceSecurity != 0
+}
+
+// SetForceSecurity sets whether signing is forced.
+func (resp *Response) SetForceSecurity(v bool) {
+	if v {
```
sbinet: please use an early `return` to help the compiler inline calls of this method.

xrootd/protocol/protocol/protocol.go
```diff
@@ -7,134 +7,234 @@
...
+	dec := xrdenc.NewDecoder(data)
+	o.ClientProtocolVersion = dec.ReadI32()
+	o.Options = requestOptions(dec.ReadU8())
+	dec.Skip(11)
+	dec.Skip(4)
+	return dec.Pos(), err
+}
+
+// Response is a response for the `Protocol` request. See details in the xrootd protocol specification.
+type Response struct {
```
sbinet: somehow, the `SetIsXyz` methods on this `protocol.Response` type feel a bit wrong to me: calling those won't impact back the server from which we got them.
I understand you introduced them so you could create a `Response` to test against in the `Client` tests.

Perhaps we could just leave the `flags` field of `Response` exported (so `s/flags/Flags/`) and re-export the `IsXyz` bitmasks ?
the `protocol` package isn't super crowded so it's fine to populate it somewhat.

that way, you could create a `want` value like so:
```go
want := protocol.Response{
￼	BinaryProtocolVersion: protocolVersion,
￼	Flags:                 protocol.IsManager | protocol.IsServer | protocol.IsMeta |
                               protocol.IsProxy   | protocol.IsSupervisor,
	SecurityLevel:         xrdproto.Pedantic,
	SecurityOverrides:     []xrdproto.SecurityOverride{{1, xrdproto.SignNeeded}},
}
```

EgorMatirov: >I understand you introduced them so you could create a Response to test against in the Client tests.

was thinking about using them for the server implementation, not just tests. However, I don't argue about usage of `Flags`, it's fine too. :)

xrootd/internal/xrdenc/xrdenc.go
EgorMatirov: Yep, both `Size` and `Available` looked awkward to me and I haven't noticed that `Len` is about unread portion. Thanks!

xrootd/client/protocol_mock_test.go
```diff
@@ -41,47 +53,23 @@ func TestClient_Protocol_WithSecurityInfo(t *testing.T) {
...
+			DataLength: 14 + xrdproto.SecurityOverrideLength,
 		}
 
-		securityOverride := protocol.SecurityOverride{1, protocol.SignNeeded}
-
-		response, err := marshalResponse(responseHeader, protocolResponse, protocolSecurityInfo, securityOverride)
+		response, err := marshalResponse(responseHeader, want)
 		if err != nil {
-			cancel()
+			defer cancel()
```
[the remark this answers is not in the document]
EgorMatirov: Sorry, looks like a strange merge from somewhere (Have no idea how it leaked here...).

xrootd/protocol/marshaling.go
```diff
@@ -11,6 +11,32 @@ import (
...
+}
+
+// Unmarshaler is the interface implemented by a type that can
+// unmarshal a binary representation of itself, following the XRootD protocol.
+//
+// UnmarshalXrd must be able to decode the form generated by MarshalXrd.
+// UnmarshalXrd must copy the data if it wishes to retain the data after
+// returning.
+type Unmarshaler interface {
+	UnmarshalXrd(data []byte) (int, error)
```
sbinet: ok... to address this annoying (to me, at least) issue of this extra `int` return value, what about modifying the (un)marshalers to something like:
```go
type Marshaler interface {
    MarshalXrd(buf *Buffer) error
}

type Unmarshaler interface {
    UnmarshalXrd(buf *Buffer) error
}
```

where `Buffer` is the union of the current `Encoder` and `Decoder` type.

that way, the current issue of unmarshaling `protocol/protocol.Request`:
```go
// UnmarshalXrd implements xrootd/protocol.Unmarshaler
func (o *Response) UnmarshalXrd(data []byte) (pos int, err error) {
	dec := xrdenc.NewDecoder(data)
	o.BinaryProtocolVersion = dec.ReadI32()
	o.flags = flags(dec.ReadI32())
	if dec.Available() > 0 {
		dec.Skip(1)
		dec.Skip(1)
		o.SecurityVersion = dec.ReadU8()
		o.SecurityOptions = securityOptions(dec.ReadU8())
		o.SecurityLevel = protocol.SecurityLevel(dec.ReadU8())
		o.SecurityOverrides = make([]protocol.SecurityOverride, dec.ReadU8())
		for i := 0; i < len(o.SecurityOverrides); i++ {
			n, err := o.SecurityOverrides[i].UnmarshalXrd(dec.Bytes())
			if err != nil {
				return 0, err
			}
			dec.Skip(n)
		}
	}
	return dec.Pos(), err
}
```
would become:
```go
// UnmarshalXrd implements xrootd/protocol.Unmarshaler
func (o *Response) UnmarshalXrd(buf *xrdproto.Buffer) error {
	o.BinaryProtocolVersion = buf.ReadI32()
	o.flags = flags(buf.ReadI32())
	if buf.Len() > 0 {
		buf.Skip(1)
		buf.Skip(1)
		o.SecurityVersion = buf.ReadU8()
		o.SecurityOptions = securityOptions(buf.ReadU8())
		o.SecurityLevel = protocol.SecurityLevel(buf.ReadU8())
		o.SecurityOverrides = make([]protocol.SecurityOverride, buf.ReadU8())
		for i := 0; i < len(o.SecurityOverrides); i++ {
			err := o.SecurityOverrides[i].UnmarshalXrd(buf)
			if err != nil {
				return err
			}
		}
	}
	return nil
}
```

most notably, the `dec.Skip(n)` statement (which could easily be overlooked and fogotten) just disappeared.

what do you think?

xrootd/protocol/protocol/protocol.go
```diff
@@ -7,134 +7,234 @@
...
+	}
+	return enc.Bytes(), err
+}
+
+// UnmarshalXrd implements xrootd/protocol.Unmarshaler
+func (o *Response) UnmarshalXrd(data []byte) (pos int, err error) {
+	dec := xrdenc.NewDecoder(data)
+	o.BinaryProtocolVersion = dec.ReadI32()
+	o.flags = flags(dec.ReadI32())
+	if dec.Available() > 0 {
```
sbinet: please invert the logic and apply an early `return`.

xrootd/protocol/protocol/protocol.go
```diff
@@ -7,134 +7,234 @@
...
+	} else {
+		resp.SecurityOptions &^= forceSecurity
+	}
+}
+
+// MarshalXrd implements xrootd/protocol.Marshaler
+func (o Response) MarshalXrd() (data []byte, err error) {
+	var enc xrdenc.Encoder
+	enc.WriteI32(o.BinaryProtocolVersion)
+	enc.WriteI32(int32(o.flags))
```
sbinet: hum... why isn't the marshaling handling the optional writing part, symmetric to what is done on the unmarshaling side?

xrootd/internal/xrdenc/xrdenc.go
EgorMatirov: What do you think about `Skip` cause what we really do is skipping some of the bytes? However, I'm not sure how it fits in with other `WriteXXX` methods...

xrootd/protocol/marshaling.go
EgorMatirov: What do you think about keeping `Encoder` and `Decoder` and passing them instead of `xrdproto.Buffer`? Because we don't need to perform any writing in `UnmarshalXrd` and any reading in `MarshalXrd`.
So it will look like:
```go
type Marshaler interface {
	MarshalXrd(enc *xrdenc.Encoder) error
}

type Unmarshaler interface {
	UnmarshalXrd(dec *xrdenc.Decoder) error
}
```

sbinet: Then, what about renaming 'Decoder' into 'RBuffer' and 'Encoder' into 'WBuffer'?

sbinet: Then, what about renaming 'Decoder' into 'RBuffer' and 'Encoder' into 'WBuffer'?

EgorMatirov: Sounds good!

xrootd/internal/xrdenc/xrdenc.go
sbinet: If we go with the RBuffer, then I'd be tempted to rename it as 'Next', to follow 'bytes.Buffer' lead.

PR #215, opened 2018-06-02
xrootd: add open request

EgorMatirov:
Updates go-hep/hep#170 (haven't included in the commit message since I expect rebases because of https://github.com/go-hep/hep/pull/214 and https://github.com/go-hep/hep/pull/211 and it will pollute issue with each commit)

Note that `open` request isn't added to `signing.go` because it creates a cycle import. Any suggestions on how to deal with that? Probably we can move `signing` to `protocol/signing` to fix that?

Also, note that there is a test workaround due to https://github.com/xrootd/xrootd/issues/721.

Also, I'm unsure about `UnmarshalXrd` in `EntryStat` and `NewEntryStat`. It duplicates logic, but otherwise  call to `UnmarshalXrd` in `dirlist.go`, `(Response) UnmarshalXrd(...)` will be like that:
```go
o.Entries[i/2].UnmarshalXrd(xrdenc.NewRBuffer([]byte(lines[i+1)]))
```

### DIFF
--- a/xrootd/client/file.go
+++ b/xrootd/client/file.go
@@ -1,0 +1,35 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package client // import "go-hep.org/x/hep/xrootd/client"
+
+import "go-hep.org/x/hep/xrootd/xrdfs"
+
+// File implements access to a content and meta information of file over XRootD.
+type file struct {
+	fs          *fileSystem
+	handle      xrdfs.FileHandle
+	compression *xrdfs.FileCompression
+	info        *xrdfs.EntryStat
+}
+
+// Compression returns the compression info.
+func (f file) Compression() *xrdfs.FileCompression {
+	return f.compression
+}
+
+// Info returns the cached stat info.
+// Note that it may return nil if info was not yet fetched and info may be not up-to-date.
+func (f file) Info() *xrdfs.EntryStat {
+	return f.info
+}
+
+// Handle returns the file handle.
+func (f file) Handle() xrdfs.FileHandle {
+	return f.handle
+}
+
+var (
+	_ xrdfs.File = (*file)(nil)
+)
--- a/xrootd/client/filesystem.go
+++ b/xrootd/client/filesystem.go
@@ -9,6 +9,7 @@ import (
 
 	"go-hep.org/x/hep/xrootd/xrdfs"
 	"go-hep.org/x/hep/xrootd/xrdproto/dirlist"
+	"go-hep.org/x/hep/xrootd/xrdproto/open"
 )
 
 // FS returns a xrdfs.FileSystem which uses this client to make requests.
@@ -29,6 +30,16 @@ func (fs *fileSystem) Dirlist(ctx context.Context, path string) ([]xrdfs.EntrySt
 		return nil, err
 	}
 	return resp.Entries, err
+}
+
+// Open returns the file handle for a file together with the compression and the stat info.
+func (fs *fileSystem) Open(ctx context.Context, path string, mode xrdfs.OpenMode, options xrdfs.OpenOptions) (xrdfs.File, error) {
+	var resp open.Response
+	err := fs.c.Send(ctx, &resp, open.NewRequest(path, mode, options))
+	if err != nil {
+		return nil, err
+	}
+	return &file{fs, resp.FileHandle, resp.Compression, resp.Stat}, nil
 }
 
 var (
--- a/xrootd/client/filesystem_test.go
+++ b/xrootd/client/filesystem_test.go
@@ -48,3 +48,64 @@ func TestFileSystem_Dirlist(t *testing.T) {
 		})
 	}
 }
+
+func testFileSystem_Open(t *testing.T, addr string, options xrdfs.OpenOptions, wantFileHandle xrdfs.FileHandle, wantFileCompression *xrdfs.FileCompression, wantFileInfo *xrdfs.EntryStat) {
+	client, err := NewClient(context.Background(), addr, "gopher")
+	if err != nil {
+		t.Fatalf("could not create client: %v", err)
+	}
+	defer client.Close()
+
+	fs := client.FS()
+
+	gotFile, err := fs.Open(context.Background(), "/tmp/dir1/file1.txt", xrdfs.OpenModeOtherRead, options)
+	if err != nil {
+		t.Fatalf("invalid open call: %v", err)
+	}
+
+	if !reflect.DeepEqual(gotFile.Handle(), wantFileHandle) {
+		t.Errorf("Filesystem.Open()\ngotFile.Handle() = %v\nwantFileHandle = %v", gotFile.Handle(), wantFileHandle)
+	}
+
+	if !reflect.DeepEqual(gotFile.Compression(), wantFileCompression) {
+		// TODO: Remove this workaround when fix for https://github.com/xrootd/xrootd/issues/721 will be released.
+		skippedDefaultCompressionValue := reflect.DeepEqual(wantFileCompression, &xrdfs.FileCompression{}) && gotFile.Compression() == nil
+		if !skippedDefaultCompressionValue {
+			t.Errorf("Filesystem.Open()\ngotFile.Compression() = %v\nwantFileCompression = %v", gotFile.Compression(), wantFileCompression)
+		}
+	}
+
+	if !reflect.DeepEqual(gotFile.Info(), wantFileInfo) {
+		t.Errorf("Filesystem.Open()\ngotFile.Info() = %v\nwantFileInfo = %v", gotFile.Info(), wantFileInfo)
+	}
+}
+
+func TestFileSystem_Open(t *testing.T) {
+	emptyCompression := xrdfs.FileCompression{}
+	entryStat := &xrdfs.EntryStat{
+		HasStatInfo: true,
+		ID:          60129606914,
+		Mtime:       1528218208,
+		Flags:       xrdfs.StatIsWritable | xrdfs.StatIsReadable,
+	}
+
+	testCases := []struct {
+		name        string
+		options     xrdfs.OpenOptions
+		handle      xrdfs.FileHandle
+		compression *xrdfs.FileCompression
+		info        *xrdfs.EntryStat
+	}{
+		{"WithoutCompressionAndStat", xrdfs.OpenOptionsOpenRead, xrdfs.FileHandle{0, 0, 0, 0}, nil, nil},
+		{"WithCompression", xrdfs.OpenOptionsOpenRead | xrdfs.OpenOptionsCompress, xrdfs.FileHandle{0, 0, 0, 0}, &emptyCompression, nil},
+		{"WithStat", xrdfs.OpenOptionsOpenRead | xrdfs.OpenOptionsReturnStatus, xrdfs.FileHandle{0, 0, 0, 0}, &emptyCompression, entryStat},
+	}
+
+	for _, addr := range testClientAddrs {
+		for _, tc := range testCases {
+			t.Run(addr+"/"+tc.name, func(t *testing.T) {
+				testFileSystem_Open(t, addr, tc.options, tc.handle, tc.compression, tc.info)
+			})
+		}
+	}
+}
--- a/xrootd/xrdfs/file.go
+++ b/xrootd/xrdfs/file.go
@@ -1,0 +1,43 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package xrdfs
+
+import (
+	"go-hep.org/x/hep/xrootd/internal/xrdenc"
+)
+
+// File implements access to a content and meta information of file over XRootD.
+type File interface {
+	// Compression returns the compression info.
+	Compression() *FileCompression
+	// Info returns the cached stat info.
+	// Note that it may return nil if info was not yet fetched and info may be not up-to-date.
+	Info() *EntryStat
+	// Handle returns the file handle.
+	Handle() FileHandle
+}
+
+// FileHandle is the file handle, which should be treated as opaque data.
+type FileHandle [4]byte
+
+// FileCompression holds the compression parameters such as the page size and the type of compression.
+type FileCompression struct {
+	PageSize int32
+	Type     [4]byte
+}
+
+// MarshalXrd implements xrdproto.Marshaler
+func (o FileCompression) MarshalXrd(wBuffer *xrdenc.WBuffer) error {
+	wBuffer.WriteI32(o.PageSize)
+	wBuffer.WriteBytes(o.Type[:])
+	return nil
+}
+
+// UnmarshalXrd implements xrdproto.Unmarshaler
+func (o *FileCompression) UnmarshalXrd(rBuffer *xrdenc.RBuffer) error {
+	o.PageSize = rBuffer.ReadI32()
+	rBuffer.ReadBytes(o.Type[:])
+	return nil
+}
--- a/xrootd/xrdfs/fs.go
+++ b/xrootd/xrdfs/fs.go
@@ -12,4 +12,61 @@ import (
 // FileSystem implements access to a collection of named files over XRootD.
 type FileSystem interface {
 	Dirlist(ctx context.Context, path string) ([]EntryStat, error)
+	Open(ctx context.Context, path string, mode OpenMode, options OpenOptions) (File, error)
 }
+
+// OpenMode is the mode in which path is to be opened.
+// The mode is an "or`d" combination of ModeXyz flags.
+type OpenMode uint16
+
+const (
+	OpenModeOwnerRead    OpenMode = 0x100 // OpenModeOwnerRead indicates that owner has read access.
+	OpenModeOwnerWrite   OpenMode = 0x080 // OpenModeOwnerWrite indicates that owner has write access.
+	OpenModeOwnerExecute OpenMode = 0x040 // OpenModeOwnerExecute indicates that owner has execute access.
+
+	OpenModeGroupRead    OpenMode = 0x020 // OpenModeGroupRead indicates that group has read access.
+	OpenModeGroupWrite   OpenMode = 0x010 // OpenModeGroupWrite indicates that group has write access.
+	OpenModeGroupExecute OpenMode = 0x008 // OpenModeGroupExecute indicates that group has execute access.
+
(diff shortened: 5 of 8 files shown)


xrootd/client/filesystem_test.go
```diff
@@ -52,3 +53,128 @@ func TestFilesystem_Dirlist(t *testing.T) {
...
+	}
+
+	if !reflect.DeepEqual(gotFileCompression, wantFileCompression) {
+		t.Errorf("Filesystem.Open()\ngotFileCompression = %v\nwantFileCompression = %v", gotFileCompression, wantFileCompression)
+	}
+	if !reflect.DeepEqual(gotFileStat, wantFileStat) {
+		t.Errorf("Filesystem.Open()\ngotFileStat = %v\nwantFileStat = %v", gotFileStat, wantFileStat)
+	}
+
+	client.Close()
```
sbinet: could we move it up to line 66 as a `defer client.Close()` ?

xrootd/client/filesystem_test.go
```diff
@@ -52,3 +53,128 @@ func TestFilesystem_Dirlist(t *testing.T) {
...
+			// Workaround https://github.com/xrootd/xrootd/issues/721
+			t.Errorf("Filesystem.Open()\ngotFileCompression = %v\nwantFileCompression = %v", gotFileCompression, wantFileCompression)
+		}
+	}
+
+	if !reflect.DeepEqual(gotFileStat, wantFileStat) {
+		t.Errorf("Filesystem.Open()\ngotFileStat = %v\nwantFileStat = %v", gotFileStat, wantFileStat)
+	}
+
+	client.Close()
```
sbinet: could we move it up to line 104 as a `defer`?

xrootd/client/filesystem_test.go
```diff
@@ -52,3 +53,128 @@ func TestFilesystem_Dirlist(t *testing.T) {
...
+	}
+
+	if !reflect.DeepEqual(gotFileCompression, wantFileCompression) {
+		t.Errorf("Filesystem.Open()\ngotFileCompression = %v\nwantFileCompression = %v", gotFileCompression, wantFileCompression)
+	}
+	if !reflect.DeepEqual(gotFileStat, wantFileStat) {
+		t.Errorf("Filesystem.Open()\ngotFileStat = %v\nwantFileStat = %v", gotFileStat, wantFileStat)
+	}
+
+	client.Close()
```
sbinet: please move it up as a `defer` :)

xrootd/protocol/open/open.go
```diff
@@ -0,0 +1,157 @@
...
+	// OptionsForce specifies that file is opened ignoring  file usage rules.
+	OptionsForce Options = 4
+	// OptionsNew specifies that file is opened only if it does not already exist.
+	OptionsNew Options = 8
+	// OptionsOpenRead specifies that file is opened only for reading.
+	OptionsOpenRead Options = 16
+	// OptionsOpenUpdate specifies that file is opened only for reading and writing.
+	OptionsOpenUpdate Options = 32
+	// OptionsAsyncOpen specifies that file is opened for asynchronous i/o.
+	OptionsAsync Options = 64
```
sbinet: the doc says `OptionsAsyncOpen` but the `const` is called `OptionsAsync`.

xrootd/protocol/protocol.go
```diff
@@ -157,6 +158,91 @@ func (es EntryStat) Mode() os.FileMode {
...
+	}
+	flagsStr := strconv.Itoa(flags)
+	mtimeStr := strconv.Itoa(int(o.Mtime))
+
+	wBuffer.WriteBytes([]byte(idStr + " " + sizeStr + " " + flagsStr + " " + mtimeStr))
+	return nil
+}
+
+// UnmarshalXrd implements xrootd/protocol.Unmarshaler
+func (o *EntryStat) UnmarshalXrd(rBuffer *xrdenc.RBuffer) error {
```
sbinet: `dirlist` should probably be modified to reuse this code, right?
instead of writing it like:
```go
o.Entries[i/2].UnmarshalXrd(xrdenc.NewRBuffer([]byte(lines[i+1])))
```

I would do it like so:
```go
var rbuf =xrdenc.NewRBuffer(lines[i+1])
o.Entries[i/2].UnmarshalXrd(rbuf)
```
(we could have `dirlist` work in terms of `[][]byte` instead of converting to `[]string`.)

xrootd/client/filesystem.go
```diff
@@ -25,3 +26,14 @@ func (fs *FileSystem) Dirlist(ctx context.Context, path string) ([]protocol.Entr
 	}
 	return resp.Entries, err
 }
+
+// Open returns the file handle for a file together with the compression and the stat info.
+func (fs *FileSystem) Open(ctx context.Context, path string, mode open.Mode, options open.Options) (protocol.FileHandle, *protocol.FileCompression, *protocol.EntryStat, error) {
```
sbinet: looking at the returned types, I have the urge to merge them into a single type, `File`, which would eventually implement `io.Reader`, `io.Writer`, `io.Closer`, ... with 2 additional methods:
- one to return this `FileCompression`
- the other to return `os.FileInfo` that our `EntryStat` implements already.

but perhaps it's a bit too soon to introduce such "high-level" abstractions? (or the wrong "level"?)

still... this really "feels" right:
```go
func (fs *FileSystem) Open(ctx context.Context, 
     path string, mode open.Mode, options open.Options) 
     (xrootd.File, error) { ... }
```

EgorMatirov: I gave it a try. From my point of view, it looks fine, but I will be happy to hear your opinion. :)

sbinet:
as https://github.com/xrootd/xrootd/issues/721 has been fixed, could you add a blurb about that for the actions to take (if any) when a release with the fix percolates?
(also, if there are any actions to be taken on our side, please file an issue here so we don't forget :))

EgorMatirov:
Yep, I tweaked note a bit.

We haven't to do anything specific - just update the server after release and remove two lines (near that `TODO:` in the test) cause they will lose meaning. If you still want me to file the issue for that, then say and I'll be pleased to do that. :)

EgorMatirov:
Also, note that I have extracted an interface for `File` and moved file implementation to `Client` like it's done for `FileSystem`. It will help to implement a `Stat` request since it can operate on both the file path and the file handle.

So, there will be `Stat` in the `FileSystem` which returns `EntryStat` for the path and `FetchStat` in the `File` which will populate `File` with the fresh-fetched stat info by the file handle.

xrootd/client/filesystem.go
```diff
@@ -31,6 +32,17 @@ func (fs *fileSystem) Dirlist(ctx context.Context, path string) ([]xrdfs.EntrySt
...
 }
 
+// Open returns the file handle for a file together with the compression and the stat info.
+func (fs *fileSystem) Open(ctx context.Context, path string, mode xrdfs.OpenMode, options xrdfs.OpenOptions) (xrdfs.File, error) {
+	var resp open.Response
+	err := fs.c.Send(ctx, &resp, open.NewRequest(path, mode, options))
+	if err != nil {
+		return file{}, err
+	}
+
```
sbinet: please remove this blank line.

xrootd/xrdfs/file.go
```diff
@@ -0,0 +1,39 @@
...
+package xrdfs
+
+import (
+	"go-hep.org/x/hep/xrootd/internal/xrdenc"
+)
+
+// File implements access to a content and meta information of file over XRootD.
+type File interface {
+	Compression() *FileCompression // Compression returns the compression info.
+	Info() *EntryStat              // Info returns the stat info.
```
sbinet: same comment than on `client.file.Info`.
this should probably mention the possible staleness of the retrieved informations.

xrootd/xrdproto/open/open.go
```diff
@@ -0,0 +1,101 @@
...
+// Response is a response for the open request,
+// which contains the file handle, the compression page size,
+// the compression type and the stat information.
+type Response struct {
+	FileHandle  xrdfs.FileHandle
+	Compression *xrdfs.FileCompression
+	Stat        *xrdfs.EntryStat
+}
+
+// MarshalXrd implements xrootd/protocol.Marshaler
```
sbinet: `s|xrootd/protocol|xrdproto|`

xrootd/xrdproto/open/open.go
```diff
@@ -0,0 +1,101 @@
...
+	if o.Stat == nil {
+		return nil
+	}
+	if err := o.Stat.MarshalXrd(wBuffer); err != nil {
+		return err
+	}
+	return nil
+}
+
+// UnmarshalXrd implements xrootd/protocol.Unmarshaler
```
sbinet: `s|xrootd/protocol|xrdproto|`

xrootd/xrdproto/open/open.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil
+	}
+	o.Stat = &xrdfs.EntryStat{}
+	if err := o.Stat.UnmarshalXrd(rBuffer); err != nil {
+		return err
+	}
+	return nil
+}
+
+// RespID implements protocol.Response.RespID
```
sbinet: `s/protocol/xrdproto/`

xrootd/xrdproto/open/open.go
```diff
@@ -0,0 +1,101 @@
...
+	_       [12]byte
+	Path    string
+}
+
+// NewRequest forms a Request according to provided path, mode, and options.
+func NewRequest(path string, mode xrdfs.OpenMode, options xrdfs.OpenOptions) *Request {
+	return &Request{Mode: mode, Options: options, Path: path}
+}
+
+// MarshalXrd implements xrootd/protocol.Marshaler
```
sbinet: `s|xrootd/protocol|xrdproto|`

xrootd/xrdproto/open/open.go
```diff
@@ -0,0 +1,101 @@
...
+func (o Request) MarshalXrd(wBuffer *xrdenc.WBuffer) error {
+	wBuffer.WriteU16(uint16(o.Mode))
+	wBuffer.WriteU16(uint16(o.Options))
+	wBuffer.Next(12)
+	wBuffer.WriteStr(o.Path)
+	return nil
+}
+
+// UnmarshalXrd implements xrootd/protocol.Unmarshaler
```
sbinet: `s|xrootd/protocol|xrdproto|`

xrootd/xrdproto/open/open.go
```diff
@@ -0,0 +1,101 @@
...
+func (o *Request) UnmarshalXrd(rBuffer *xrdenc.RBuffer) error {
+	o.Mode = xrdfs.OpenMode(rBuffer.ReadU16())
+	o.Options = xrdfs.OpenOptions(rBuffer.ReadU16())
+	rBuffer.Skip(12)
+	o.Path = rBuffer.ReadStr()
+	return nil
+}
+
+// ReqID implements protocol.Request.ReqID
```
sbinet: `s|protocol|xrdproto|`

xrootd/xrdproto/signing.go
```diff
@@ -96,16 +97,20 @@ func NewSignRequirements(level SecurityLevel, overrides []SecurityOverride) Sign
 
 	if level >= Compatible {
 		// TODO: set requirements
+		sr.requirements[open.RequestID] = SignLikely
```
sbinet: is the `TODO` still needed?

xrootd/xrdfs/file.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// MarshalXrd implements xrootd/protocol.Marshaler
+func (o FileCompression) MarshalXrd(wBuffer *xrdenc.WBuffer) error {
+	wBuffer.WriteI32(o.PageSize)
+	wBuffer.WriteBytes(o.Type[:])
+	return nil
+}
+
+// UnmarshalXrd implements xrootd/protocol.Unmarshaler
```
sbinet: `s|xrootd/protocol|xrdproto|`

xrootd/xrdfs/file.go
```diff
@@ -0,0 +1,39 @@
...
+// FileHandle is the file handle, which should be treated as opaque data.
+type FileHandle [4]byte
+
+// FileCompression holds the compression parameters such as the page size and the type of compression.
+type FileCompression struct {
+	PageSize int32
+	Type     [4]byte
+}
+
+// MarshalXrd implements xrootd/protocol.Marshaler
```
sbinet: `s|xrootd/protocol|xrdproto|`

xrootd/xrdfs/file.go
```diff
@@ -0,0 +1,39 @@
...
+	Info() *EntryStat              // Info returns the stat info.
+	Handle() FileHandle            // Handle returns the file handle.
+}
+
+// FileHandle is the file handle, which should be treated as opaque data.
+type FileHandle [4]byte
+
+// FileCompression holds the compression parameters such as the page size and the type of compression.
+type FileCompression struct {
+	PageSize int32
```
sbinet: it would be interesting to document in what unit (kB, MB, ...) this page size is.

xrootd/xrdfs/file.go
```diff
@@ -0,0 +1,39 @@
...
+	Type     [4]byte
```
sbinet: please document this exported field as well. (I guess it's the name of the compression algorithm?)

xrootd/client/file.go
```diff
@@ -0,0 +1,34 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package client // import "go-hep.org/x/hep/xrootd/client"
+
+import "go-hep.org/x/hep/xrootd/xrdfs"
+
+// File implements access to a content and meta information of file over XRootD.
+type file struct {
+	c           *Client
```
sbinet: instead of `*Client`, wouldn't it make more sense to have `*fileSystem` instead?

xrootd/client/file.go
```diff
@@ -0,0 +1,34 @@
...
+	compression *xrdfs.FileCompression
+	info        *xrdfs.EntryStat
+}
+
+// Compression returns the compression info.
+func (f file) Compression() *xrdfs.FileCompression {
+	return f.compression
+}
+
+// Info returns the stat info.
```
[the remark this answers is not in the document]
EgorMatirov: You are right about the documentation.

Re `kXR_stat` I was thinking about something like:
```go
// FetchStatInfo updates the stat info of this file from the XRootD server.
func (f *file) FetchVirtualStatInfo(ctx context.Context) error {
	var resp stat.DefaultResponse
	err := f.c.Send(ctx, &resp, stat.NewRequestForHandle(f.handle))
	if err != nil {
		return err
	}
	f.info = &resp.EntryStat
	return nil
}
```
which will update this cached info.
This way user can easily re-use cached info and fetch fresh one (I'm not sure about naming though).

xrootd/xrdproto/signing.go
EgorMatirov: Well, we need to add requirements for the most of requests, so I think we can leave `TODO` here to be sure.

xrootd/client/file.go
EgorMatirov: From some point of view, it is so, but I can't come up with the scenario when `fileSystem` will be needed in the `file` yet.

sbinet: ok. well, one can always reach at the underlying `Client` from `*fileSystem` and get a `*fileSystem` from `Client` (although, in that case, at the price of a type-check.)

let's go with `fs *fileSystem` instead?

xrootd/client/file.go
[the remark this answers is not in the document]
sbinet: I think updating the cached `*info` field can be achieved as a (documented) side-effect of calling the yet-to-be-defined `file.Stat() (os.FileInfo, error)` (not sure yet whether we'd make it return `os.FileInfo` or `xrdfs.EntryStat`...)

xrootd/xrdproto/signing.go
sbinet: ok.

sbinet:
(also, could you rebase?)

xrootd/xrdfs/file.go
EgorMatirov: I was assuming that it's in bytes but it's unclear from the documentation (`cpsize` part): http://xrootd.org/doc/dev45/XRdv310.htm#_Toc464248823 :(

Same goes for the `cptype` - it's unclear from the documentation whether it's name or not.

I have tried to call `open` with `kxr_compress` option on `gzip`-ed file some time ago but haven't got any success with that.

sbinet: feel free to send an issue asking for some clarifications of the specs :)

EgorMatirov: I've filled an issue: https://github.com/xrootd/xrootd/issues/724.

sbinet: thanks.

xrootd/client/filesystem.go
```diff
@@ -31,6 +32,16 @@ func (fs *fileSystem) Dirlist(ctx context.Context, path string) ([]xrdfs.EntrySt
...
 }
 
+// Open returns the file handle for a file together with the compression and the stat info.
+func (fs *fileSystem) Open(ctx context.Context, path string, mode xrdfs.OpenMode, options xrdfs.OpenOptions) (xrdfs.File, error) {
+	var resp open.Response
+	err := fs.c.Send(ctx, &resp, open.NewRequest(path, mode, options))
+	if err != nil {
+		return file{}, err
+	}
+	return file{fs, resp.FileHandle, resp.Compression, resp.Stat}, nil
```
sbinet: you probably want to `return &file{...}` here (and `return nil, err` above).
right now, both `client.file` and `*client.file` implement `xrdfs.File` but I suspect that later on, only `*client.file` will be able to implement it (_ie:_  for all the methods that modify the internal state of `client.file`, a pointer-receiver will be required.)

xrootd/xrdfs/fs.go
```diff
@@ -12,4 +12,61 @@ import (
...
+)
+
+// OpenOptions are the options to apply when path is opened.
+type OpenOptions uint16
+
+const (
+	// OpenOptionsNone specifies that file is opened without specific options.
+	OpenOptionsNone OpenOptions = 0
+	// OpenOptionsCompress specifies that file is opened even when compressed.
+	OpenOptionsCompress OpenOptions = 1
```
sbinet: could we replace these with the explicit way of defining a bitmask? (ie: `1<<0`, `1<<1`, ... (or `1<<iota`))

EgorMatirov:
PTAL at https://github.com/go-hep/hep/pull/220 for the failure fix. :-/

PR #221, opened 2018-06-05
xrootd: add stat request

EgorMatirov:
Note that TestFile_FetchVirtualStatInfo is failing with an error: `xrootd: error 3001: Required argument not present`. I think that it's because it is called with `handle` instead of `path` (`FileSystem` calls it with `path` and it works correctly),

Can you take a look at [specs for kXR_stat](http://xrootd.org/doc/dev45/XRdv310.htm#_Toc464248850), please?
Especially:
>Notes
>...
>4)      kXR_stat - kXR_vfs requests need not specify an existing filesystem object. The specified path is used as a path prefix in order to filter out servers and partitions that could not be used to hold objects whose path starts with the specified path prefix.

I think that since there is no mention of `handle` it's not obvious that `kXR_stat` *requires* `path` to be passed and doesn't work when only `handle` is present.

Am I right? I'm a bit in doubt since I'm not very good at English and `requests need not specify an existing filesystem object`  gives a hint that `handle` would not work.

I'll fill an issue if you too think that documentation is not clear on that topic.

### DIFF
--- a/xrootd/xrdproto/stat/stat.go
+++ b/xrootd/xrdproto/stat/stat.go
@@ -1,0 +1,76 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package stat contains the structures describing request and response for stat request.
+// See xrootd protocol specification (http://xrootd.org/doc/dev45/XRdv310.pdf, p. 113) for details.
+package stat // import "go-hep.org/x/hep/xrootd/xrdproto/stat"
+
+import (
+	"go-hep.org/x/hep/xrootd/internal/xrdenc"
+	"go-hep.org/x/hep/xrootd/xrdfs"
+)
+
+// RequestID is the id of the request, it is sent as part of message.
+// See xrootd protocol specification for details: http://xrootd.org/doc/dev45/XRdv310.pdf, 2.3 Client Request Format.
+const RequestID uint16 = 3017
+
+// DefaultResponse is a response for the stat request which contains stat information such as:
+// the OS-dependent identifier, the size of the data, the entry attributes and the modification time.
+type DefaultResponse struct {
+	xrdfs.EntryStat
+}
+
+// VirtualFSResponse is a response for the stat request
+// which contains virtual file system stat information such as:
+// nrw - the number of nodes that can provide read/write access,
+// frw - the size of the largest contiguous area of r/w free space,
+// urw - the percent utilization of the partition represented by frw,
+// nstg - the number of nodes that can provide staging access,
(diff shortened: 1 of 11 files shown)


sbinet:
actually, before implementing `kXR_stat`, could you go with `kXR_close`?
I fear that we may crash the server with opening gazillions of files w/o closing them...

also, this will make other tests cleaner, spelling out the correct semantics like:
```go
f, err := fs.Open(...)
if err != nil { ... }
defer f.Close()
```

EgorMatirov:
Yep, I'll do.

sbinet:
as for the specs, I agree the fact that `path` is always required, even when one passes a `handle` could be better phrased.

xrootd/xrdproto/stat/stat.go
```diff
@@ -0,0 +1,96 @@
...
+	Options    Options
+	_          [11]uint8
+	FileHandle xrdfs.FileHandle
+	Path       string
+}
+
+// NewRequestForPath forms a Request according to provided path
+// and virtualFS parameter which indicates whether virtual file system information
+// should be returned instead of default stat information.
+func NewRequestForPath(path string, virtualFS bool) *Request {
```
sbinet: not sure this function (nor the one below: `NewRequestForHandle`) actually improves the usability.

consider:
```go
req := stat.NewRequestForPath("/foo/bar", false)
req := stat.Request{Path: "/foo/bar"}
```
or:
```go
req := stat.NewRequestForPath("/foo/bar", true)
req := stat.Request{Path: "/foo/bar", Options: stat.OptionsVFS}
```

IMHO, it reads better using the named-struct-fields way :)

I would have probably gone with the `NewXyz` functions for forward compatibility in the event that `stat.Request` gained new `Options` or new fields (recouped from the `[11]byte` reserve), but the specs require that everything is zeroed, so new versions of `Request` would have to differ.

what do you think?

EgorMatirov: > but the specs require that everything is zeroed

I got it this way: in the future, these zeroed values can be used for passing some values that will have zero as their default value.

Let's assume that some of the reserved values is changed to the meaningful parameter. Probably the default value (`0`) will be suitable for the most of cases, so backward compatibility will be kept.

I think that in most of cases we will not want to change these default values (to keep backward compatibility from the our side too). So, `NewRequestForPath` will not be modified and another func will be introduced.

From that point of view it's ok to use `Request{ ... }` instead of `NewRequestForXyz`.

sbinet:
while still waiting on an answer on https://github.com/xrootd/xrootd/issues/728, could you please rebase and skip the failing test (but also adding a `FIXME` note that points to the `xrootd/xrootd#728` issue) ?

thanks.

PR #222, opened 2018-06-06
xrootd: add close request

EgorMatirov:
Updates go-hep/hep#170.

Note that `TestFile_CloseVerify` test is skipped.
From the [specs, p. 41](http://xrootd.org/doc/dev45/XRdv310.pdf):
>fsize the size, in bytes, that the file is to have. The close operation fails and the
file is erased if it is not of the indicated size. An fsize of zero suppresses the
check

As far as I understand, `close operation fails` means that response will contain the error, however, the response is `ok` when I provide wrong `fsize`.

I'm going to fill an issue to clarify this behavior. :)

### DIFF
--- a/xrootd/client/file.go
+++ b/xrootd/client/file.go
@@ -4,7 +4,12 @@
 
 package client // import "go-hep.org/x/hep/xrootd/client"
 
-import "go-hep.org/x/hep/xrootd/xrdfs"
+import (
+	"context"
+
+	"go-hep.org/x/hep/xrootd/xrdfs"
+	"go-hep.org/x/hep/xrootd/xrdproto/xrdclose"
+)
 
 // File implements access to a content and meta information of file over XRootD.
 type file struct {
@@ -28,6 +33,19 @@ func (f file) Info() *xrdfs.EntryStat {
 // Handle returns the file handle.
 func (f file) Handle() xrdfs.FileHandle {
 	return f.handle
+}
+
+// Close closes the file.
+func (f file) Close(ctx context.Context) error {
+	_, err := f.fs.c.call(ctx, &xrdclose.Request{Handle: f.handle})
+	return err
+}
+
+// CloseVerify closes the file and checks whether the file has the provided size.
+// A zero size suppresses the verification.
+func (f file) CloseVerify(ctx context.Context, size int64) error {
+	_, err := f.fs.c.call(ctx, &xrdclose.Request{Handle: f.handle, Size: size})
+	return err
 }
 
 var (
--- a/xrootd/client/file_test.go
+++ b/xrootd/client/file_test.go
@@ -1,0 +1,73 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package client // import "go-hep.org/x/hep/xrootd/client"
+
+import (
+	"context"
+	"testing"
+
+	"go-hep.org/x/hep/xrootd/xrdfs"
+)
+
+func testFile_Close(t *testing.T, addr string) {
+	client, err := NewClient(context.Background(), addr, "gopher")
+	if err != nil {
+		t.Fatalf("could not create client: %v", err)
+	}
+	defer client.Close()
+
+	fs := client.FS()
+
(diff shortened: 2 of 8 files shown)


sbinet:
> I'm going to fill an issue to clarify this behavior. :)

please link back to this PR when you file it :)

xrootd/client/file.go
```diff
@@ -30,6 +35,19 @@ func (f file) Handle() xrdfs.FileHandle {
 	return f.handle
 }
 
+// Close closes the file.
+func (f file) Close(ctx context.Context) error {
+	_, err := f.fs.c.call(ctx, &closereq.Request{Handle: f.handle})
+	return err
+}
+
+// CloseVerify closes the file with verification that the file has provided size.
```
sbinet: `s/with verification that/and checks whether/`
`s/has provided/has the provided/`

xrootd/client/file.go
```diff
@@ -30,6 +35,19 @@ func (f file) Handle() xrdfs.FileHandle {
...
+// An size of 0 suppresses the verification.
```
sbinet: `s/An size of 0/A zero size/`

xrootd/xrdproto/closereq/closereq.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package closereq contains the structures describing request and response for close request.
+// See xrootd protocol specification (http://xrootd.org/doc/dev45/XRdv310.pdf, p. 41) for details.
+package closereq // import "go-hep.org/x/hep/xrootd/xrdproto/closereq"
```
sbinet: `closereq` reads funny when you ask for a `Response` (and stutters somewhat with `Request`.)

naming is hard but here are some possibilities:
- `closexrd`, `xrdclose`
- `closeop`, `opclose`

xrootd/xrdproto/closereq/closereq.go
```diff
@@ -0,0 +1,46 @@
...
+	wBuffer.WriteI64(o.Size)
+	wBuffer.Next(8)
+	return nil
+}
+
+// UnmarshalXrd implements xrdproto.Unmarshaler
+func (o *Request) UnmarshalXrd(rBuffer *xrdenc.RBuffer) error {
+	rBuffer.ReadBytes(o.Handle[:])
+	o.Size = rBuffer.ReadI64()
+	return nil
```
sbinet: we should probably `Skip(8)` there too, right?

xrootd/client/file_test.go
```diff
@@ -0,0 +1,71 @@
...
+	defer client.Close()
+
+	fs := client.FS()
+
+	file, err := fs.Open(context.Background(), "/tmp/dir1/file1.txt", xrdfs.OpenModeOtherRead, xrdfs.OpenOptionsNone)
+	if err != nil {
+		t.Fatalf("invalid open call: %v", err)
+	}
+
+	err = file.Close(context.Background())
```
sbinet: could the other tests that `Open` a file be modified to add a `defer close` call?
so we don't leak file descriptors during those tests either.

xrootd/client/filesystem_test.go
```diff
@@ -62,6 +62,7 @@ func testFileSystem_Open(t *testing.T, addr string, options xrdfs.OpenOptions, w
 	if err != nil {
 		t.Fatalf("invalid open call: %v", err)
 	}
+	defer gotFile.Close(context.Background())
```
sbinet: yep. but I think we are also missing a call to `Close` in https://github.com/EgorMatirov/hep/blob/b01fecaacad6396e62701a8de9eb97c3fbdbb311/xrootd/client/filesystem_mock_test.go#L243

EgorMatirov: That's mocked test. Server exists in the other goroutine and doesn't actually allocate any file handlers. I am not sure that we need to bother with that workflow here (it would probably require to wait for another request from the server side).

Since there is `TestFile_Close_Mock` we still test that we can (un)marshal and parse `Close` from both the server and the client side.

sbinet: ok.

sbinet: perhaps just a little `TODO/FIXME` note there?

sbinet: thanks :)

PR #225, opened 2018-06-06
xrootd: add read request (without read-ahead usage)

EgorMatirov:
Updates go-hep/hep#170.

I would like to implement [io.ReaderAt](https://golang.org/pkg/io/#ReaderAt) (and probably `io.Reader`) interface but I'm not sure about naming of methods that take a `context` as a parameter.

And basically it will be like:
```go
func (f file) ReadAt(p []byte, off int64) (n int, err error) {
        return f.ReadAtWithContext(context.Background(), p, off)
}
```

What do you think? Should we implement such methods? Any suggestions about naming?

### DIFF
--- a/xrootd/client/file.go
+++ b/xrootd/client/file.go
@@ -8,6 +8,7 @@ import (
 	"context"
 
 	"go-hep.org/x/hep/xrootd/xrdfs"
+	"go-hep.org/x/hep/xrootd/xrdproto/read"
 	"go-hep.org/x/hep/xrootd/xrdproto/sync"
 	"go-hep.org/x/hep/xrootd/xrdproto/xrdclose"
 )
@@ -53,6 +54,22 @@ func (f file) CloseVerify(ctx context.Context, size int64) error {
 func (f file) Sync(ctx context.Context) error {
 	_, err := f.fs.c.call(ctx, &sync.Request{Handle: f.handle})
 	return err
+}
+
+// ReadAtContext reads len(p) bytes into p starting at offset off.
+func (f file) ReadAtContext(ctx context.Context, p []byte, off int64) (n int, err error) {
+	var resp read.Response
+	err = f.fs.c.Send(ctx, &resp, &read.Request{Handle: f.handle, Offset: off, Length: int32(len(p))})
+	if err != nil {
+		return 0, err
+	}
+	copy(p, resp.Data)
+	return len(resp.Data), nil
+}
+
+// ReadAt reads len(p) bytes into p starting at offset off.
+func (f file) ReadAt(p []byte, off int64) (n int, err error) {
+	return f.ReadAtContext(context.Background(), p, off)
 }
 
 var (
--- a/xrootd/client/file_test.go
+++ b/xrootd/client/file_test.go
@@ -6,6 +6,7 @@ package client // import "go-hep.org/x/hep/xrootd/client"
 
 import (
 	"context"
+	"reflect"
 	"testing"
 
 	"go-hep.org/x/hep/xrootd/xrdfs"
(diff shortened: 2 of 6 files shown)


sbinet:
For the belong part, we could follow the lead of database/sql:
- https://godoc.org/database/sql#DB

(See all the methods that are named 'Foo' and 'FooContext'.)

Then to implement e.g. 'io.ReaderAt', I would probably investigate having 'file' grow a context field, captured during creation of that 'file' (so, a 'child' of the one from filesystem.)

xrootd/xrdproto/read/read.go
```diff
@@ -0,0 +1,141 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// ReadAhead is the pre-read request. It is considered only a hint
+// and can be used to schedule the pre-reading of data that will be asked
+// int the very near future.
```
sbinet: `s/int/in/`

xrootd/client/file_test.go
```diff
@@ -71,3 +72,42 @@ func TestFile_CloseVerify(t *testing.T) {
...
+func testFile_ReadAt(t *testing.T, addr string) {
+	client, err := NewClient(context.Background(), addr, "gopher")
+	if err != nil {
+		t.Fatalf("could not create client: %v", err)
+	}
+	defer client.Close()
+
+	fs := client.FS()
+
+	file, err := fs.Open(context.Background(), "/tmp/file1.txt", xrdfs.OpenModeOtherRead, xrdfs.OpenOptionsNone)
```
sbinet: perhaps we should add another (somewhat bigger, let's say 8k) file to further test `ReadAt` ?

EgorMatirov: I agree.

We can do it in `write` test which will write a 8k file to the server and then read it to check back.
Otherwise, we can have a separate test for `read` which will read from the precreated file. 

What do you think?

sbinet:
you're right.

I guess I wanted to reap off the benefits of using `context` w/o the cost(s).
it's true that your code snippet may be confusing and that's because, indeed, `ReadAt` makes no mention of `context.Context`, so a not so careful reader/user might overlook that the "lifeline" of `file` will end in 2s, for all its pending operations.

having `ctx` captured from the `fileSystem` would ensure that all the goroutines that may or may not have been spawned are tied back to `fileSystem`.
so, no goroutine leak.

it boils down to what kind of semantics we want for:
```go
n, err := f.ReadAt(data, offset)
```
- life time tied to `f`, or
- life time tied to the whole program.

I am reaching my level of incompetence to decide on what is the best semantics.
but: "when in doubt do it like the Go core team" would be my strategy :)
(so, TL;DR: ok. let's leave it like you wrote it in the first place)

xrootd/client/file_test.go
sbinet: SGTM.

EgorMatirov:
@sbinet PTAL. :)

xrootd/client/file.go
```diff
@@ -55,6 +56,22 @@ func (f file) Sync(ctx context.Context) error {
 	return err
 }
 
+// ReadAtContext reads len(p) bytes into p starting at offset off.
+func (f file) ReadAtContext(ctx context.Context, p []byte, off int64) (n int, err error) {
+	var resp read.Response
```
sbinet: actually, couldn't we directly create the response pointing at the destination `p` ?
something like:
```go
resp := read.Response{Data:p}
```

xrootd/client/file.go
```diff
@@ -55,6 +56,22 @@ func (f file) Sync(ctx context.Context) error {
...
+	err = f.fs.c.Send(ctx, &resp, &read.Request{Handle: f.handle, Offset: off, Length: int32(len(p))})
+	if err != nil {
+		return 0, err
+	}
+	copy(p, resp.Data)
```
sbinet: then, we can drop this extra copy.

xrootd/client/file.go
sbinet: actually, it seems the `copy` is necessary (at least, the way we currently implement things).
at the very least, it should be:
```go
resp := read.Response{Data: make([]byte, 0, len(p))}
```

EgorMatirov:
Thanks! Note that it's also possible to use
```go
data, err := f.fs.c.call(ctx, &read.Request{Handle: f.handle, Offset: off, Length: int32(len(p))})
```

sbinet:
yes, but this would still involve a copying a temporary buffer into the final location.

see: go-hep/hep#233

PR #231, opened 2018-06-07
xrootd: add rm request

EgorMatirov:
Updates go-hep/hep#170.


xrootd/xrdfs/fs.go
```diff
@@ -13,6 +13,7 @@ import (
 type FileSystem interface {
 	Dirlist(ctx context.Context, path string) ([]EntryStat, error)
 	Open(ctx context.Context, path string, mode OpenMode, options OpenOptions) (File, error)
+	RemoveFile(ctx context.Context, path string) error
```
sbinet: that's unfortunate... the Go equivalent is called `os.Remove` and works for both (empty) dir and file...
we could of course emulate it but this would involve one network extra roundtrip (for `Stat`, to decide whether this is a file or a dir...)

oh well...

sbinet:
fixed w/ 7f037c8

PR #233, opened 2018-06-07
xrootd/{client,xrdproto/read}: make ReadAt directly use provided buffer

sbinet:
This CL modifies the read.Response unmarshaler to use the provided
buffer if it's big enough.
Otherwise, a new big enough buffer is allocated.


EgorMatirov:
Looks like parallel CI breaks writing test. :(

sbinet:
ah! interesting... and not completely surprising either :)

we should probably root all the tests that perform some modification of the state of the remote dirs+files under some `/tmp/test-xxxxx` where `test-xxxxx` has been obtained with `io/ioutil.TempDir`...
(this means we need `kXR_mkdir` :))

PR #234, opened 2018-06-08
Benchmarking : BenchmarkChainTreeScannerStruct

maloft:


rootio/scanner.go
```diff
@@ -54,6 +54,11 @@ func (s *baseScanner) SeekEntry(i int64) error {
 	if s.err != nil {
 		return s.err
 	}
+	if s.chain {
+		if s.cur >= s.n {
+			s.nextTree(0)
```
sbinet: this works in the bench because we do `SeekEntry(0)`, which by construction is indeed the tree `0`.

but that wouldn't work if we had 2 trees (like we do) and issued `SeekEntry(12)` (which should land inside `chain.2.root`.)

so, this:
```go
s.nextTree(0)
```
should be replaced with:
```go
itree := s.findTree(i) // find the tree number in the chain that contains entry i
s.loadTree(itree)
```

rootio/scanner.go
```diff
@@ -72,16 +77,17 @@ func (s *baseScanner) Next() bool {
...
+			ch := s.tree.(*tchain)
+			s.nextTree(ch.cur + 1)
 		}
 	}
 
 	return next
 }
 
-func (s *baseScanner) nextTree() {
+func (s *baseScanner) nextTree(i int) {
```
sbinet: this method should probably be renamed `loadTree`.
it made sense to name it `nextTree` when it wasn't taking any argument, but now this is a bit confusing:
```go
scan.nextTree(3)
```
does this mean we go to the next 3 trees from the current position or the 3rd tree (_ie:_ absolute position) ?
:)

IMHO, `loadTree(i int)` reads better.

rootio/tchain.go
```diff
@@ -40,12 +40,12 @@ func Chain(trees ...Tree) Tree {
...
 		off += n
 	}
-	ch.nextTree()
+	ch.nextTree(ch.cur + 1)
 	return ch
 }
 
-func (ch *tchain) nextTree() {
-	ch.cur++
+func (ch *tchain) nextTree(i int) {
```
sbinet: under the same argument, I would rename this method to `loadTree`.

rootio/tchain_test.go
```diff
@@ -86,6 +86,8 @@ func TestChainScanStruct(t *testing.T) {
 	files := []string{
 		"testdata/chain.1.root",
 		"testdata/chain.2.root",
+		//"testdata/chain.2.root",
```
sbinet: please remove these 2 commented out lines.

rootio/tchain_test.go
```diff
@@ -181,3 +183,70 @@ func TestChainScanStruct(t *testing.T) {
...
 	}
 }
+
+var SumF64 float64 // global variable to prevent unwanted compiler optimization
+
+func BenchmarkChainTreeScannerStruct(b *testing.B) {
+	files := []string{
+		"testdata/chain.1.root",
+		"testdata/chain.2.root",
+		//"testdata/chain.1.root",
```
sbinet: please remove these 2 lines.

rootio/scanner.go
```diff
@@ -54,11 +54,36 @@ func (s *baseScanner) SeekEntry(i int64) error {
...
+	for j <= len(ch.trees) {
+		if i <= ch.tree.Entries() {
+			break
+		}
+		j++
+	}
+	return j
+}
+
+/*
```
EgorMatirov: Sorry for the interference, but could you please remove these commented lines? :)

rootio/scanner.go
```diff
@@ -54,11 +54,36 @@ func (s *baseScanner) SeekEntry(i int64) error {
...
+			itree := s.findTree(i)
+			s.loadTree(itree)
+		}
+	}
 	s.i = i
 	s.cur = i - 1
 	return s.err
 }
 
+// find the tree number in the chain that contains entry i
```
sbinet: the usual doc line would be more like:
```go
// findTree finds the tree ...
```

also: could you document what happens when we don't find the tree?
(w/o looking at the method, I would expect the returned value to be, say, `-1`.)

rootio/scanner.go
```diff
@@ -54,11 +54,36 @@ func (s *baseScanner) SeekEntry(i int64) error {
...
+func (s *baseScanner) findTree(i int64) int {
+	ch := s.tree.(*tchain)
+	var j int = 0
+	for j <= len(ch.trees) {
```
sbinet: the way it's implemented as of now, `findTree` will return an invalid index (ie `len(ch.trees)`) when the requested entry `i` is outside of the chain.
fine, but I think `-1` would be better and less surprising considering a few of the Go APIs.
_e.g.:_
```
$> go doc strings.Index
func Index(s, substr string) int
    Index returns the index of the first instance of substr in s, or -1 if
    substr is not present in s.
```

rootio/scanner.go
```diff
@@ -54,11 +54,36 @@ func (s *baseScanner) SeekEntry(i int64) error {
...
+	}
 	s.i = i
 	s.cur = i - 1
 	return s.err
 }
 
+// find the tree number in the chain that contains entry i
+func (s *baseScanner) findTree(i int64) int {
+	ch := s.tree.(*tchain)
+	var j int = 0
```
sbinet: `s/var j int = 0/j := 0/`
(as clear, yet shorter to write and grok)

rootio/scanner.go
```diff
@@ -54,11 +54,36 @@ func (s *baseScanner) SeekEntry(i int64) error {
...
+	for j <= len(ch.trees) {
+		if i <= ch.tree.Entries() {
+			break
```
sbinet: `s/break/return j/`

rootio/scanner.go
```diff
@@ -54,11 +54,36 @@ func (s *baseScanner) SeekEntry(i int64) error {
...
 	s.cur = i - 1
 	return s.err
 }
 
+// find the tree number in the chain that contains entry i
+func (s *baseScanner) findTree(i int64) int {
+	ch := s.tree.(*tchain)
+	var j int = 0
+	for j <= len(ch.trees) {
+		if i <= ch.tree.Entries() {
```
sbinet: I don't think this works.
`ch.tree` is always pointing at the same tree, so `ch.tree.Entries()` will always return the same value.

I believe the correct code would be:
```go
for j, tree := range ch.trees {
    if i < tree.Entries() {
         return j
    }
}
return -1
```

rootio/scanner.go
```diff
@@ -54,11 +54,36 @@ func (s *baseScanner) SeekEntry(i int64) error {
 	if s.err != nil {
 		return s.err
 	}
+	if s.chain {
+		if s.cur >= s.n {
+			itree := s.findTree(i)
```
sbinet: we should probably add a test here that we actually did find the tree.

rootio/scanner.go
```diff
@@ -57,25 +57,25 @@ func (s *baseScanner) SeekEntry(i int64) error {
 	if s.chain {
 		if s.cur >= s.n {
 			itree := s.findTree(i)
-			s.loadTree(itree)
+			if itree != -1 {
```
sbinet: I would actually reverse the test. we do want to propagate back to the user that we couldn't find the entry she was desperately needing to get her Nobel prize.
```go
if itree == -1 {
    s.err = fmt.Errorf("rootio: could not find Tree for entry %d", i)
    return s.err
}
s.loadTree(itree)
```

rootio/tchain_test.go
```diff
@@ -246,3 +246,55 @@ func BenchmarkChainTreeScannerStruct(b *testing.B) {
...
+
+	sc, err := rootio.NewTreeScanner(chain, &Data{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer sc.Close()
+
+	var entries = []int64{1, 0, 2, 9, 10, 11, 0, 19}
+	for _, entry := range entries {
+		err := sc.SeekEntry(entry)
```
sbinet: ok, but we should also test that we landed on the correct entry.
_e.g._: we know that the field `F64` should be equal to `float64(entry)`.

rootio/tchain_test.go
```diff
@@ -292,9 +292,20 @@ func TestSeekEntry(t *testing.T) {
 
 	var entries = []int64{1, 0, 2, 9, 10, 11, 0, 19}
 	for _, entry := range entries {
-		err := sc.SeekEntry(entry)
+		err = sc.SeekEntry(entry)
 		if err != nil {
-			t.Error("Could not seek entry %d", entry)
+			t.Fatal(err)
 		}
+		for sc.Next() {
```
sbinet: this will loop over the whole remaining entries in the file.
you do have to call `Next()` to load the data associated with the current entry, though.
so:
```go
if !sc.Next() {
    // handle error
    ...
}
```

maloft: i keep having different values when i compare float64(entry) and F64 field,
please, is there something else that needs a correction ?

sbinet: what are those?

maloft: It kinds of work, but : 
If for example the slice of entries is {1, 2, 14, 13}, ie : the transition from one tree to another is not via the first element of that tree, then it doesn't work. (this will work as an example : {2, 4, 7, 10, 15, 0}

maloft: kind of works*

sbinet:
hum... could you try on top of this commit: c47f95a3ea3481748ff4cc9bc85c088630e406cb ?

rootio/tchain_test.go
```diff
@@ -181,3 +181,142 @@ func TestChainScanStruct(t *testing.T) {
...
+		} `rootio:"evt"`
+	}
+
+	sc, err := rootio.NewTreeScanner(chain, &Data{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer sc.Close()
+
+	var entries = []int64{1, 9, 2, 5, 11, 15, 2, 14, 3}
```
sbinet: as said offline, please add seeking to `0` after seeking to `3` to exercize this interesting boundary condition.

PR #238, opened 2018-06-08
Rootio xrd stat support

sbinet:
This CL adds:

- an `os.File.Name`-like method to `xrdfs.File`
- the according implementation for `xrootd/client.file`
- proper computation of the `rootio.xrdFile` size, using `kXR_stat`.

### DIFF
--- a/rootio/fileplugin.go
+++ b/rootio/fileplugin.go
@@ -101,7 +101,7 @@ func xrdOpen(name string) (*xrdFile, error) {
 		return nil, errors.Errorf("rootio: could not open %q: %v", name, err)
 	}
 
-	return &xrdFile{cli: xrd, fs: fs, f: f}, nil
+	return &xrdFile{cli: xrd, fs: fs, f: f, name: name}, nil
 }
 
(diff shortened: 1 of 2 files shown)


xrootd/client/file.go
```diff
@@ -24,6 +24,14 @@ type file struct {
 	info        *xrdfs.EntryStat
 }
 
+// Name returns the name of the file.
+func (f file) Name() string {
+	if f.info == nil {
+		return ""
+	}
+	return f.info.Name()
```
EgorMatirov: XRootD doesn't return name as part of stat info for the `open` and `stat` requests: http://xrootd.org/doc/dev45/XRdv310.htm#_Toc464248823
We can only guess it from the path provided to the `fileSystem.Open` unfortunately.

sbinet: ah! you know the specs better than I do, now :)
you're officially the expert :P 

ok. I'll drop the xrootd modification and will just cache the path inside `rootio.xrdFile` for now.

thanks!

PR #241, opened 2018-06-10
xrootd: fix virtual stat test

EgorMatirov:
Use ranges instead of exact values to prevent failing due to changes in files.


sbinet:
build failure is unrelated (`lz4` seems to have applied some deep refactoring...)

PR #248, opened 2018-06-11
Xrootd fs removeall

sbinet:


xrootd/client/filesystem.go
```diff
@@ -102,6 +107,33 @@ func (fs *fileSystem) RemoveDir(ctx context.Context, path string) error {
...
+// It removes everything it can but returns the first error it encounters.
+// If the path does not exist, RemoveAll returns nil (no error.)
+func (fs *fileSystem) RemoveAll(ctx context.Context, path string) error {
+	st, err := fs.Stat(ctx, path)
+	if err != nil {
+		return err
+	}
+	switch {
+	case st.IsDir():
+		dirs, err := fs.Dirlist(ctx, path)
```
EgorMatirov: In fact it's not only `dirs` but files as well. Maybe rename it to `entries` or something like that? (Probably `Dirlist` is not very exact naming...)

sbinet: good point about `dirs`.

Issue #250, opened 2018-06-11
xrootd: authentication


sbinet:
The `xrootd` [specs]( http://xrootd.org/doc/dev45/XRdv310.htm) have a number of things to say about authentication:

- [kXR_auth](http://xrootd.org/doc/dev45/XRdv310.htm#_Toc464248810)
- [kXR_authmore](http://xrootd.org/doc/dev45/XRdv310.htm#_Toc464248800)
- [Authentication & Access Control Configuration Reference](http://xrootd.org/doc/dev47/sec_config.htm#_Toc489606563)

`xrdsec` support 6 authentication protocols:

- `host`: authenticates a user by originating host name only,
- `gsi`: authenticates a user using GSI protocol,
- `krb5`: authenticates a user using Kerberos V protocol, and
- `pwd`: authenticates a user using a password-based protocol
- `sss`: authenticates a user using a simple shared secret protocol
- `unix`: authenticates using the Unix login name and group name

For kerberos, we might use:
- https://github.com/jcmturner/gokrb5

For GSI, something on top of [crypto/x509](https://godoc.org/crypto/x509) *might* be used/developed.

EgorMatirov:
According to [gsi-msg-specs, p.3](https://github.com/go-hep/hep/files/2089932/gsi-msg-specs.pdf): `the context establishment phase is nothing more than normal SSLv3 handshake messages being exchanged`.
I was using [RFC6101, p. 25](https://tools.ietf.org/html/rfc6101#page-25) as a specification of SSLv3 handshake messages.
I have got TCP dump of xrootd authentication process ([gist](https://gist.github.com/EgorMatirov/8df311ad7adfba7556176d4adfbc8434)):
- the first message is a client `auth` request,
- the second message is a server `auth_more` response,
- the third message is `auth` request again, following with `ok` response (not included in the dump, since it contains only `ok` status).

However, the problem is that the payload of `auth` request doesn't look like SSLv3 `ClientHello` to me.

Also, the server response, which should be `ServerHello` following by `ServerCertificate` should contain ASN.1 encoded cert (you can see an example and decoding [here](http://www.lapo.it/asn1js/)).
But what is present looks more like the plain content of the server's `crt` file. The content between `-----BEGIN CERTIFICATE-----` and `-----END CERTIFICATE-----`  matches the certificate body.

Any thoughts?
Can you suggest me where to look for additional info?

sbinet:
[...]

PR #251, opened 2018-06-11
xrootd: add auth request together with unix auth

EgorMatirov:
Updates go-hep/hep#170.
Updates go-hep/hep#250.

I have tested it locally against the XRootD server with such config:

```
ofs.trace all
xrd.trace all
#sec.level all pedantic force
xrootd.seclib /usr/lib/libXrdSec.so
sec.protocol /usr/lib/ unix
sec.protocol /usr/lib/ pwd
sec.protbind * unix pwd
acc.authdb /tmp/Authfile
ofs.authorize 1
```

/tmp/Authfile is:
```
u ematirov /tmp a
u DESKTOP-R8CLF5C\ematirov /tmp a
```

Note that even Windows is supported by **unix** auth. :-P

### DIFF
--- a/xrootd/client/auth.go
+++ b/xrootd/client/auth.go
@@ -1,0 +1,46 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package client // import "go-hep.org/x/hep/xrootd/client"
+
+import (
+	"bytes"
+	"context"
+	"os/user"
+
+	"github.com/pkg/errors"
+	"go-hep.org/x/hep/xrootd/xrdproto/auth"
+)
+
+func (client *Client) auth(ctx context.Context, securityInformation []byte) error {
+	securityInformation = bytes.TrimLeft(securityInformation, "&")
+	securityProviders := bytes.Split(securityInformation, []byte{'&'})
+
+	var errs []error
+	for _, securityProvider := range securityProviders {
+		securityProvider = bytes.TrimLeft(securityProvider, "P=")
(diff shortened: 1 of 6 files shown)


xrootd/client/auth.go
```diff
@@ -0,0 +1,49 @@
...
+		securityProvider = bytes.TrimLeft(securityProvider, "P=")
+		if bytes.Equal(securityProvider, auth.UnixType[:]) {
+			u, err := user.Current()
+			if err != nil {
+				errs = append(errs, errors.WithMessage(err, "xrootd: could not authorize using unix"))
+				continue
+			}
+			var group string
+			g, err := user.LookupGroupId(u.Gid)
+			if err != nil {
```
sbinet: rather than this relatively _ad hoc_ handling, perhaps we could create a little helper function that forwards to `os/user.LookupGroupId` when on unix and do what you did below when on windows.

that could perhaps look something like (for `lookupgid_unix.go`):
```go
//+build !windows

func lookupGroupID(usr *user.User) (string, error) { ... }
```
and (for `lookupgid_windows.go`):
```go
//+build windows

func lookupGroupID(usr *user.User) (string, error) { ... }
```

xrootd/client/lookupgid_windows.go
```diff
@@ -0,0 +1,15 @@
...
+//+build windows
+
+package client // import "go-hep.org/x/hep/xrootd/client"
+
+import (
+	"os/user"
+)
+
+func lookupGroupID(usr *user.User) (string, error) {
+	return usr.Username, nil
```
sbinet: perhaps add some comment here?
because for somebody not super familiar with Windows (like me), it looks like a bug or a typo :)

PR #253, opened 2018-06-12
Rootio test seekentry

sbinet:


sbinet:
@maloft see how #234 should work in the case of regular `Tree`s ? :)

PR #259, opened 2018-06-15
xrootd: add krb5 auth

EgorMatirov:
@sbinet, could you please give this a try?

I have managed to install and configure Kerberos and that PR works locally with my config. However, yours may differ and it would be cool to test that before polishing it. :)

For the tests, edit the `xrootd/client/auth.go:26` and following lines. That enables Kerberos auth with the provided parameters to all usages of the client, so you can test it against `xrd-cp` for example. :)

I think that we'll receive in the `Client` configured and logged-in client to the Kerberos, so it will allow a user to fully customize the creation of the client.

Another possibility is to introduce `xrootd/auth/krb5` and depend on `gokrb5` only from that package, requiring only basic interface from the `Client` side. Something like:
```go
type KerberosAuth interface {
    Authenticate(serviceName string) ([]byte, error)
}
```

This way it will be optional to install the `xrootd/auth/krb5` and `gokrb5` package since the `Client` will not depend on them.

What do you think?

Updates go-hep/hep#170.
Updates go-hep/hep#250.

### DIFF
--- a/xrootd/client/client.go
+++ b/xrootd/client/client.go
@@ -31,6 +31,9 @@ import (
 	"go-hep.org/x/hep/xrootd/internal/mux"
 	"go-hep.org/x/hep/xrootd/internal/xrdenc"
 	"go-hep.org/x/hep/xrootd/xrdproto"
+	"go-hep.org/x/hep/xrootd/xrdproto/auth"
+	"go-hep.org/x/hep/xrootd/xrdproto/auth/krb5"
+	"go-hep.org/x/hep/xrootd/xrdproto/auth/unix"
 	"go-hep.org/x/hep/xrootd/xrdproto/sigver"
 )
 
@@ -44,12 +47,41 @@ type Client struct {
 	protocolVersion  int32
 	signRequirements xrdproto.SignRequirements
 	seqID            int64
+	auths            map[string]auth.Auther
+}
+
+// Option configures an XRootD client.
+type Option func(*Client) error
+
+// WithAuth adds an authentication mechanism to the XRootD client.
+// If an authentication mechanism was already registered for that provider,
+// it will be silently replaced.
+func WithAuth(a auth.Auther) Option {
+	return func(client *Client) error {
+		return client.addAuth(a)
+	}
+}
+
+func (client *Client) addAuth(auth auth.Auther) error {
+	client.auths[auth.Provider()] = auth
+	return nil
+}
+
+func (client *Client) initSecurityProviders() {
+	providers := []auth.Auther{krb5.Default, unix.Default}
+	for _, provider := range providers {
+		if provider == nil {
+			continue
+		}
+		client.auths[provider.Provider()] = provider
+	}
 }
 
 // NewClient creates a new xrootd client that connects to the given address using username.
+// Options opts configure the client and are applied in the order they were specified.
 // When the context expires, a response handling is stopped, however, it is
 // necessary to call Cancel to correctly free resources.
-func NewClient(ctx context.Context, address string, username string) (*Client, error) {
+func NewClient(ctx context.Context, address string, username string, opts ...Option) (*Client, error) {
 	ctx, cancel := context.WithCancel(ctx)
 
 	var d net.Dialer
@@ -59,7 +91,18 @@ func NewClient(ctx context.Context, address string, username string) (*Client, e
 		return nil, err
 	}
 
-	client := &Client{cancel: cancel, conn: conn, mux: mux.New()}
+	client := &Client{cancel: cancel, conn: conn, mux: mux.New(), auths: make(map[string]auth.Auther)}
+	client.initSecurityProviders()
+
+	for _, opt := range opts {
+		if opt == nil {
+			continue
+		}
+		if err := opt(client); err != nil {
+			client.Close()
+			return nil, err
+		}
+	}
 
 	go client.consume(ctx)
 
--- a/xrootd/xrdproto/auth/krb5/krb5.go
+++ b/xrootd/xrdproto/auth/krb5/krb5.go
@@ -1,0 +1,135 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package krb5 contains the implementation of krb5 (Kerberos) security provider.
+package krb5
+
+import (
+	"strings"
+
+	"github.com/pkg/errors"
+	"go-hep.org/x/hep/xrootd/xrdproto/auth"
+	"gopkg.in/jcmturner/gokrb5.v5/client"
+	"gopkg.in/jcmturner/gokrb5.v5/config"
+	"gopkg.in/jcmturner/gokrb5.v5/credentials"
+	"gopkg.in/jcmturner/gokrb5.v5/crypto"
+	"gopkg.in/jcmturner/gokrb5.v5/messages"
+	"gopkg.in/jcmturner/gokrb5.v5/types"
+)
+
+// Default is a Kerberos 5 client configured from cached credentials.
+// If the credentials could not be correctly configured, Default will be nil.
+var Default auth.Auther
+
+func init() {
+	v, err := WithCredCache()
+	if err == nil {
+		Default = v
+	}
+}
+
+// Auth implements krb5 (Kerberos) security provider.
+type Auth struct {
+	client *client.Client
+}
+
+// WithPassword creates a new Auth configured from the provided user, realm and password.
+func WithPassword(user, realm, password string) (*Auth, error) {
+	krb := client.NewClientWithPassword(user, realm, password)
+
+	cfg, err := config.Load(configPath)
+	if err != nil {
+		return nil, errors.WithMessage(err, "auth/krb5: could not load kerberos-5 configuration")
+	}
+
+	krb.WithConfig(cfg)
+
+	err = krb.Login()
+	if err != nil {
+		return nil, errors.WithMessage(err, "auth/krb5: could not login")
+	}
+
+	return &Auth{client: &krb}, nil
+}
+
+// WithCredCache creates a new Auth configured from cached credentials.
+func WithCredCache() (*Auth, error) {
+	cred, err := credentials.LoadCCache(cachePath())
+	if err != nil {
+		return nil, errors.WithMessage(err, "auth/krb5: could not load kerberos-5 cached credentials")
+	}
+
+	krb, err := client.NewClientFromCCache(cred)
+	if err != nil {
+		return nil, errors.WithMessage(err, "auth/krb5: could not create kerberos-5 client from cached credentials")
+	}
+
+	cfg, err := config.Load(configPath)
+	if err != nil {
+		return nil, errors.WithMessage(err, "auth/krb5: could not load kerberos-5 configuration")
+	}
+
+	krb.WithConfig(cfg)
+
+	return &Auth{client: &krb}, nil
+}
+
+// WithClient creates a new Auth using the provided krb5 client.
+func WithClient(client *client.Client) *Auth {
+	return &Auth{client: client}
+
+}
+
+// Provider implements auth.Auther
+func (*Auth) Provider() string {
+	return "krb5"
+}
+
+// Type indicates that krb5 (Kerberos) authentication protocol is used.
+var Type = [4]byte{'k', 'r', 'b', '5'}
+
+// Request implements auth.Auther
+func (a *Auth) Request(params []string) (*auth.Request, error) {
+	if len(params) == 0 {
+		return nil, errors.New("auth/krb5: want at least 1 parameter, got 0")
+	}
+	serviceName := string(params[0])
+	if strings.Contains(serviceName, "@") {
+		// Service name from the XRootD server may be in the following format: "xrootd/server.example.com@example.com"
+		// While gokrb5 expects server name in that format: "xrootd/server.example.com".
+		// The "@example.com" part (realm) will be guessed from the instance name "server.example.com".
+		index := strings.Index(serviceName, "@")
+		serviceName = serviceName[:index]
+	}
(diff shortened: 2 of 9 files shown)


sbinet:
apologies for the belated answer.

I've played a bit with it.
I managed to log on the public opendata server and almost started to copy some random (small) file:

```
$> xrd-cp  root://eospublic.cern.ch:1094//eos/opendata/atlas/OutreachDatasets/2016-07-29/MC/mc_173045.DYtautauM08to15.root .
[...]
xrd-cp: could not open remote file "root://eospublic.cern.ch:1094//eos/opendata/atlas/OutreachDatasets/2016-07-29/MC/mc_173045.DYtautauM08to15.root": xrd-cp: could not open "root://eospublic.cern.ch:1094//eos/opendata/atlas/OutreachDatasets/2016-07-29/MC/mc_173045.DYtautauM08to15.root": xrootd: statinfo "0q35853.cern.ch?&cap.sym=q+CUHEBX8oAEqtPu2nHLfSuxInA=&cap.msg=QHtGNWO287qHIYzxGIORRghAXLvb9Zzn/EDiQOUGKMBkj1qLPS71NA7maHvsIvfDWSwYTzoo4bdXuIFpHzbcEDGn52XRBVwcWQxvNxhBDQRgRM0DY0pRk8RshaQODX76QzGyTpTRukuoniXnFmzViIaMLK03/pAKl+GPi+5GFu571fA2V3X1kc4B3ZS9tL4i86Ra5H+fzIolwIes3nLIhVG9xXrAnwNyYtBaQucaHzl4e4PeiLkelTeS7XiD4GPjJqPrld1rZnskcVsZgWD3z/VwSicIKEHHBsm7QvQ2ivbtRbFFVltHZddMsXC2DKSLrxF4YaBEJ+DTPtZUZhruet6nYiSAZ8YCD3EjwOFWVLpDpro6G+BEVCaYDv8wCZxpGrdUQyTEhy/sU0oPYsg++TXS+nFPfhCmvHCjhRRlyA7K5xbtXZsxii9D92IWDkmcJ2rdxpv5enbDfOmDjhHCdSt89gpZPs3vEE6GjzP9I2QdY8KtPq68HlI/YOwjZxsJ45ydoRr9vY+aAJCwYcR7mSxnG5OK0VMzkMVEtCRwOiSZ3R/U2AbiIxQZD7Haj/BpHTcBqqZWv1Tc+nSGp8Pbo/3SOZ2TPOxmvZfahzy42yJc0VHBoqjESm3difOD51a5K6K7dFJ7MIXIK5AImWwQLL1/9jo2q7Xf&mgm.logid=8682aa44-721a-11e8-bc80-0cc47aeaf56f&mgm.blockchecksum=ignore&mgm.replicaindex=1&mgm.replicahead=1&mgm.id=0adf7a19" doesn't have enough fields, expected format is: "id size flags modtime"
```

not sure whether this weird thing is coming from EOS or from the xrootd server...

anyways, that's a good first step :)

that said, I am not completely sure directly/explicitly adding the `krb5` client to the `xrootd/client.NewClient` API is something we want to do.

I think I would be more inclined to hide it completely and resort to some plugin mechanism (something like [database/sql](https://godoc.org/database/sql) perhaps ?) to register and use "extra" authentication mechanisms.
(although, for something as widely used as `krb5` and `gsi`, we'd probably automatically pre-register the associated plugins)

the registry of authentication plugins would probably live inside `xrootd/xrdproto/auth`.
then, the `client.Client.auth` method would read something like:
```go
for _, securityProvider := range securityProviders {
    securityProvider = bytes.TrimLeft(securityProvider, "P=")
    secMgr, ok := auth.Provider(securityProvider)
    if !ok { continue }
    req, err := secMgr.Request()
    if err != nil { ... }
    _, err = client.call(ctx, req)
    if err != nil { ... }
    return nil
}
```

WDYT?

EgorMatirov:
> the registry of authentication plugins would probably live inside xrootd/xrdproto/auth.
then, the client.Client.auth method would read something like:

Do you suggest to store the registry of authentication plugins globally like in the `sql`? I see one scenario where that may become not very useful to the user:
 -  User want's to connect to A xrootd server using specific Kerberos credentials.
 -  User want's to connect to B xrootd server at the same time using other Kerberos credentials.

If authentication plugins are configured per each `Client` instance, then it's easily achievable, however.

What do you think?

sbinet:
> We can't do it totally automatically because we need to have some user-provided configs and credentials.

*theoretically* this can be addressed (at least for kerberos) with cached credentials:

```
$> kinit binet@CERN.CH
$> klist
Ticket cache: FILE:/tmp/krb5cc_1000
Default principal: binet@CERN.CH

Valid starting       Expires              Service principal
06/17/2018 14:10:19  06/18/2018 15:10:14  krbtgt/CERN.CH@CERN.CH
	renew until 06/22/2018 14:10:14
```

I haven't managed to successfully use the cached credentials file with `krb5`, though...
that said, we could also prompt the user for his `user/pass` credentials.

> Do you suggest to store the registry of authentication plugins globally like in the sql? I see one scenario where that may become not very useful to the user:
>
> - User want's to connect to A xrootd server using specific Kerberos credentials.
> - User want's to connect to B xrootd server at the same time using other Kerberos credentials.

this is an interesting use case.
for this, my proposal doesn't fit.

sbinet:
so... a middle ground solution:
- keep the plugin mechanism (perhaps adapt it?)
- provide a set of sensible default baked-in plugins
- provide an optional `WithAuth(...)` function that can be passed to `NewClient` for when we know what kind of security provider we want.

e.g.:
```go
type func(*Client) error Option

func WithAuth(a Auther) Option {
    return func(c *Client) error {
        return c.setAuth(a)
    }
}

func NewClient(ctx context.Context, addr string, user string, opts ...Option) (*Client, error) {
    ...
}

type Auther interface {
    Provider() string
    Request() (xrdproto.Request, error)
}
```

so users could do:

```go
cli, err := NewClient(ctx, addr, usr, WithAuth(myauth.Login(usr, passwd)))
```

while `xrd-cp` and `root-ls` (as they need to be "generic") would do:
```go
cli, err := NewClient(ctx, addr, usr)
```

sbinet:
I have filed https://github.com/jcmturner/gokrb5/issues/157 for the `ccache` issue.

sbinet:
looks better :)

here is what I had in mind: https://github.com/go-hep/hep/compare/xrootd-krb5-auth

I actually prefer the way you've defined the `Auther` interface (not coupled to `xrdproto.Request` but directly to `auth.Request`), as this doesn't introduce an import cycle (that's why I had to introduce `authplugin`...)

EgorMatirov:
@sbinet, I have updated on top of your branch.
- Under Windows, using `WithPassword` while krb5 config is present at `c:\winnt\krb5.ini` works fine (but I wasn't able to find up-to-date kerberos client and get caching working).
- Under Arch Linux, krb5 works out-of-the-box after `kinit`.

xrootd/client/client.go
```diff
@@ -44,12 +47,39 @@ type Client struct {
...
+
+func (client *Client) setAuth(auth auth.Auther) error {
+	client.auths[auth.Provider()] = auth
+	return nil
+}
+
+func (client *Client) initSecurityProviders() {
+	providers := []auth.Auther{krb5.Default, unix.Default}
+	for _, provider := range providers {
+		if provider != nil {
```
sbinet: could you invert the condition and `continue`?

xrootd/client/client.go
```diff
@@ -44,12 +47,39 @@ type Client struct {
...
+			client.auths[provider.Provider()] = provider
+		}
+	}
 }
 
 // NewClient creates a new xrootd client that connects to the given address using username.
 // When the context expires, a response handling is stopped, however, it is
 // necessary to call Cancel to correctly free resources.
-func NewClient(ctx context.Context, address string, username string) (*Client, error) {
+func NewClient(ctx context.Context, address string, username string, opts ...Option) (*Client, error) {
```
sbinet: please add a doc comment about this new optional `opts` argument. (probably before the blurb about the `context.Context` expiration.)

xrootd/client/client.go
```diff
@@ -44,12 +47,39 @@ type Client struct {
...
+// WithAuth adds an authentication mechanism to the XRootD client.
+// If an authentication mechanism was already registered for that provider,
+// it will be silently replaced.
+func WithAuth(a auth.Auther) Option {
+	return func(client *Client) error {
+		return client.setAuth(a)
+	}
+}
+
+func (client *Client) setAuth(auth auth.Auther) error {
```
sbinet: the way it's done now, I would rather rename `setAuth` as `addAuth`.

that said, I am not completely sure this warrants its own method, but ok. (I can live with it.) :P

xrootd/xrdproto/auth/krb5/krb5.go
```diff
@@ -0,0 +1,131 @@
...
+
+func init() {
+	v, err := WithCredCache()
+	if err == nil {
+		Default = v
+	}
+}
+
+// KerberosAuth implements krb5 (Kerberos) security provider.
+type KerberosAuth struct {
```
sbinet: looking at it after a few days, and considering the hosting package is named `krb5`, I'd be leaning towards naming this type `Auth`. _ie:_ `krb5.Auth`.

also, perhaps add somewhere:
```go
var (
    _ auth.Auther = (*Auth)(nil)
)
```

to make sure (and document) that `krb5.Auth` implements `auth.Auther`.

xrootd/xrdproto/auth/krb5/krb5.go
```diff
@@ -0,0 +1,131 @@
...
+
+}
+
+// Provider implements auth.Auther
+func (KerberosAuth) Provider() string {
+	return "krb5"
+}
+
+// KerberosType indicates that krb5 (Kerberos) authentication protocol is used.
+var KerberosType = [4]byte{'k', 'r', 'b', '5'}
```
sbinet: `s/KerberosType/Type/`

xrootd/xrdproto/auth/krb5/krb5.go
```diff
@@ -0,0 +1,131 @@
...
+	"gopkg.in/jcmturner/gokrb5.v5/config"
+	"gopkg.in/jcmturner/gokrb5.v5/credentials"
+	"gopkg.in/jcmturner/gokrb5.v5/crypto"
+	"gopkg.in/jcmturner/gokrb5.v5/messages"
+	"gopkg.in/jcmturner/gokrb5.v5/types"
+)
+
+// Default is a Kerberos 5 client configured from cached credentials.
+// If the credentials could not be correctly configured, Default will be nil.
+var Default auth.Auther
```
sbinet: `s/auth.Auther/*Auth/`
(so `Default.Provider()` will work even if `Default` is `nil`.)

xrootd/xrdproto/auth/krb5/krb5.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// WithClient creates a new KerberosAuth using the provided krb5 client.
+func WithClient(client *client.Client) *KerberosAuth {
+	return &KerberosAuth{client: client}
+
+}
+
+// Provider implements auth.Auther
+func (KerberosAuth) Provider() string {
```
sbinet: `s/(KerberosAuth)/(*Auth)/`
so calling `Provider` on a `nil` value will work, w/o segfaulting.

xrootd/xrdproto/auth/unix/unix.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"os/user"
+
+	"go-hep.org/x/hep/xrootd/xrdproto/auth"
+)
+
+// Default is an unix security provider configured from current username and group.
+// If the credentials could not be correctly configured, Default will be nil.
+var Default auth.Auther
```
sbinet: `s/auth.Auther/*Auth/`

xrootd/xrdproto/auth/unix/unix.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+	g, err := lookupGroupID(u)
+	if err != nil {
+		return
+	}
+	Default = &UnixAuth{User: u.Username, Group: g}
+}
+
+// UnixAuth implements unix security provider.
+type UnixAuth struct {
```
sbinet: `s/UnixAuth/Auth/`

xrootd/xrdproto/auth/unix/unix.go
```diff
@@ -0,0 +1,47 @@
...
+	User  string
+	Group string
+}
+
+// Provider implements auth.Auther
+func (UnixAuth) Provider() string {
```
sbinet: `s/UnixAuth/*Auth/`

xrootd/xrdproto/auth/unix/unix.go
```diff
@@ -0,0 +1,47 @@
...
+	return "unix"
+}
+
+// UnixType indicates that unix authentication protocol is used.
+var UnixType = [4]byte{'u', 'n', 'i', 'x'}
```
sbinet: `s/UnixType/Type/`

xrootd/xrdproto/auth/unix/unix.go
sbinet: also please add somewhere:
```go
var (
  _ auth.Auther = (*Auth)(nil)
)
```

xrootd/xrdproto/auth/krb5/krb5.go
EgorMatirov: In that case during:
```go
providers := []auth.Auther{krb5.Default, unix.Default}
	for _, provider := range providers {
		if provider != nil {
                     //...
```
`provider` is always not `nil` since there is an assignment from pointer to the interface. :-/

sbinet: right. ok, le'ts leave `var Default auth.Auther` then :)

sbinet:
thanks!

PR #266, opened 2018-06-21
[WIP] xrootd: add support for multiple sessions and redirect response

EgorMatirov:
Brief overview of redirect response ([specs](http://xrootd.org/doc/dev45/XRdv310.htm#_Toc464248805)):
1) It may occur at any time. Connection failures and server errors are treated as a redirect to the initial XRootD server ([spec](http://xrootd.org/doc/dev45/XRdv310.htm#_Toc464248789)).
2) Redirection may contain `token` which needs to be passed during `Login` request.
3) Redirection may contain `opaque` which need to be added to the filename associated with the operation.
4) Redirection may occur during request with a file handle. In that case, `open` request should be re-issued before proceeding with the current request. Also, probably, in that case, we should update the `handle` we have in `File`.
5) It looks to me like we can end up with multiple files being opened on different servers.

Sorry, WIP branch again, I would like to hear your opinion on current direction.

Btw, it's now able to cp the file from eospublic:
```
C:\Users\ematirov\Documents\tmp>xrd-cp  root://eospublic.cern.ch:1094//eos/opendata/atlas/OutreachDatasets/2016-07-29/MC/mc_173045.DYtautauM08to15.root .

C:\Users\ematirov\Documents\tmp>
```
Updates go-hep/hep#170.


xrootd/client/client.go
```diff
@@ -104,182 +94,77 @@ func NewClient(ctx context.Context, address string, username string, opts ...Opt
...
+	address := sessionID
+	client.mu.RLock()
+	session, ok := client.sessions[address]
+	client.mu.RUnlock()
+	if !ok {
+		return "", errors.Errorf("xrootd: session with id = %q was not found", sessionID)
 	}
 
-	var data []byte
+	redirection, err := session.Send(ctx, resp, req)
```
sbinet: shouldn't `err` be handled somehow?

xrootd/xrdproto/protocol.go
```diff
@@ -116,3 +118,62 @@ const (
 	// DataServer indicates whether this is a data server.
 	DataServer
 )
+
+type FilepathRequest interface {
```
sbinet: please document this interface and its methods.

xrootd/xrdproto/signing/signing.go
```diff
@@ -0,0 +1,92 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package signing contains implementation of a way to check if request should be signed
+//// according to XRootD protocol specification v. 3.1.0, p.75-76.
```
sbinet: please remove the extra `//`

xrootd/client/client.go
```diff
@@ -104,182 +94,77 @@ func NewClient(ctx context.Context, address string, username string, opts ...Opt
...
-	return resp.UnmarshalXrd(xrdenc.NewRBuffer(data))
+// If the resp is nil, then no response is stored.
+func (client *Client) Send(ctx context.Context, resp xrdproto.Response, req xrdproto.Request) (string, error) {
+	return client.SendSession(ctx, client.initialSessionID, resp, req)
 }
 
-func (client *Client) send(ctx context.Context, responseChannel mux.DataRecvChan, request []byte) ([]byte, error) {
-	if _, err := client.conn.Write(request); err != nil {
-		return nil, err
+func (client *Client) SendSession(ctx context.Context, sessionID string, resp xrdproto.Response, req xrdproto.Request) (string, error) {
```
sbinet: do we need to export this method?
it seems to me it can be kept internal. (ie: `s/SendSession/sendSession/`)

xrootd/client/client.go
```diff
@@ -104,182 +94,77 @@ func NewClient(ctx context.Context, address string, username string, opts ...Opt
...
+		// TODO: pass token to the session.Login()
+		session, err = client.getSession(address)
+		if err != nil {
+			return address, err
 		}
+		if filepathRequest, ok := req.(xrdproto.FilepathRequest); ok {
+			filepathRequest.SetOpaque(redirection.Opaque)
+		}
+		// TODO: we should check if the request contains file handle and re-issue open request in that case.
+		redirection, err = session.Send(ctx, resp, req)
```
sbinet: shouldn't this `err` also be handled somehow?

xrootd/client/client.go
```diff
@@ -104,182 +94,77 @@ func NewClient(ctx context.Context, address string, username string, opts ...Opt
...
-	}
-
-	if len(securityInfo.SecurityInformation) > 0 {
-		err = client.auth(ctx, securityInfo.SecurityInformation)
-		if err != nil {
-			client.Close()
-			return nil, err
-		}
-	}
+	session, err := NewSession(ctx, address, username, client)
```
sbinet: does `NewSession` need to be exported?

xrootd/xrdproto/protocol.go
EgorMatirov: Just an additional clarification: this interface will be implemented by all requests that use paths: `open`, `mv`, `rm`, `rmdir`, etc...
So if a redirect occurs as a response to the one of these request, an opaque data will be added to that request.

xrootd/client/client.go
```diff
@@ -104,182 +94,83 @@ func NewClient(ctx context.Context, address string, username string, opts ...Opt
...
-		data, err = client.sign(streamID, req.ReqID(), data)
+	for cnt := 10; redirection != nil && cnt > 0; cnt-- {
+		address = redirection.Host + ":" + strconv.Itoa(redirection.Port)
+		// TODO: pass token to the session.Login()
+		session, err = client.getSession(address)
 		if err != nil {
-			return nil, err
+			return address, err
+		}
+		if filepathRequest, ok := req.(xrdproto.FilepathRequest); ok {
```
sbinet: `s/filepathRequest/fpReq/`
or perhaps just even:
`s/filepathRequest/fp/`

(we see the type of that new variable in the type-assert, so `filepathRequest` is just noise)

xrootd/client/client.go
```diff
@@ -104,182 +94,83 @@ func NewClient(ctx context.Context, address string, username string, opts ...Opt
...
 // Send sends the request to the server and stores the response inside the resp.
-func (client *Client) Send(ctx context.Context, resp xrdproto.Response, req xrdproto.Request) error {
-	data, err := client.call(ctx, req)
-	if err != nil {
-		return err
-	}
-
-	return resp.UnmarshalXrd(xrdenc.NewRBuffer(data))
+// If the resp is nil, then no response is stored.
+func (client *Client) Send(ctx context.Context, resp xrdproto.Response, req xrdproto.Request) (string, error) {
```
sbinet: please also document what is the extra `string` return value.

xrootd/client/session.go
```diff
@@ -0,0 +1,252 @@
...
+				go func(streamID xrdproto.StreamID) {
+					// TODO: fetch sleep value from response.
+					time.Sleep(time.Second * 5)
+					if _, err := session.conn.Write(session.requests[streamID]); err != nil {
+						log.Printf("received error: %v", err)
+					}
+				}(header.StreamID)
+				continue
+			case xrdproto.Redirect:
+				port := binary.BigEndian.Uint32(resp.Data)
```
sbinet: this whole port+opaque+token extraction business should probably be lifted as a function, which could then be properly tested for a bunch of cases.
```go
func parseRedirect(raw []byte) (host string, port int, opaque, token string, err error) {
    ...
}
```

or, directly inside `internal/mux`:
```go
func ParseRedirection(raw []byte) (*Redirection, error) {
    ...
}
```

xrootd/client/session.go
```diff
@@ -0,0 +1,252 @@
...
+	data := wBuffer.Bytes()
+
+	if session.signRequirements.Needed(req) {
+		data, err = session.sign(streamID, req.ReqID(), data)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	session.requests[streamID] = data
```
sbinet: I am unsure whether this won't trigger a data race (or a concurrent map-write, that will make the runtime panic (with Go>=1.10.x))
IIRC, only `session.send` is thread-safe.
couldn't this be handled inside `session.send` ?

xrootd/internal/mux/mux.go
```diff
@@ -49,8 +49,16 @@ import (
...
 // during response decoding.
 type ServerResponse struct {
-	Data []byte
-	Err  error
+	Data        []byte
+	Err         error
+	Redirection *Redirection
+}
+
+type Redirection struct {
```
sbinet: please document that type and its fields.

xrootd/xrdproto/open/open.go
```diff
@@ -74,6 +77,22 @@ type Request struct {
...
+// Opaque implements xrdproto.FilepathRequest.Opaque.
+func (req *Request) Opaque() string {
+	separatorPos := strings.LastIndex(req.Path, "?")
+	return req.Path[separatorPos+1:]
+}
+
+// SetOpaque implements xrdproto.FilepathRequest.SetOpaque.
+func (req *Request) SetOpaque(opaque string) {
+	path := req.Path
+	separatorPos := strings.LastIndex(path, "?")
```
sbinet: `s/separatorPos/pos/`

xrootd/xrdproto/open/open.go
```diff
@@ -74,6 +77,22 @@ type Request struct {
 	Path    string
 }
 
+// Opaque implements xrdproto.FilepathRequest.Opaque.
+func (req *Request) Opaque() string {
+	separatorPos := strings.LastIndex(req.Path, "?")
```
sbinet: `s/separatorPos/pos/`

xrootd/xrdproto/protocol.go
```diff
@@ -25,7 +25,9 @@ const (
 	OkSoFar ResponseStatus = 4000
 	// Error indicates that an error occurred during request handling.
 	// Error code and error message are sent as part of response (see xrootd protocol specification v3.1.0, p. 27).
-	Error ResponseStatus = 4003
+	Error    ResponseStatus = 4003
+	Redirect ResponseStatus = 4004
```
sbinet: please document `Redirect` and `Wait`.

xrootd/xrdproto/signing/signing.go
```diff
@@ -0,0 +1,92 @@
...
+	"go-hep.org/x/hep/xrootd/xrdproto/xrdclose"
+)
+
+// Requirements implements a way to check if request should be signed
+// according to XRootD protocol specification v. 3.1.0, p.75-76.
+type Requirements struct {
+	requirements map[uint16]xrdproto.RequestLevel
+}
+
+// Needed return whether the request should be signed.
```
sbinet: `s/return/returns/`

xrootd/client/client.go
```diff
@@ -104,182 +94,83 @@ func NewClient(ctx context.Context, address string, username string, opts ...Opt
...
-	}
-	if err = req.MarshalXrd(&wBuffer); err != nil {
-		return nil, err
-	}
-	data := wBuffer.Bytes()
-
-	if client.signRequirements.Needed(req) {
-		data, err = client.sign(streamID, req.ReqID(), data)
+	for cnt := 10; redirection != nil && cnt > 0; cnt-- {
+		address = redirection.Host + ":" + strconv.Itoa(redirection.Port)
```
sbinet: this could be a method defined on `type Redirection`.

xrootd/client/client.go
```diff
@@ -104,182 +94,83 @@ func NewClient(ctx context.Context, address string, username string, opts ...Opt
...
-				// TODO: should we just ignore responses to unclaimed stream IDs?
-			}
-
-			if header.Status != xrdproto.OkSoFar {
-				client.mux.Unclaim(header.StreamID)
-			}
-		}
-	}
+	defer client.cancel()
+	// TODO: cleanup sessions.
```
sbinet: this would need to be addressed :)

xrootd/client/client.go
```diff
@@ -24,30 +24,27 @@ package client // import "go-hep.org/x/hep/xrootd/client"
...
-	signRequirements xrdproto.SignRequirements
-	seqID            int64
-	auths            map[string]auth.Auther
+	cancel   context.CancelFunc
+	auths    map[string]auth.Auther
+	username string
+
+	initialSessionID string
+	mu               sync.RWMutex
+	sessions         map[string]*session
```
sbinet: I am wondering about the impact (semantically) of having one `internal/mux.Mux` per-session instead of per-client.
operationally, it will reduce contention on the channels of the `mux.Mux`, but won't that muddy things when a request is initiated from, say, the main one (`initialSessionID`, right now), and ends up being handled by another one?

I guess some overall, overarching, description of the session dispatch mechanism is needed somewhere.
probably as a doc-comment on the `type session`.

xrootd/client/session.go
```diff
@@ -0,0 +1,252 @@
...
+
+	"github.com/pkg/errors"
+	"go-hep.org/x/hep/xrootd/internal/mux"
+	"go-hep.org/x/hep/xrootd/internal/xrdenc"
+	"go-hep.org/x/hep/xrootd/xrdproto"
+	"go-hep.org/x/hep/xrootd/xrdproto/signing"
+	"go-hep.org/x/hep/xrootd/xrdproto/sigver"
+)
+
+type session struct {
```
sbinet: please document this type.
it's not exported, but it's central to the use of that package and how the client ticks.
so, it's paramount to know how it works to be able to maintain the whole package :)

xrootd/client/client.go
```diff
@@ -104,182 +94,83 @@ func NewClient(ctx context.Context, address string, username string, opts ...Opt
...
-		return nil, err
-	}
-	if err = req.MarshalXrd(&wBuffer); err != nil {
-		return nil, err
-	}
-	data := wBuffer.Bytes()
-
-	if client.signRequirements.Needed(req) {
-		data, err = client.sign(streamID, req.ReqID(), data)
+	for cnt := 10; redirection != nil && cnt > 0; cnt-- {
```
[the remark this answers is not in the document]
EgorMatirov: It's not specified. I have changed it into a `maxRedirections` inside `Client`.

EgorMatirov:
> I don't yet quite clearly see what could be the code flow that could lead to having multiple files opened on different servers (through redirections.)
> I am not saying it can't happen! (I trust your judgement.)
> but could you clarify it out for me? :)

Well, as far as I understand, such scenario can happen:
1) `Client` makes a request to open a file `x` to the server `A`. (Makes a request to the initial session from the code perspective).
2) Server `A` responds with a redirect to the server `B` since only it contains the file. 
3) `Client` opens file `x` on the server `B`. (Makes a request to new session `B:port` from the code perspective).
4) `Client` makes a request to open a file `y` to the server `A`.
5) Server `A` responds with a redirect to the server `C` since only it contains the file.
6) `Clients` opens file `y` on the server `C`. (Makes a request to new session `C:port` from the code perspective).

This way, we have a 2 `File`s with different `sessionID`s, both the `B:port` and `C:port`.

I am not 100% sure, but it looks like a valid scenario from my point of view.

sbinet:
ok. yes, it's a valid scenario.
but I don't think it's a pathologic one.

I thought you meant that one could have an open file handle to `file-A` on server `A` that would also -- for some reason -- redirect to server `B` to get the content of that same `file-A`.
so you'd end up with 2 (open) file handles for the same request but on different servers.

I guess now my question is: why do you think having multiple open files on multiple servers could be an issue ? (short of forgetting to close them when needed, or consuming all file handles resources before closing some of them.)

xrootd/client/client.go
```diff
@@ -24,30 +24,30 @@ package client // import "go-hep.org/x/hep/xrootd/client"
...
+	auths    map[string]auth.Auther
+	username string
+	// initialSessionID is the sessionID of the server which is used as default
+	// for all requests that don't specify sessionID explicitly.
+	// Any failed request with another sessionID should be redirected to the initialSessionID.
+	// See http://xrootd.org/doc/dev45/XRdv310.pdf, page 11 for details.
+	initialSessionID string
+	mu               sync.RWMutex
+	sessions         map[string]*session
+	maxRedirections  int
```
sbinet: as `maxRedirections` isn't meant to be protected by the `sync.RWMutex`, please add an empty line before `maxRedirection` to clearly demark a logical block.

xrootd/client/client.go
```diff
@@ -104,182 +104,97 @@ func NewClient(ctx context.Context, address string, username string, opts ...Opt
...
 // Send sends the request to the server and stores the response inside the resp.
-func (client *Client) Send(ctx context.Context, resp xrdproto.Response, req xrdproto.Request) error {
-	data, err := client.call(ctx, req)
-	if err != nil {
-		return err
-	}
-
-	return resp.UnmarshalXrd(xrdenc.NewRBuffer(data))
+// If the resp is nil, then no response is stored.
+// Returns a session id which identifies the server that provided response.
```
sbinet: `s/Returns/Send returns/`

xrootd/client/session.go
```diff
@@ -0,0 +1,277 @@
...
+	"go-hep.org/x/hep/xrootd/internal/xrdenc"
+	"go-hep.org/x/hep/xrootd/xrdproto"
+	"go-hep.org/x/hep/xrootd/xrdproto/signing"
+	"go-hep.org/x/hep/xrootd/xrdproto/sigver"
+)
+
+// session is a connection to the specific XRootD server
+// which allows to send requests and receive responses.
+// Concurrent requests are supported.
+// Zero value is invalid, Session should be instantiated using NewSession.
```
sbinet: `s/Session should be instantiated using NewSession./session should be instantiated using newSession./`

xrootd/client/session.go
```diff
@@ -0,0 +1,277 @@
...
+			return
+		default:
+			if _, err := io.ReadFull(sess.conn, headerBytes); err != nil {
+				if ctx.Err() != nil {
+					// something happened to the context.
+					// ignore this error.
+					continue
+				}
+				panic(err)
+				// TODO: handle EOF by redirection as specified at http://xrootd.org/doc/dev45/XRdv310.pdf, page 11
```
sbinet: is this comment/TODO still valid ?

EgorMatirov:
> why do you think having multiple open files on multiple servers could be an issue ?

Sorry, maybe I wasn't clear, but I haven't thought about it in that way. :)
I was just giving an overview of situations \ scenarios that can occur with `redirect` response.
I don't see why this case can be unusable now - each file belongs to some session (i.e. server).

xrootd/xrdproto/protocol.go
```diff
@@ -26,6 +26,10 @@ const (
 	// Error indicates that an error occurred during request handling.
 	// Error code and error message are sent as part of response (see xrootd protocol specification v3.1.0, p. 27).
 	Error ResponseStatus = 4003
+	// Redirect indicates that client must re-issue request to the another server.
```
sbinet: `s/re-issue request to the another/re-issue the request to another/`

xrootd/xrdproto/protocol.go
```diff
@@ -26,6 +26,10 @@ const (
...
+	Redirect ResponseStatus = 4004
+	// Wait indicates that client must wait the indicated number of seconds and retry the request.
```
sbinet: `s/that client/that the client/`

sbinet:
> Sorry, maybe I wasn't clear, but I haven't thought about it in that way. :)

aha! ok :)

xrootd/client/session.go
EgorMatirov: I have added a redirection here. However, not everything in a worst-case scenario is handled yet, so there is still a plenty of `TODO`s.

PR #268, opened 2018-06-26
xrootd: implement FilepathRequest interface in corresponding requests

EgorMatirov:
This allows any request with a file path to have correct `opaque` after redirection.

### DIFF
--- a/xrootd/xrdproto/protocol.go
+++ b/xrootd/xrdproto/protocol.go
@@ -9,6 +9,7 @@ package xrdproto // import "go-hep.org/x/hep/xrootd/xrdproto"
 import (
 	"encoding/binary"
 	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 	"go-hep.org/x/hep/xrootd/internal/xrdenc"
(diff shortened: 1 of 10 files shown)


xrootd/xrdproto/protocol.go
```diff
@@ -182,3 +183,18 @@ func (o *SecurityOverride) UnmarshalXrd(dec *xrdenc.RBuffer) error {
 	o.RequestLevel = RequestLevel(dec.ReadU8())
 	return nil
 }
+
+// SetOpaque appends opaque data to the provided path.
+func SetOpaque(path, opaque string) string {
```
sbinet: I think the name `SetOpaque` is confusing.
the doc string above would suggest a better name to be `AppendOpaque`.
but it's not quite that, as it replaces the opaque part in `path` with what is provided with `opaque`.
hum...

perhaps the confusing part is that `SetOpaque` returns the new value instead of adding it inplace?
it definitely looks confusing at call site:
```go
req.Path = xrdproto.SetOpaque(req.Path, opaque)
```

perhaps if it was:
```go
xrdproto.SetOpaque(&req.Path, opaque)
```
it would look better (and we could keep the name `SetOpaque`.)

what do you think?
ok to change to:
```go
func SetOpaque(path *string, opaque string) string { ... }
```
?

(alternatively, we could introduce a `type Opaque string` with methods attached to handle this...)

PR #270, opened 2018-06-27
rootio-write

maloft:


sbinet:
looks good (well, that's my code, so... :P)

could you resolve the merge conflicts? (rebasing on the latest `origin/master`)
a lot happen since I first investigated this, last year...

maloft:
by the 2 last commits, I just want to check if WriteI8, WriteI16... give the right results

sbinet:
I think it's possible it doesn't run when there are conflicts.
it's also possible it's going through a rough patch.

you can run `go test ./...` inside the `rootio` directory :)

rootio/wbuffer.go
```diff
@@ -4,6 +4,159 @@
...
+func (w *wbuff) Write(p []byte) (int, error) {
+	if w.c >= len(w.p) {
+		return 0, io.EOF
+	}
+	n := copy(w.p[w.c:], p)
+	w.c += n
+	return n, nil
+}
+
+func (w *wbuff) WriteBufByte(p byte) error {
```
sbinet: `s/WriteBufByte/WriteByte/`

rootio/wbuffer.go
```diff
@@ -4,6 +4,159 @@
...
+	}
+	v := byte(i)
+	w.w.WriteBufByte(v)
+}
+
+func (w *WBuffer) WriteI16(i int16) {
+	if w.err != nil {
+		fmt.Errorf("error in WriteI16")
+	}
+	err := binary.Write(w.w, binary.BigEndian, i)
```
sbinet: this is fine as a first pass.
but it's not very efficiency-friendly as `binary.Write` will go through a bunch of `reflect` code (and reflection is slow) to write this `int16`.

a more efficient implementation would be to use `binary.ByteOrder.PutUint16` and the `wbuff` field of `WBuffer`.
(so we need to directly address its underlying buffer and make sure there's enough room.)

`binary.BigEndian` implements `binary.ByteOrder`.

maloft: This remark is just for WriteI16 or is it for all the write functions?

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+	if w.err != nil {
+		return
+	}
+	err := binary.Write(w.w, binary.BigEndian, i)
+	if err != nil {
+		fmt.Println(err)
+	}
+}
+
+func (w *WBuffer) WriteBool(boolean bool) {
```
sbinet: `s/boolean/v/`

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+	if w.err != nil {
+		return
+	}
+	err := binary.Write(w.w, binary.BigEndian, boolean)
```
sbinet: this probably won't work at runtime.
`binary.Write` doesn't have support for `bool`, IIRC.

prefer:
```go
var o byte = 0
if v {
    o = 1
}
w.err = binary.Write(w.w, binary.BigEndian, o)
```

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+	if w.err != nil {
+		return
+	}
+	err := binary.Write(w.w, binary.BigEndian, i)
+	if err != nil {
+		fmt.Println(err)
+	}
+}
+
+func (w *WBuffer) WriteF64(i float64) {
```
sbinet: `s/i/v/`

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+	if w.err != nil {
+		return
+	}
+	err := binary.Write(w.w, binary.BigEndian, i)
+	if err != nil {
+		fmt.Println(err)
+	}
+}
+
+func (w *WBuffer) WriteF32(i float32) {
```
sbinet: `s/i/v/`

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+	if w.err != nil {
+		return
+	}
+	err := binary.Write(w.w, binary.BigEndian, i)
+	if err != nil {
+		fmt.Println(err)
+	}
+}
+
+func (w *WBuffer) WriteU64(i uint64) {
```
sbinet: `s/i/v/`

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+}
+
+func (w *WBuffer) WriteU16(i uint16) {
+	if w.err != nil {
+		return
+	}
+	binary.BigEndian.PutUint16(w.w.p, uint16(i))
+}
+
+func (w *WBuffer) WriteU32(i uint32) {
```
sbinet: `s/i/v/`

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+
+func (w *WBuffer) WriteU8(i uint8) {
+	if w.err != nil {
+		return
+	}
+	v := byte(i)
+	w.w.WriteByte(v)
+}
+
+func (w *WBuffer) WriteU16(i uint16) {
```
sbinet: `s/i/v/`

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+
+	return &WBuffer{
+		w:      &wbuff{p: data, c: 0},
+		refs:   refs,
+		offset: offset,
+		sictx:  ctx,
+	}
+}
+
+func (w *WBuffer) WriteI8(i int8) {
```
sbinet: `s/i/v/`

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+	if w.err != nil {
+		return
+	}
+	v := byte(i)
+	w.w.WriteByte(v)
+}
+
+func (w *WBuffer) WriteI16(i int16) {
```
sbinet: `s/i/v/`

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+	if w.err != nil {
+		return
+	}
+	binary.BigEndian.PutUint16(w.w.p, uint16(i))
+
+}
+
+func (w *WBuffer) WriteI32(i int32) {
```
sbinet: `s/i/v/`

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+	if w.err != nil {
+		return
+	}
+	err := binary.Write(w.w, binary.BigEndian, i)
+	if err != nil {
+		fmt.Println(err)
+	}
+}
+
+func (w *WBuffer) WriteI64(i int64) {
```
sbinet: `s/i/v/`

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+		sictx:  ctx,
+	}
+}
+
+func (w *WBuffer) WriteI8(i int8) {
+	if w.err != nil {
+		return
+	}
+	v := byte(i)
+	w.w.WriteByte(v)
```
sbinet: here, you drop the `error` value that `WriteByte` returns, on the floor.
you should store it inside `w.err` instead.

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+}
+
+func (w *WBuffer) WriteI16(i int16) {
+	if w.err != nil {
+		return
+	}
+	binary.BigEndian.PutUint16(w.w.p, uint16(i))
```
sbinet: here, you don't check whether there is enough room inside the `w.w.p` buffer for `i`.
also, you write it at the beginning of that `p` slice of bytes.
you should instead write it at the current cursor location (_ie_ `w.w.c`)
and, after the write, you didn't update the current cursor location (_ie_, it should have advanced by 2 bytes.)

_e.g._:
```go
beg := w.w.c
end := w.w.c + 2
binary.BigEndian.PutUint16(w.w.p[beg:end], uint16(i))
w.w.c += 2
```

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+
+}
+
+func (w *WBuffer) WriteI32(i int32) {
+	if w.err != nil {
+		return
+	}
+	err := binary.Write(w.w, binary.BigEndian, i)
+	if err != nil {
```
sbinet: no, instead of printing the error, you should store it inside `w.err`:
```go
w.err = binary.Write(w.w, binary.BigEndian, i)
```

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+		fmt.Println(err)
+	}
+}
+
+func (w *WBuffer) WriteU8(i uint8) {
+	if w.err != nil {
+		return
+	}
+	v := byte(i)
+	w.w.WriteByte(v)
```
sbinet: here you also drop the error returned by `WriteByte`.
please store it inside `w.err`.

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+}
+
+func (w *WBuffer) WriteU16(i uint16) {
+	if w.err != nil {
+		return
+	}
+	binary.BigEndian.PutUint16(w.w.p, uint16(i))
```
sbinet: same comments than for `WriteI16`.

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+}
+
+func (w *WBuffer) WriteU32(i uint32) {
+	if w.err != nil {
+		return
+	}
+	err := binary.Write(w.w, binary.BigEndian, i)
```
sbinet: ditto.

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+		fmt.Println(err)
+	}
+}
+
+func (w *WBuffer) WriteU64(i uint64) {
+	if w.err != nil {
+		return
+	}
+	err := binary.Write(w.w, binary.BigEndian, i)
+	if err != nil {
```
sbinet: ditto.

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+		fmt.Println(err)
+	}
+}
+
+func (w *WBuffer) WriteF32(i float32) {
+	if w.err != nil {
+		return
+	}
+	err := binary.Write(w.w, binary.BigEndian, i)
```
sbinet: ditto.

rootio/wbuffer.go
```diff
@@ -4,6 +4,154 @@
...
+	if err != nil {
+		fmt.Println(err)
+	}
+}
+
+func (w *WBuffer) WriteF64(i float64) {
+	if w.err != nil {
+		return
+	}
+	err := binary.Write(w.w, binary.BigEndian, i)
```
sbinet: ditto.

rootio/wbuffer_test.go
```diff
@@ -0,0 +1,25 @@
...
+// license that can be found in the LICENSE file.
+
+package rootio
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestWbuffer(t *testing.T) {
```
sbinet: `s/TestWbuffer/TestWBuffer_WriteBool/`

rootio/wbuffer_test.go
```diff
@@ -0,0 +1,25 @@
...
+	data := make([]byte, 20)
+	wbuf := NewWBuffer(data, nil, 0, nil)
+	var i bool
```
sbinet: these 2 lines could just be written as:
```go
i := true
```

actually, please rename that variable as `want` instead of `i`.

rootio/wbuffer_test.go
```diff
@@ -0,0 +1,25 @@
...
+	i = true
+	wbuf.WriteBool(i)
+	rbuf := NewRBuffer(wbuf.w.p, nil, 0, nil)
+	v := rbuf.ReadBool()
```
sbinet: `s/v/got/`

rootio/wbuffer_test.go
```diff
@@ -0,0 +1,25 @@
...
+	if v != i {
+		t.Errorf("WriteBool was incorrect, got: %v, want: %v.", v, i)
```
sbinet: I would rather go with `Fatalf`:
```go
t.Fatalf("invalid value. got=%v, want=%v", got, want)
```

rootio/wbuffer_test.go
```diff
@@ -0,0 +1,25 @@
...
+	}
+	fmt.Printf("want: %v , got: %v\n", v, i)
```
sbinet: please remove.
we've already tested that :)

rootio/wbuffer.go
```diff
@@ -55,103 +54,108 @@ func NewWBuffer(data []byte, refs map[int64]interface{}, offset uint32, ctx Stre
...
 		return
 	}
-	err := binary.Write(w.w, binary.BigEndian, i)
-	if err != nil {
-		fmt.Println(err)
+	var o byte = 0
+	if v {
+		o = 1
 	}
+	w.err = binary.Write(w.w, binary.BigEndian, o)
```
sbinet: as the value that is now being written out is just a `byte`, one could as well directly use `w.w.WriteByte(o)`

rootio/wbuffer.go
```diff
@@ -55,103 +54,108 @@ func NewWBuffer(data []byte, refs map[int64]interface{}, offset uint32, ctx Stre
...
 
-func (w *WBuffer) WriteBool(boolean bool) {
+func (w *WBuffer) WriteString(s string) {
 	if w.err != nil {
 		return
 	}
-	err := binary.Write(w.w, binary.BigEndian, boolean)
-	if err != nil {
-		fmt.Println(err)
+	b1 := []byte(s)
```
sbinet: this is a bit long winded, and full of temporaries.
what about:
```go
b := make([]byte, len(s)+1)
b[0] = len(s)
copy(b[1:], []byte(s)...)
```

also: this will only work "small" strings.

rootio/wbuffer_test.go
```diff
@@ -5,21 +5,29 @@
 package rootio
 
 import (
-	"fmt"
 	"testing"
 )
 
-func TestWbuffer(t *testing.T) {
+func TestWbuffer_WriteBool(t *testing.T) {
```
sbinet: `s/TestWbuffer/TestWBuffer/` (note the `B`.)

rootio/wbuffer_test.go
```diff
@@ -5,21 +5,29 @@
...
 	data := make([]byte, 20)
 	wbuf := NewWBuffer(data, nil, 0, nil)
-	var i bool
-	i = true
-	wbuf.WriteBool(i)
+	want := true
```
sbinet: could you also add a test for when the `want` value is `false` ?

rootio/wbuffer_test.go
```diff
@@ -5,21 +5,29 @@
...
-	if v != i {
-		t.Errorf("WriteBool was incorrect, got: %v, want: %v.", v, i)
+	got := rbuf.ReadBool()
+	if got != want {
+		t.Fatalf("Invalid value. got:%v, want:%v", got, want)
 	}
-	fmt.Printf("want: %v , got: %v\n", v, i)
+}
 
+func TestWbuffer_WriteString(t *testing.T) {
```
sbinet: `s/TestWbuffer/TestWBuffer/` (note the `B`.)

rootio/wbuffer_test.go
```diff
@@ -5,21 +5,29 @@
...
+	data := make([]byte, 20)
+	wbuf := NewWBuffer(data, nil, 0, nil)
+	want := "hello world"
```
sbinet: could you also add a test for a long string?
_e.g.:_
```go
want := strings.Repeat("=", 254)
want := strings.Repeat("=", 255)
want := strings.Repeat("=", 256)
want := strings.Repeat("=", 512)
```

rootio/wbuffer.go
```diff
@@ -141,19 +141,22 @@ func (w *WBuffer) WriteBool(v bool) {
...
-	for i := range b1 {
-		b[i+1] = b1[i]
+	l := len(v)
+	b := make([]byte, l + 1)
+	if l < 255 {
+		b[0] = byte(l)
+		copy(b[1:],[]byte(v))
+	}else{
+		b[0] = 255
+		b[1] = byte(l)
```
sbinet: no, if the string's length is larger than `255`, we need more than 1 `byte` to store that length.
`ROOT` chose to encode it over 4 bytes (_ie_ a `uint32`).
that's why the reading counterpart of `WriteString` does this:
https://github.com/go-hep/hep/blob/master/rootio/rbuffer.go#L149
```go
	if u8 == 255 {
		// large string
		n = int(r.ReadU32())
	}
```

so you need to do the reverse:
- write `255` at `b[0]`
- write the actual length of the string as a `uint32` over `b[1:5]`
- write the string starting from `b[5]`.

maloft: i tried it this way and still doesn't work tho (cf last commit) :)

rootio/wbuffer.go
```diff
@@ -4,6 +4,168 @@
...
+		b[0] = byte(l)
+		copy(b[1:], []byte(v))
+		var n int
+		n, w.err = w.w.Write(b)
+		w.w.c += n
+	} else {
+		b[0] = 255
+		copy(b[5:], []byte(v))
+		w.w.WriteByte(b[0]) //writing 255 in the first byte
+		w.w.c++
```
sbinet: that's already done by `func (w *wbuff) WriteByte(p byte) error`.

rootio/wbuffer.go
```diff
@@ -4,6 +4,168 @@
...
+		w.WriteU32(uint32(l)) //writing the length as a uint32
+		w.w.c += 4
```
sbinet: that's already done by `func (w *WBuffer) WriteU32(v uint32)`.

rootio/wbuffer.go
```diff
@@ -4,6 +4,168 @@
...
+		copy(b[5:], []byte(v))
+		var n int
+		n, w.err = w.w.Write(b[5:]) //writing the string
+		w.w.c += n
```
sbinet: that's already done by `func (w *wbuff) Write(p []byte) (int, error)`.

rootio/wbuffer.go
```diff
@@ -153,19 +154,12 @@ func (w *WBuffer) WriteString(v string) { //FIXME :works only on small strings
 	if l < 255 {
 		b[0] = byte(l)
 		copy(b[1:], []byte(v))
-		var n int
-		n, w.err = w.w.Write(b)
-		w.w.c += n
+		w.w.Write(b)
```
sbinet: here you drop the error that is returned by `w.w.Write` on the floor.
you should store it inside `w.err`.

rootio/wbuffer.go
```diff
@@ -153,19 +154,12 @@ func (w *WBuffer) WriteString(v string) { //FIXME :works only on small strings
...
 	} else {
 		b[0] = 255
-		copy(b[5:], []byte(v))
-		w.w.WriteByte(b[0]) //writing 255 in the first byte
-		w.w.c++
+		w.w.WriteByte(b[0])   //writing 255 in the first byte
```
sbinet: here you drop the error that is returned by `w.w.WriteByte` on the floor.
you should store it inside `w.err` (and return early if it's non-`nil`.)

rootio/wbuffer.go
```diff
@@ -153,19 +154,12 @@ func (w *WBuffer) WriteString(v string) { //FIXME :works only on small strings
...
 		w.WriteU32(uint32(l)) //writing the length as a uint32
-		w.w.c += 4
 		copy(b[5:], []byte(v))
-		var n int
-		n, w.err = w.w.Write(b[5:]) //writing the string
-		w.w.c += n
+		w.w.Write(b[5:]) //writing the string
```
sbinet: here you drop the error that is returned by `w.w.Write` on the floor.
you should store it inside `w.err`.

rootio/wbuffer.go
```diff
@@ -4,6 +4,166 @@
...
+	w.err = w.w.WriteByte(o)
+}
+
+func (w *WBuffer) WriteString(v string) {
+	if w.err != nil {
+		return
+	}
+	l := len(v)
+	if l < 255 {
+		b := make([]byte, l+1)
```
sbinet: now that the mechanics work, we can simplify this a bit: no need to go through this temporary `b` buffer:
- first write the length of the string (as a single `byte`)
- write the string (as a `[]byte`)

rootio/wbuffer.go
```diff
@@ -4,6 +4,166 @@
...
+		b[0] = byte(l)
+		copy(b[1:], []byte(v))
+		_, w.err = w.w.Write(b)
+	} else {
+		b := make([]byte, l+5)
```
sbinet: similarly, now that the mechanics work, we can also simplify this a bit, foregoing completely the temporary buffer as well:
- first write `255` as a `byte` value
- write the `uint32` length
- write the string as a `[]byte` value.

rootio/wbuffer_test.go
```diff
@@ -0,0 +1,37 @@
...
+	want := true
+	wbuf.WriteBool(want)
+	rbuf := NewRBuffer(wbuf.w.p, nil, 0, nil)
+	got := rbuf.ReadBool()
+	if got != want {
+		t.Fatalf("Invalid value. got:%v, want:%v", got, want)
+	}
+}
+
+func TestWBuffer_WriteString(t *testing.T) {
```
sbinet: ok. but we also need to test for a short string.

sbinet:
(also, please rebase on top of the latest `master`.)

rootio/wbuffer.go
```diff
@@ -4,6 +4,164 @@
...
+	w.err = w.w.WriteByte(o)
+}
+
+func (w *WBuffer) WriteString(v string) {
+	if w.err != nil {
+		return
+	}
+	l := len(v)
+	if l < 255 {
+		w.err = w.w.WriteByte(byte(len(v)))
```
sbinet: ok, so now, we could directly use `WriteU8` so all the business of checking `w.err` could be removed here.

rootio/wbuffer.go
```diff
@@ -4,6 +4,164 @@
...
+		if w.err != nil {
+			return
+		}
+		_, w.err = w.w.Write([]byte(v))
```
sbinet: if we were to add a method, say `write([]byte)` on `WBuffer` that performs the same kind of checks than the `WriteU8`, `WriteF64`, etc... ones, then this whole code could be:
```go
w.WriteU8(uint8(l))
w.write([]byte(v))
```

rootio/wbuffer.go
```diff
@@ -4,6 +4,164 @@
...
+	} else {
+		w.err = w.w.WriteByte(255)
```
sbinet: similarly, we could write it like so:
```go
w.WriteU8(255)
w.WriteU32(uint32(l))
w.write([]byte(v))
```

rootio/wbuffer_test.go
```diff
@@ -0,0 +1,46 @@
...
+	got := rbuf.ReadString()
+	if got != want {
+		t.Fatalf("Invalid value.\ngot: %q\nwant:%q", got, want)
+	}
+}
+
+func TestWBuffer_WriteString_Short(t *testing.T) {
+	data := make([]byte, 520)
+	wbuf := NewWBuffer(data, nil, 0, nil)
+	want := strings.Repeat("=", 254)
```
sbinet: ok. but as I said in a previous comment, we should test the `255` boundary as well.
so, the same kind of code (probably in a `for`-loop), but for `i` running from `0` to `512`.

then, we could merge the 2 `TestWBuffer_WriteString_xyz` functions into `TestWBuffer_WriteString`

rootio/wbuffer_test.go
```diff
@@ -0,0 +1,36 @@
...
+func TestWBuffer_WriteString(t *testing.T) {
+	data := make([]byte, 520)
+	for i := 0; i < 512; i++ {
+		wbuf := NewWBuffer(data, nil, 0, nil)
+		want := strings.Repeat("=", i)
+		wbuf.WriteString(want)
+		rbuf := NewRBuffer(wbuf.w.p, nil, 0, nil)
+		got := rbuf.ReadString()
+		if got != want {
+			t.Fatalf("Invalid value.\ngot: %q\nwant:%q", got, want)
```
sbinet: could you also add an indication of which iteration failed?
_e.g.:_
```go
t.Fatalf("Invalid value for len=%d.\ngot: %q\nwant:%q", i, got, want)
```

Issue #175, opened 2018-04-24
xrootd: implement a xrootd-fuse application


sbinet:
this would be the application launched by `systemd`, `upstart` or directly by the user, actually mounting a remote `xrootd` server content on the local machine.

sbinet:
closed by #271.

Issue #174, opened 2018-04-24
xrootd: implement a FUSE backend for xrootd


sbinet:

EgorMatirov:
Regarding library choice... There is:
- https://github.com/bazil/fuse: 2 commits for the last 2 years, 61 open issues, 24 PRs. Supports Linux, OS X.
- https://github.com/hanwen/go-fuse: 15 commits since 2018, 18 open issues, 8 PRs. Has a list of known problems, but from first sight, there is nothing very scary. Supports Linux, OS X.
- https://github.com/jacobsa/fuse: not so much commits since 2018, 5 open issues, 3 PRs. Has some [application](https://github.com/gohxs/cloudmount) built on top of it (I am not saying that others don't have). Supports Linux, not sure about OS X.
- https://github.com/billziss-gh/cgofuse: a lot of commits since 2018, 4 open issues. Supports macOS,*BSD, Linux, Windows. Uses CGO.

`bazil/fuse` looks like a most popular choice, however, it bothers me that there is a number of open issues and not much development is happening.
`cgofuse` looks like a good promise, but I'm not sure that it's a good thing to use CGO. :)

So IMHO, priority is:
1) `cgofuse` if CGO is acceptable;
2) `go-fuse` since it has less open issues and more active development;
3) `bazil/fuse`.

What do you think?

sbinet:
closed by #271 
(there are missing syscalls such as `hardlink` and `symlink` but these aren't available from the `xrootd` protocol.)

PR #271, opened 2018-06-29
xrootd/{cmd/xrootd-fuse,xrdfuse}: first import

EgorMatirov:
Supports:
- ls
- echo "..." > file
- echo "..." >> file
- cat file

Note: probably I should move error codes to the xrdproto package?

Updates go-hep/hep#174.
Updates go-hep/hep#175.

### DIFF
--- a/xrootd/cmd/xrootd-fuse/main.go
+++ b/xrootd/cmd/xrootd-fuse/main.go
@@ -1,0 +1,95 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Command xrootd-fuse mounts the directory contents of a remote xrootd server to a local directory.
+//
+// Usage:
+//
+//  $> xrootd-fuse [OPTIONS] <remote-dir> <local-dir>
+//
+// Example:
+//
+//  $> xrootd-fuse root://server.example.com/some/dir /mnt
+//  $> xrootd-fuse -v root://server.example.com/some/dir /mnt
+//
+// Options:
+//   -v	enable verbose mode
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"os/signal"
+
+	"github.com/hanwen/go-fuse/fuse/nodefs"
+	"github.com/hanwen/go-fuse/fuse/pathfs"
+	"go-hep.org/x/hep/xrootd/client"
+	"go-hep.org/x/hep/xrootd/xrdfuse"
+	"go-hep.org/x/hep/xrootd/xrdio"
+)
+
+func init() {
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr, `xrootd-fuse mounts the directory contents of a remote xrootd server to a local directory.
+
+Usage:
+
+ $> xrootd-fuse [OPTIONS] <remote-dir> <local-dir>
+
+Example:
+
+ $> xrootd-fuse root://server.example.com/some/dir /mnt
+ $> xrootd-fuse -v root://server.example.com/some/dir /mnt
+
+Options:
+`)
+		flag.PrintDefaults()
+	}
+}
+
+func main() {
+	log.SetPrefix("xrootd-fuse: ")
+	log.SetFlags(0)
+
+	verbose := flag.Bool("v", false, "enable verbose mode")
+
+	flag.Parse()
+
+	if flag.NArg() != 2 {
+		flag.Usage()
+		log.Fatalf("missing directory operands")
+	}
+
+	url, err := xrdio.Parse(flag.Arg(0))
+
+	c, err := client.NewClient(context.Background(), url.Addr, url.User)
+	if err != nil {
+		log.Fatalf("could not create client: %v", err)
+	}
+
+	fs := xrdfuse.NewFS(c, url.Path, func(e error) {
+		log.Println(e)
+	})
+	nfs := pathfs.NewPathNodeFs(fs, nil)
+	server, _, err := nodefs.MountRoot(flag.Arg(1), nfs.Root(), &nodefs.Options{
+		Debug: *verbose,
+	})
+	if err != nil {
+		log.Fatalf("could not mount: %v", err)
+	}
+
+	ch := make(chan os.Signal, 1)
+	signal.Notify(ch, os.Interrupt)
+
+	go server.Serve()
+
+	<-ch
+	err = server.Unmount()
+	if err != nil {
+		log.Fatalf("could not unmount: %v", err)
+	}
+}
--- a/xrootd/xrdfuse/file.go
+++ b/xrootd/xrdfuse/file.go
@@ -1,0 +1,97 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package xrdfuse // import "go-hep.org/x/hep/xrootd/xrdfuse"
+
+import (
+	"context"
+
+	"github.com/hanwen/go-fuse/fuse"
+	"github.com/hanwen/go-fuse/fuse/nodefs"
+	"github.com/pkg/errors"
+	"go-hep.org/x/hep/xrootd/xrdfs"
+)
+
+// File represents a file on the remote XRootD server.
+type File struct {
+	nodefs.File
+	xrdfile xrdfs.File
+	fs      *FS
+}
+
+// Read implements nodefs.File.Read
+func (f *File) Read(dest []byte, off int64) (fuse.ReadResult, fuse.Status) {
+	n, err := f.xrdfile.ReadAt(dest, off)
+	if err != nil {
+		f.fs.handler(errors.WithMessage(err, "xrdfuse: error calling ReadAt"))
+		return nil, fuse.EIO
+	}
+
+	return fuse.ReadResultData(dest[:n]), fuse.OK
+}
+
+// Write implements nodefs.File.Write
+func (f *File) Write(data []byte, off int64) (uint32, fuse.Status) {
+	n, err := f.xrdfile.WriteAt(data, off)
+	if err != nil {
+		f.fs.handler(errors.WithMessage(err, "xrdfuse: error calling WriteAt"))
+		return 0, fuse.EIO
+	}
+
+	return uint32(n), fuse.OK
+}
+
+// Truncate implements nodefs.File.Truncate
+func (f *File) Truncate(size uint64) fuse.Status {
+	err := f.xrdfile.Truncate(context.Background(), int64(size))
+	if err != nil {
+		f.fs.handler(errors.WithMessage(err, "xrdfuse: error calling Truncate"))
+		return fuse.EIO
+	}
+
+	return fuse.OK
+}
+
+// Fsync implements nodefs.File.Fsync
+func (f *File) Fsync(flags int) (code fuse.Status) {
+	return f.Flush()
+}
+
+// Flush implements nodefs.File.Flush
+func (f *File) Flush() (code fuse.Status) {
+	err := f.xrdfile.Sync(context.Background())
+	if err != nil {
+		f.fs.handler(errors.WithMessage(err, "xrdfuse: error calling Sync"))
+		return fuse.EIO
+	}
+
+	return fuse.OK
+}
+
+// GetAttr implements nodefs.File.GetAttr
+func (f *File) GetAttr(out *fuse.Attr) fuse.Status {
+	stat, err := f.xrdfile.Stat(context.Background())
+	if err != nil {
+		f.fs.handler(errors.WithMessage(err, "xrdfuse: error calling Stat"))
+		return fuse.EIO
+	}
+
+	out.Size = uint64(stat.Size())
+	out.Mtime = uint64(stat.Mtime)
+	out.Mode = entryStatToMode(stat)
+
+	return fuse.OK
+}
+
+// Release implements nodefs.File.Release
+func (f *File) Release() {
+	err := f.xrdfile.Close(context.Background())
+	if err != nil {
+		f.fs.handler(errors.WithMessage(err, "xrdfuse: error calling Close"))
+	}
+}
+
+var (
+	_ nodefs.File = (*File)(nil)
+)
--- a/xrootd/xrdfuse/xrdfuse.go
+++ b/xrootd/xrdfuse/xrdfuse.go
@@ -1,0 +1,262 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package xrdfuse contains the implementation of the FUSE API
+// accessing a remote filesystem served over the XRootD protocol.
+package xrdfuse // import "go-hep.org/x/hep/xrootd/xrdfuse"
+
+import (
+	"context"
+	"os"
+	"path"
+
+	"github.com/hanwen/go-fuse/fuse"
+	"github.com/hanwen/go-fuse/fuse/nodefs"
+	"github.com/hanwen/go-fuse/fuse/pathfs"
+	"github.com/pkg/errors"
+	"go-hep.org/x/hep/xrootd/client"
+	"go-hep.org/x/hep/xrootd/xrdfs"
+	"go-hep.org/x/hep/xrootd/xrdproto"
+)
+
+// FS implements a pathfs.FileSystem that makes requests to the remote server over the XRootD protocol.
+type FS struct {
+	pathfs.FileSystem
+	client  *client.Client
+	xrdfs   xrdfs.FileSystem
+	root    string
+	handler ErrorHandler
+}
+
+// ErrorHandler is the function which handles occurred error (e.g. logs it).
+type ErrorHandler func(error)
+
+// NewFS returns a new path.FileSystem representing the filesystem on the remote XRootD server.
+// client is a client connected to the remote XRootD server.
+// root is the path to the remote directory to be used as a root directory.
+// handler is the function which handles occurred error (e.g. logs it). If the handler is nil,
+// then a default handler is used that does nothing.
+func NewFS(client *client.Client, root string, handler ErrorHandler) *FS {
+	if handler == nil {
+		handler = func(error) {}
+	}
+	return &FS{
+		FileSystem: pathfs.NewDefaultFileSystem(),
+		client:     client,
+		xrdfs:      client.FS(),
+		root:       root,
+		handler:    handler,
+	}
+}
+
+// GetAttr implements pathfs.FileSystem.GetAttr
+func (fs *FS) GetAttr(name string, ctx *fuse.Context) (*fuse.Attr, fuse.Status) {
+	stat, err := fs.xrdfs.Stat(context.Background(), path.Join(fs.root, name))
+	status := errorToStatus(err)
+	if status == fuse.EIO {
+		fs.handler(errors.WithMessage(err, "xrdfuse: error calling Stat"))
+	}
+	if status != fuse.OK {
+		return nil, status
+	}
+
+	return &fuse.Attr{
+		Size:  uint64(stat.Size()),
+		Mtime: uint64(stat.Mtime),
+		Mode:  entryStatToMode(stat),
+	}, fuse.OK
+}
+
+// OpenDir implements pathfs.FileSystem.OpenDir
+func (fs *FS) OpenDir(name string, ctx *fuse.Context) (c []fuse.DirEntry, code fuse.Status) {
+	entries, err := fs.xrdfs.Dirlist(context.Background(), path.Join(fs.root, name))
+	status := errorToStatus(err)
+	if status == fuse.EIO {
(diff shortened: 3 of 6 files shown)


xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,166 @@
...
+	for _, entry := range entries {
+		s = append(s, fuse.DirEntry{Name: entry.Name(), Mode: entryStatToMode(entry)})
+	}
+
+	return s, fuse.OK
+}
+
+func convertFlagsToMode(flags uint32) xrdfs.OpenMode {
+	var result xrdfs.OpenMode
+	if flags&uint32(os.O_RDWR) != 0 {
```
sbinet: this would be more readable with a `switch`:
```go
switch {
case flags&uint32(os.O_RDWR) != 0:
case flags&uint32(os.O_RDONLY) != 0:
case flags&uint32(os.O_WRONLY) != 0:
}
```

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,166 @@
...
+		xrdfs:      client.FS(),
+		root:       root,
+	}
+}
+
+// GetAttr implements pathfs.FileSystem.GetAttr
+func (fs *fs) GetAttr(name string, ctx *fuse.Context) (*fuse.Attr, fuse.Status) {
+	stat, err := fs.xrdfs.Stat(context.Background(), path.Join(fs.root, name))
+	if serverError, ok := err.(xrdproto.ServerError); ok {
+		if serverError.Code == 3011 {
```
sbinet: please remove this "magic" constant.

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,166 @@
...
+	return result
+}
+
+// Open implements pathfs.FileSystem.Open
+func (fs *fs) Open(name string, flags uint32, ctx *fuse.Context) (file nodefs.File, code fuse.Status) {
+	mode := convertFlagsToMode(flags)
+	options := convertFlagsToOptions(flags)
+	f, err := fs.xrdfs.Open(context.Background(), path.Join(fs.root, name), mode, options)
+	if serverError, ok := err.(xrdproto.ServerError); ok {
+		if serverError.Code == 3006 {
```
sbinet: please remove this "magic" constant.

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,166 @@
...
+	out.Mode = entryStatToMode(stat)
+
+	return &out, fuse.OK
+}
+
+// OpenDir implements pathfs.FileSystem.OpenDir
+func (fs *fs) OpenDir(name string, ctx *fuse.Context) (c []fuse.DirEntry, code fuse.Status) {
+	entries, err := fs.xrdfs.Dirlist(context.Background(), path.Join(fs.root, name))
+	if err != nil {
+		log.Printf("xrdfuse: error calling dirlist: %v\n", err)
```
sbinet: perhaps, instead of logging this and hoping somebody might look at the screen, we could store this `err` value on the side ? (as a field of `fs` ?)

xrootd/cmd/xrootd-fuse/main.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/hanwen/go-fuse/fuse/nodefs"
+	"github.com/hanwen/go-fuse/fuse/pathfs"
+	"go-hep.org/x/hep/xrootd/client"
+	"go-hep.org/x/hep/xrootd/xrdfuse"
+	"go-hep.org/x/hep/xrootd/xrdio"
+)
+
+func init() {
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr, `xrootd-fuse mounts directory contents on a remote xrootd server to a local directory.
```
sbinet: `s/mounts directory/mounts the directory/`
`s/on a remote/of a remote/`

xrootd/cmd/xrootd-fuse/main.go
```diff
@@ -0,0 +1,86 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Command xrootd-fuse mounts directory contents on a remote xrootd server to a local directory.
```
sbinet: `s/mounts directory contents on a/mounts the directory contents of a/`

xrootd/xrdfuse/file.go
```diff
@@ -0,0 +1,92 @@
...
+type File struct {
+	nodefs.File
+	xrdfile xrdfs.File
+}
+
+// Read implements nodefs.File.Read
+func (f *File) Read(dest []byte, off int64) (fuse.ReadResult, fuse.Status) {
+	n, err := f.xrdfile.ReadAt(dest, off)
+	if err != nil {
+		log.Printf("xrdfuse: error calling readAt: %v\n", err)
```
sbinet: ditto.
alternatively, perhaps there's a way to send our "error" messages to `fuse` "debug" stream?
(hum... doesn't seem to be possible...)

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,166 @@
+// Copyright 2018 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package xrdfuse contains implementation of a FUSE api for the XRootD.
```
sbinet: `s/contains implementation of a FUSE api for the XRootD./contains the implementation of the FUSE API accessing a remote filesystem served over the XRootD protocol./`

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,166 @@
...
+)
+
+type fs struct {
+	pathfs.FileSystem
+	client *client.Client
+	xrdfs  xrdfs.FileSystem
+	root   string
+}
+
+// NewFS returns a new path.FileSystem representing filesystem on the remote XRootD server.
```
sbinet: `s/representing filesystem/representing the filesystem/`

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,166 @@
...
+// client is a client to the remote XRootD server.
```
sbinet: `s/a client to/a client connected to/`

xrootd/xrdfuse/xrdfuse.go
EgorMatirov: What about having a `chan error`, so we will be able to receive all errors in `xrootd-fuse` and log them?

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		log.Printf("xrdfuse: error calling close: %v\n", err)
+		return fuse.EIO
+	}
+
+	return fuse.OK
+}
+
+// Rename implements pathfs.FileSystem.Rename
+func (fs *fs) Rename(oldName string, newName string, ctx *fuse.Context) (code fuse.Status) {
```
sbinet: not sure we need the named return value here.

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		log.Printf("xrdfuse: error calling rename: %v\n", err)
+		return fuse.EIO
+	}
+
+	return fuse.OK
+}
+
+// Unlink implements pathfs.FileSystem.Unlink
+func (fs *fs) Unlink(name string, ctx *fuse.Context) (code fuse.Status) {
```
sbinet: ditto

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		log.Printf("xrdfuse: error calling removeFile: %v\n", err)
+		return fuse.EIO
+	}
+
+	return fuse.OK
+}
+
+// Rmdir implements pathfs.FileSystem.Rmdir
+func (fs *fs) Rmdir(name string, ctx *fuse.Context) (code fuse.Status) {
```
sbinet: ditto

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		log.Printf("xrdfuse: error calling mkdir: %v\n", err)
+		return fuse.EIO
+	}
+
+	return fuse.OK
+}
+
+// Chmod implements pathfs.FileSystem.Chmod
+func (fs *fs) Chmod(name string, mode uint32, ctx *fuse.Context) (code fuse.Status) {
```
sbinet: ditto

xrootd/xrdfuse/xrdfuse.go
sbinet: that could be an option.
we should have a look at how it could be used in real life as a user, though.

xrootd/cmd/xrootd-fuse/main.go
```diff
@@ -0,0 +1,86 @@
...
+		fmt.Fprintf(os.Stderr, `xrootd-fuse mounts the directory contents of a remote xrootd server to a local directory.
+
+Usage:
+
+ $> xrootd-fuse [OPTIONS] <remote-dir> <local-dir>
+
+Example:
+
+ $> xrootd-fuse root://server.example.com/some/dir /mnt
+ $> xrootd-fuse -debug=true root://server.example.com/some/dir
```
sbinet: isn't this example missing the local dir?

xrootd/cmd/xrootd-fuse/main.go
```diff
@@ -0,0 +1,86 @@
...
+// Command xrootd-fuse mounts the directory contents of a remote xrootd server to a local directory.
+//
+// Usage:
+//
+//  $> xrootd-fuse [OPTIONS] <remote-dir> <local-dir>
+//
+// Example:
+//
+//  $> xrootd-fuse root://server.example.com/some/dir /mnt
+//  $> xrootd-fuse -debug=true root://server.example.com/some/dir
```
sbinet: isn't this example missing the local dir?

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// Open implements pathfs.FileSystem.Open
+func (fs *fs) Open(name string, flags uint32, ctx *fuse.Context) (file nodefs.File, code fuse.Status) {
+	mode := convertFlagsToMode(flags)
+	options := convertFlagsToOptions(flags)
+	f, err := fs.xrdfs.Open(context.Background(), path.Join(fs.root, name), mode, options)
+	if serverError, ok := err.(xrdproto.ServerError); ok {
+		if serverError.Code == xrdproto.InvalidRequestCode {
+			// Is is possible that request is invalid because file already exists.
```
sbinet: `s/Is is/It is/`
`s/that request/the request/`
`s/file/the file/`

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,230 @@
...
+			// O_CREAT flag can be passed to the fuse api despite the fact that file
```
sbinet: `s/api/API/`

xrootd/xrdfuse/xrdfuse.go
EgorMatirov: It looks like something like `FS.AddErrorHandler(...)` will be better since it will avoid possible blocks and will not require extra cleanup.

What do you think about it?

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -74,8 +76,20 @@ func (fs *FS) GetAttr(name string, ctx *fuse.Context) (*fuse.Attr, fuse.Status)
 // OpenDir implements pathfs.FileSystem.OpenDir
 func (fs *FS) OpenDir(name string, ctx *fuse.Context) (c []fuse.DirEntry, code fuse.Status) {
 	entries, err := fs.xrdfs.Dirlist(context.Background(), path.Join(fs.root, name))
+	if serverError, ok := err.(xrdproto.ServerError); ok {
+		switch serverError.Code {
```
sbinet: Perhaps all this handling (and in other functions) could be factorized in a function that translates an xrd error into a fuse one?

xrootd/xrdfuse/file.go
```diff
@@ -0,0 +1,109 @@
...
+	xrdfile xrdfs.File
+	fs      *FS
+}
+
+// Read implements nodefs.File.Read
+func (f *File) Read(dest []byte, off int64) (fuse.ReadResult, fuse.Status) {
+	n, err := f.xrdfile.ReadAt(dest, off)
+	if err != nil {
+		if f.fs.handler != nil {
+			f.fs.handler(errors.WithMessage(err, "xrdfuse: error calling readAt"))
```
sbinet: `s/readAt/ReadAt/`

xrootd/xrdfuse/file.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return fuse.ReadResultData(dest[:n]), fuse.OK
+}
+
+// Write implements nodefs.File.Write
+func (f *File) Write(data []byte, off int64) (uint32, fuse.Status) {
+	n, err := f.xrdfile.WriteAt(data, off)
+	if err != nil {
+		if f.fs.handler != nil {
+			f.fs.handler(errors.WithMessage(err, "xrdfuse: error calling writeAt"))
```
sbinet: `s/writeAt/WriteAt/`

xrootd/xrdfuse/file.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return uint32(n), fuse.OK
+}
+
+// Truncate implements nodefs.File.Truncate
+func (f *File) Truncate(size uint64) fuse.Status {
+	err := f.xrdfile.Truncate(context.Background(), int64(size))
+	if err != nil {
+		if f.fs.handler != nil {
+			f.fs.handler(errors.WithMessage(err, "xrdfuse: error calling truncate"))
```
sbinet: `s/truncate/Truncate/`

xrootd/xrdfuse/file.go
```diff
@@ -0,0 +1,109 @@
...
+func (f *File) Fsync(flags int) (code fuse.Status) {
+	return f.Flush()
+}
+
+// Flush implements nodefs.File.Flush
+func (f *File) Flush() (code fuse.Status) {
+	err := f.xrdfile.Sync(context.Background())
+	if err != nil {
+		if f.fs.handler != nil {
+			f.fs.handler(errors.WithMessage(err, "xrdfuse: error calling sync"))
```
sbinet: `s/sync/Sync/`

xrootd/xrdfuse/file.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return fuse.OK
+}
+
+// GetAttr implements nodefs.File.GetAttr
+func (f *File) GetAttr(out *fuse.Attr) fuse.Status {
+	stat, err := f.xrdfile.Stat(context.Background())
+	if err != nil {
+		if f.fs.handler != nil {
+			f.fs.handler(errors.WithMessage(err, "xrdfuse: error calling stat"))
```
sbinet: `s/stat/Stat/`

xrootd/xrdfuse/file.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return fuse.OK
+}
+
+// Release implements nodefs.File.Release
+func (f *File) Release() {
+	err := f.xrdfile.Close(context.Background())
+	if err != nil {
+		if f.fs.handler != nil {
+			f.fs.handler(errors.WithMessage(err, "xrdfuse: error calling close"))
```
sbinet: `s/close/Close/`

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+// ErrorHandler is the function which handles occurred error (e.g. logs it).
+type ErrorHandler func(error)
+
+// NewFS returns a new path.FileSystem representing the filesystem on the remote XRootD server.
+// client is a client connected to the remote XRootD server.
+// root is the path to the remote directory to be used as a root directory.
+// handler is the function which handles occurred error (e.g. logs it).
+func NewFS(client *client.Client, root string, handler ErrorHandler) *FS {
```
sbinet: perhaps we could check here whether `handler` is `nil`.
in the case it's `nil`, we could add a no-op handler, so all the `fs.handle != nil` checks that are sprinkled in `FS` and `File` could be removed.

(we should also document this in the doc-string of `NewFS`.)

alternatively, we could go with the usual optional-func design:
```go
func NewFS(client *client.Client, root string, opts ...Option) *FS { ... }

type Option func(*FS)

func WithErrorHandler(f func (error)) Option {
    return func(fs *FS) {
         fs.handler = f
    }
}
```

xrootd/cmd/xrootd-fuse/main.go
```diff
@@ -0,0 +1,89 @@
...
+	nfs := pathfs.NewPathNodeFs(fs, nil)
+	server, _, err := nodefs.MountRoot(flag.Arg(1), nfs.Root(), &nodefs.Options{
+		Debug: *debug,
+	})
+	if err != nil {
+		fmt.Printf("Mount fail: %v\n", err)
+		os.Exit(1)
+	}
+
+	server.Serve()
```
sbinet: we should probably arrange things to properly unmount the filesystem when the program receives `^C`.
kind of what I did in the `rootio` FUSE command:
https://github.com/go-hep/hep/pull/223/files#diff-9884b867f20406412cddcbbf7413f557R19

EgorMatirov:
Thanks!

I am going to add more tests to that branch (you can take a look at already added tests). :)

EgorMatirov:
Ooops, looks like tests are broken (again) because of acident erase of `/tmp`.
I'll fill an issue about removing dependency on specific file structure and try to address that ASAP, ok? :-/

EgorMatirov:
Hm... Also, we can make files that are used for read / dirlist /stat tests read-only, that would prevent them from being deleted/changed.

Maybe that's better than changing these tests. Right now we test only "read" access in read/dirlist/stat tests IIRC and do not depend on working "create/write" requests, that are tested separately.

sbinet:
SGTM.

I'll log on the server and make the files owned by `root`, read-only so `gopher` (nor `gsoc-2018`) should be able to unadvertantly modify them.

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,261 @@
...
+	xrdmode := convertModeToXrdMode(mode)
+	err := fs.xrdfs.Chmod(context.Background(), path.Join(fs.root, name), xrdmode)
+	status := errorToStatus(err)
+	if status == fuse.EIO {
+		fs.handler(errors.WithMessage(err, "xrdfuse: error calling Chmod"))
+	}
+	return status
+}
+
+func entryStatToMode(stat xrdfs.EntryStat) uint32 {
```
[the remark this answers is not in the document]
EgorMatirov: Nope, it uses different flags (`os` ones).

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,261 @@
...
+	stat, err := fs.xrdfs.Stat(context.Background(), path.Join(fs.root, name))
+	status := errorToStatus(err)
+	if status == fuse.EIO {
+		fs.handler(errors.WithMessage(err, "xrdfuse: error calling Stat"))
+	}
+	if status != fuse.OK {
+		return nil, status
+	}
+
+	return &fuse.Attr{
```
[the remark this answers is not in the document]
EgorMatirov: `stat` returns `nil` in `Sys()` and `ToAttr` uses it for conversion, so it would not work for us.

xrootd/xrdfuse/xrdfuse.go
```diff
@@ -0,0 +1,261 @@
...
+	"github.com/hanwen/go-fuse/fuse"
+	"github.com/hanwen/go-fuse/fuse/nodefs"
+	"github.com/hanwen/go-fuse/fuse/pathfs"
+	"github.com/pkg/errors"
+	"go-hep.org/x/hep/xrootd/client"
+	"go-hep.org/x/hep/xrootd/xrdfs"
+	"go-hep.org/x/hep/xrootd/xrdproto"
+)
+
+type FS struct {
```
sbinet: please document this exported type.

xrootd/xrdfuse/xrdfuse.go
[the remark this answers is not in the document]
sbinet: right. (and I guess that even if we didn't return `nil`, it would be weird to get the `linux` representation of it on a `windows` system...)

xrootd/xrdfuse/xrdfuse.go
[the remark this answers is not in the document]
sbinet: ok. I thought the values were the same (so one would just have to do `uint32(stat.Mode())`)

sbinet:
thanks!

PR #296, opened 2018-07-23
Rootio write tobject

sbinet:


sbinet:
abandonned (split up in smaller PRs)

PR #300, opened 2018-07-24
xrootd: cleanup

EgorMatirov:
Cleanup tests and main code by extracting `SendResponse`, `ReadResponse`, and `ReadRequest` functions.
Add for the sake of perfomance and avoiding extra allocations `ReadResponseWithReuse` that takes `headerBytes` slice and pointer to `header` as arguments.
Extract helper methods from session.consume for the better readability.
Rename `*Code` constants in `protocol.go`

### DIFF
--- a/xrootd/client/session.go
+++ b/xrootd/client/session.go
@@ -6,7 +6,6 @@ package client // import "go-hep.org/x/hep/xrootd/client"
 
 import (
 	"context"
-	"encoding/binary"
 	"net"
 	"sync"
 	"sync/atomic"
@@ -187,10 +186,11 @@ func (sess *session) handleReadError(err error) {
 // after the number of seconds encoded in data.
 // See http://xrootd.org/doc/dev45/XRdv310.pdf, p. 35 for the specification of the response.
 func (sess *session) handleWaitResponse(streamID xrdproto.StreamID, data []byte) error {
-	if len(data) < 4 {
-		return errors.Errorf("xrootd: error decoding wait duration, want 4 bytes, got: %v", data)
+	var resp xrdproto.WaitResponse
+	rBuffer := xrdenc.NewRBuffer(data)
+	if err := resp.UnmarshalXrd(rBuffer); err != nil {
+		return err
 	}
-	duration := time.Duration(binary.BigEndian.Uint32(data)) * time.Second
 
 	sess.mu.RLock()
 	req, ok := sess.requests[streamID]
@@ -200,7 +200,7 @@ func (sess *session) handleWaitResponse(streamID xrdproto.StreamID, data []byte)
 	}
 
 	go func(req pendingRequest) {
-		time.Sleep(duration)
+		time.Sleep(resp.Duration)
 		if err := sess.writeRequest(req); err != nil {
 			resp := mux.ServerResponse{Err: errors.WithMessage(err, "xrootd: could not send data to the server")}
 			err := sess.mux.SendData(streamID, resp)
--- a/xrootd/client/session.go
+++ b/xrootd/client/session.go
@@ -7,7 +7,6 @@ package client // import "go-hep.org/x/hep/xrootd/client"
 import (
 	"context"
 	"encoding/binary"
-	"io"
 	"net"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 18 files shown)


xrootd/xrdproto/protocol.go
```diff
@@ -260,3 +262,86 @@ func Opaque(path string) string {
 	pos := strings.LastIndex(path, "?")
 	return path[pos+1:]
 }
+
+// ReadRequest reads a XRootD request from r.
+// ReadRequest returns entire payload of the request including header.
+// ReadRequest requires serialization since multiple ReadFull calls are made.
+func ReadRequest(r io.Reader) ([]byte, error) {
```
sbinet: we should probably add a test for this function

xrootd/xrdproto/protocol.go
```diff
@@ -260,3 +262,86 @@ func Opaque(path string) string {
...
+
+	return append(request, data...), nil
+}
+
+// WriteResponse writes a XRootD response resp to the w.
+// The response is directed to the stream with id equal to the streamID.
+// The status is sent as part of response header.
+// WriteResponse writes all data to the w as single Write call, so no
+// serialization is required.
+func WriteResponse(w io.Writer, streamID StreamID, status ResponseStatus, resp Marshaler) error {
```
sbinet: we should probably add a test for this function

xrootd/xrdproto/protocol.go
```diff
@@ -260,3 +262,86 @@ func Opaque(path string) string {
...
+	if _, err := w.Write(response); err != nil {
+		return err
+	}
+	return nil
+}
+
+// ReadResponse reads a XRootD response from r.
+// ReadResponse returns the response header and the response body.
+// ReadResponse requires serialization since multiple ReadFull calls are made.
+func ReadResponse(r io.Reader) (ResponseHeader, []byte, error) {
```
sbinet: we should probably add a test for this function

xrootd/xrdproto/protocol.go
```diff
@@ -322,26 +322,29 @@ func WriteResponse(w io.Writer, streamID StreamID, status ResponseStatus, resp M
...
+	var header ResponseHeader
+	data, err := ReadResponseWithReuse(r, make([]byte, ResponseHeaderLength), &header)
+	return header, data, err
+}
+
+// ReadResponseWithReuse reads a XRootD response from r. A response header is read into headerBytes and
+// unmarshaled to header for the reusing reasons.
+// ReadResponseWithReuse returns the response body.
+// ReadResponseWithReuse requires serialization since multiple ReadFull calls are made.
+func ReadResponseWithReuse(r io.Reader, headerBytes []byte, header *ResponseHeader) ([]byte, error) {
```
sbinet: it seems to me (but I may be mistaken) we could completely drop `ReadResponse` and replace it with this `ReadResponseWithReuse` (that we could then rename into `ReadResponse`).

we could also slightly modify its signature:
```go
func ReadResponse(r io.Reader, hdr *ResponseHeader) ([]byte, error) {
    var buf [ResponseHeaderLength]byte
    if _, err := io.ReadFull(r, buf[:]); err != nil { ... }
    ...
}
```

what do you think?

sbinet:
(but I like PRs that remove a bunch of lines of code while keeping the same amount of features :P)

xrootd/xrdproto/protocol.go
EgorMatirov: That adds an extra allocation of  for each `ReadResponse` call. 
`+8 B/op, +1 allocs/op` for `BenchmarkSession_Ping`.

To be honest, I'm not sure if that's so important: we are losing a bit in readability and acquire a bit in performance, or vice-versa.

I'll be happy to hear your opinion! :)

sbinet: nah... this can be tabled for later (if really needed.)
it was just a late night idea :)

xrootd/client/session.go
```diff
@@ -163,88 +162,91 @@ func (sess *session) Close() error {
 	return nil
 }
 
+// handleReadError handles an error encountered while reading and parsing a response.
+// If the current session is equal to the initial, the error is considered critical and handleReadError panics.
+// Otherwise, the current session is closed and all requests are redirected to the initial session.
+// See http://xrootd.org/doc/dev45/XRdv310.pdf, p. 11 for details.
+func (sess *session) handleReadError(err error) {
```
sbinet: according to the coverage report, it seems `handleReadError` isn't exercized in our tests.
could we fix this?

xrootd/client/session.go
```diff
@@ -163,88 +162,91 @@ func (sess *session) Close() error {
...
+		_ = err
+	}
+	sess.mu.RUnlock()
+	sess.Close()
+}
+
+// handleWaitResponse handles a "kXR_wait" response by re-issuing the request with streamID
+// after the number of seconds encoded in data.
+// See http://xrootd.org/doc/dev45/XRdv310.pdf, p. 35 for the specification of the response.
+func (sess *session) handleWaitResponse(streamID xrdproto.StreamID, data []byte) error {
```
sbinet: according to the coverage report, it seems `handleWaitResponse` isn't exercized in our tests.
could we fix this?

PR #316, opened 2018-07-28
rootio: implement ROOTMarshaler for TObjArray

sbinet:


sbinet:
@maloft now you can focus on implementing `ROOTMarshaler` for the streamers.
(don't forget to send the PR for `TObjString`, though.)

PR #322, opened 2018-07-30
rootio: ROOTMarshaler for the streamers 

maloft:


rootio/streamers.go
```diff
@@ -61,6 +61,20 @@ func (tsi *tstreamerInfo) Elements() []StreamerElement {
...
+func (tsi *tstreamerInfo) MarshalROOT(w *WBuffer) (int, error) {
+	if w.err != nil {
+		return 0, w.err
+	}
+	pos := w.Pos()
+	w.WriteVersion(tsi.rvers)
+	tsi.named.MarshalROOT(w)
+	w.WriteU32(tsi.chksum)
+	w.WriteI32(tsi.clsver)
+	w.WriteObjectAny(tsi.objarr)
```
sbinet: it's a bit more complicated than that.
`tsi.objarr.arr` is `nil`-ed in `UnmarshalROOT` as it's duplicate informations with the `tsi.elems` slice (which is itself a more convenient way to expose informations to users.)

if we don't re-create + populate the `tsi.objarr.arr` field with meaningful informations, the `read-write-read` round trip test will just fail.

PR #323, opened 2018-08-01
rootio: ROOTMarshaler for Streamers

maloft:


rootio/streamers.go
```diff
@@ -61,6 +61,25 @@ func (tsi *tstreamerInfo) Elements() []StreamerElement {
...
+		return 0, w.err
+	}
+	pos := w.Pos()
+
+	w.WriteVersion(tsi.rvers)
+	tsi.named.MarshalROOT(w)
+	w.WriteU32(tsi.chksum)
+	w.WriteI32(tsi.clsver)
+
+	tsi.objarr.arr = make([]Object, tsi.objarr.Len())
```
sbinet: there are 2 issues with this code:
- it just write `n` `nil` pointers where `n := tsi.objarr.Len()`. in the case at hand, as `tsi.objarr.arr` has been `nil`-ed at line 108 in `tstreamerInfo.UnmarshalROOT`, it will write 0 elements. what you need is the length of the `tsi.elems` slice. _ie_ the slice that contains the informations that have been passed from `tsi.objarr.arr`, in `tstreamerInfo.UnmarshalROOT`.
- you're writing `n` elements, instead of writing a container holding (among other things) `n` elements. from the point of view of ROOT, that's really not the same thing.

rootio/streamers.go
```diff
@@ -72,10 +72,8 @@ func (tsi *tstreamerInfo) MarshalROOT(w *WBuffer) (int, error) {
 	w.WriteU32(tsi.chksum)
 	w.WriteI32(tsi.clsver)
 
-	tsi.objarr.arr = make([]Object, tsi.objarr.Len())
-	for _, obj := range tsi.objarr.arr {
-		w.WriteObjectAny(obj)
-	}
+	tsi.objarr.arr = make([]Object, len(tsi.elems))
```
sbinet: yes. but then `tsi.objarr.arr` is just a slice with `n` elements of type `rootio.Object` and value "the zero value of `rootio.Object`" (which is `nil`.)

here:
```go
func (tsi *tstreamerInfo) MarshalROOT(w *WBuffer) (int, error) {
        if w.err != nil {
                return 0, w.err
        }

        pos := w.Pos()
        w.WriteVersion(tsi.rvers)
        tsi.named.MarshalROOT(w)
        w.WriteU32(tsi.chksum)
        w.WriteI32(tsi.clsver)

        if len(tsi.elems) > 0 {
                tsi.objarr.arr = make([]Object, len(tsi.elems))
                for i, v := range tsi.elems {
                        tsi.objarr.arr[i] = v
                }
        }
        w.WriteObjectAny(tsi.objarr)
        tsi.objarr.arr = nil

        return w.SetByteCount(pos, "TStreamerInfo")
}
```

maloft: I don't understand why is tsi.objarr.arr nil-ed again at the end of MarshalROOT ?

sbinet: this is to prevent a possible hysteresis effect, (marginally) reduce memory usage and make sure we always have a single point of truth (SPOT) for the streamer elements informations.

sbinet:
ok. now, what about `tstreamerElement` and `tstreamerBase` ? :)

maloft:
Working on tstreamerElement :)

maloft:
Concerning the unmarshalling of tstreamerElement, i couldn't understand the meaning if the if test : 
	if tse.etype == 11 && (tse.ename == "Bool_t" || tse.ename == "bool") {
		tse.etype = 18
	}
Why don't we have tse.etype = 18 from the beginning, since each type hs its own etype ?

rootio/streamers.go
```diff
@@ -166,6 +166,46 @@ func (tse *tstreamerElement) TypeName() string {
...
+	w.WriteI32(tse.arrdim)
+
+	if tse.rvers == 1 {
+		w.WriteStaticArrayI32(tse.maxidx[:])
+	} else {
+		w.WriteFastArrayI32(tse.maxidx[:])
+	}
+	w.WriteString(tse.ename)
+
+	if tse.etype == 18 {
```
sbinet: actually, this should be put before writing out `tse.etype`.
and we probably don't want to change `tse.etype` variable, just modify the value being written out. and only if the `tse.ename` is the "correct" one.

ie:
```go
etype := tse.etype
if etype == kBool && (tse.ename == "Bool_t" || tse.ename == "bool") {
    etype = kUChar
}
w.WriteI32(etype)
```

sbinet:
it's probably a historic artefact from ROOT, before there was proper boolean support.

rootio/streamers.go
maloft: alright ! 
Then the switch i did isn't totally correct, is it ?

sbinet:
the `switch` is fine.

rootio/streamers.go
sbinet: the `switch` is fine.

sbinet:
ok. looks good.
now, what about the remaining streamers?
all of them (except `tstreamerBasicType`) should be pretty easy, now that you've got the hang of it.

maloft:
Concerning the Unmarshaler of tstreamerSTL : 
if tss.vtype == kSTLmultimap || tss.vtype == kSTLset {
		switch {
		case strings.HasPrefix(tss.tstreamerElement.ename, "std::set") || strings.HasPrefix(tss.tstreamerElement.ename, "set"):
			tss.vtype = kSTLset
		case strings.HasPrefix(tss.tstreamerElement.ename, "std::multimap") || strings.HasPrefix(tss.tstreamerElement.ename, "multimap"):
			tss.vtype = kSTLmultimap
		}
	}

i don't see the purpose of this portion of code ?

sbinet:
IIRC there was a big in some ROOT versions were the vtype value didn't correspond to the container name.
I think.

rootio/streamers.go
```diff
@@ -224,6 +303,33 @@ func (tsb *tstreamerBasicType) Class() string {
 	return "TStreamerBasicType"
 }
 
+func (tsb *tstreamerBasicType) MarshalROOT(w *WBuffer) (int, error) {
```
sbinet: Not completely sure that's the correct thing to do.
But the read/write buffer round-trip test will tell us.

Could you enable it for the steamers?
Thanks

maloft:
Please, i can't understand the error TRAVIS gives me : 
"This job was cancelled because the "Auto Cancellation" feature is currently enabled, and a more recent build (#1533) for pull request #323 came in while this job was waiting to be processed."

This is due to what exactly ?
thanks :)

sbinet:
It seems Travis is having one of those bad days...

sbinet:
ping?

sbinet:
Re-ping. (Pong?)
Please address my comment about the test that need to be enabled.

maloft:
Right away !  I thought the ping the other day was about starting the
report (because we were on the 7th of august) that's why i put my focus on
it and on that same day i sent you a first  part of the report.
My bad !

Le mer. 15 août 2018 à 13:02, Sebastien Binet <notifications@github.com> a
écrit :

maloft:
I put the test on the wbuffer_test.go file or on a new file ?

sbinet:
Look at how the tests are defined and run now for the r/w round-trip test.
It should be a one line change to enable it for the streamers.

rootio/wbuffer_test.go
```diff
@@ -579,9 +579,6 @@ func TestWBuffer_Write(t *testing.T) {
 func TestWriteWBuffer(t *testing.T) {
 	for _, test := range rwBufferCases {
 		t.Run("write-buffer="+test.file, func(t *testing.T) {
-			if test.skip {
```
sbinet: as this was the last test that had to be skipped, you can remove the `skip` field from the `rwBufferCases`.

maloft: Done

maloft:
For the commits that were not accepted by Travis (because he had a bad day the other day), what shall i do to them ?

sbinet:
The latest commits of today tested the code that had been produced then.
Nothing to do on your end.
(Although I maintain my comment about the tstreamerBasicType marshaling code being wrong. The current test does just not exercize this correctly. I'll produce a more thorough test tomorrow.)

sbinet:
cleaned up this PR as https://github.com/go-hep/hep/pull/331.
(also added a couple of marshalers for `TLeafXYZ` stuff.

PR #326, opened 2018-08-04
xrootd/server: move fshander from cmd\xrd-srv

EgorMatirov:


sbinet:
Looks good but I was also expecting to see most of the handler testing code be migrated to the 'server' package.

EgorMatirov:
My bad, sorry.

However, in that case we either need to import `client` from `server` (what feels a bit strange) or duplicate code from `client` part (there is a bit of the same duplication in `server_test.go` already).

Also, I think it would be nice to test `server` within `client` to maintain compatibility.

What do you think?

I can make some extra helper methods for test purposes and test `server` agains not `client` but these helper methods.

sbinet:
I don't think it's that strange to import `client` from the `fshandler_test.go` file.
if possible, this test file should be a `package server_test` but if we need to access internals of the `server` package, then, so be it (and use `package server`.)

> Also, I think it would be nice to test `server` within `client` to maintain compatibility.

yes.

it would also be nice to test:
- C++ server, Go client
- C++ client, Go server
- Go server, Go client

AFAICT, only the first item is extensively tested (that was our reference)
With the Go server nicely shaping up, the latter 2 items would be nice to extend and implement.

we should also probably create an issue with all the request/response pairs that need to be implemented for the server, like we did for the client. (that won't be part of the required GSoC portfolio, of course, but still very interesting to know where we stand.)

xrootd/server/fshandler.go
```diff
@@ -27,9 +26,9 @@ import (
 	"go-hep.org/x/hep/xrootd/xrdproto/xrdclose"
 )
 
-// handler implements server.Handler API by making request to the backing filesystem at basePath.
-type handler struct {
-	server.Handler
+// fshandler implements server.Handler API by making request to the backing filesystem at basePath.
+type fshandler struct {
+	Handler
```
sbinet: is embedding `Handler` really needed?
(well... this can be part of another PR.)

EgorMatirov: It allows to not bother about handling handshake / login / protocol requests here.

sbinet: ok.

PR #330, opened 2018-08-16
Writing file header

maloft:


sbinet:
this has been implemented as part of #336 and #338

PR #331, opened 2018-08-16
rootio: implement ROOTMarshaler for streamers+TLeaf

sbinet:


maloft:
Can't see how to tackle the ROOTMarshaler for tstreamerBasicType aside from what i did last time

PR #356, opened 2018-09-05
xrootd/cmd/xrd-ls: avoid empty printouts

pseyfert:
Old behaviour:
 * Print empty line after any directory listing.
   - This means that the full printout ends with an empty line
 * xrd-ls -l <file> with a file instead of a directory does not show
   filenames.

New behaviour:
 * Print one empty line between requested directories.
   - This means that no empty line is printed after the last of multiple
     directories.
 * Print one empty line before printing a subdirectory with -R
   - This means an empty line occurs only after listing a directory in
     the recursive mode when there are actually subdirectories
 * xrd-ls -l <file> shows the file's name after Mode, Size, and ModTime
   filenames.

TODO: the second commit is a workaround the fact that `fi, err := fs.Stat(ctx,
url.Path)` for me returns `fi` objects with empty `Name()`.

w/o `-l` this is not a problem because `path.Join(root, fi.Name())` is printed
and the entire path is in `root`. Also, when calling `xrd-ls` on a dir, the
`fi` objects for printing come from `fs.Dirlist()`.

### DIFF
--- a/xrootd/cmd/xrd-ls/main.go
+++ b/xrootd/cmd/xrd-ls/main.go
@@ -73,7 +73,11 @@ func main() {
 		log.Fatalf("missing directory operand")
 	}
 
-	for _, dir := range flag.Args() {
+	for i, dir := range flag.Args() {
+		if i > 0 {
+			// separate consecutive files by an empty line
+			fmt.Printf("\n")
+		}
 		err := xrdls(dir, *longFlag, *recFlag)
 		if err != nil {
 			log.Fatalf("could not list %q content: %v", dir, err)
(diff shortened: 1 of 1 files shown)


xrootd/cmd/xrd-ls/main.go
```diff
@@ -156,5 +163,11 @@ func format(o io.Writer, root string, fi os.FileInfo, long bool) {
 		return
 	}
 
-	fmt.Fprintf(o, "%v\t %d\t %s\t %s\n", fi.Mode(), fi.Size(), fi.ModTime().Format("Jan 02 15:04"), fi.Name())
+	var name string
```
sbinet: I would initialize name with the result of fi.Name, and then test whether 'name' is empty (and assign 'root' to it)

pseyfert: fine with me

sbinet:
thanks. I took the liberty to squash the last commit into its parent.
it's now: db74e7a.

PR #358, opened 2018-09-07
rootio: introduce Key.Type to infer key's payload's type

sbinet:
Key.Type() allows to infer a key's payload's type w/o having to actually
unmarshal the whole object from disk.

### DIFF
--- a/ci/run-tests.go
+++ b/ci/run-tests.go
@@ -46,7 +46,7 @@ func main() {
 	}
 	defer f.Close()
 
-	args := []string{"test"}
+	args := []string{"test", "-v"}
 
 	if *cover {
 		args = append(args, "-coverprofile=profile.out", "-covermode=atomic")
--- a/hbook/dist.go
+++ b/hbook/dist.go
@@ -56,9 +56,11 @@ func (d *Dist0D) scaleW(f float64) {
 
 // Dist1D is a 1-dim distribution.
 type Dist1D struct {
-	Dist   Dist0D  // weight moments
-	SumWX  float64 // 1st order weighted x moment
-	SumWX2 float64 // 2nd order weighted x moment
+	Dist  Dist0D // weight moments
+	Stats struct {
+		SumWX  float64 // 1st order weighted x moment
+		SumWX2 float64 // 2nd order weighted x moment
+	}
 }
 
 // Rank returns the number of dimensions of the distribution.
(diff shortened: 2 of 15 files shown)


sbinet:
@pseyfert PTAL.

Perhaps it should be called `ObjectType() reflect.Type`, instead...

pseyfert:
Yes, ObjectType is probably better to avoid confusion that it doesn't return
Key as type.

That said, I tried to use ObjectType like this:
https://github.com/pseyfert/go-list-root-trees/compare/huh?expand=1

And see a regression in benchmarks:

```
with string comparisons
>>> go test -cpuprofile cpu.prof -memprofile mem.prof -bench . -count 5
BenchmarkSmallCLI-4   	    5000	    315088 ns/op
BenchmarkSmallCLI-4   	    5000	    328147 ns/op
BenchmarkSmallCLI-4   	    5000	    321479 ns/op
BenchmarkSmallCLI-4   	    5000	    322361 ns/op
BenchmarkSmallCLI-4   	    5000	    332037 ns/op
BenchmarkBigCLI-4     	    1000	   1530561 ns/op
BenchmarkBigCLI-4     	    1000	   1522950 ns/op
BenchmarkBigCLI-4     	    1000	   1607886 ns/op
BenchmarkBigCLI-4     	    1000	   1611037 ns/op
BenchmarkBigCLI-4     	    1000	   1618161 ns/op
```

```
>>> go test -cpuprofile cpu.prof -memprofile mem.prof -bench . -count 5
with ObjectType
BenchmarkSmallCLI-4   	    5000	    312847 ns/op
BenchmarkSmallCLI-4   	    5000	    329072 ns/op
BenchmarkSmallCLI-4   	    5000	    326829 ns/op
BenchmarkSmallCLI-4   	    5000	    334198 ns/op
BenchmarkSmallCLI-4   	    5000	    330072 ns/op
BenchmarkBigCLI-4     	     500	   2467448 ns/op
BenchmarkBigCLI-4     	     500	   2425339 ns/op
BenchmarkBigCLI-4     	     500	   2460256 ns/op
BenchmarkBigCLI-4     	     500	   2520564 ns/op
BenchmarkBigCLI-4     	     500	   2464305 ns/op
```

sbinet:
Indeed. Well, you're trading a couple of string comparisons with a function call that uses reflection (and a map lookup) + the creation of a zero value (through reflect.)
And then, a type-switch :)

So... The new approach is probably more robust, but it can't beat a string comparison.
(Also it may actually be less robust right now because I don't think I have added support for all the different kinds of TTrees beside tchain and TTree. Not sure about tntuple.)

I'll run the profiler to see which part is the slowest and assess whether that can be addressed.

pseyfert:
okay. i think to some extent the reflect solution is still desirable as it can catch typos "tree" "Tree" "TTree" "Ttree" at compiler level, and once there is inheritance checking, the inheritance needs to be right only once on library level and users can just check one type.

sbinet:
ok. merging this in as it's useful :)
(BTW, I think `root-ls` could get cleverer and do what you do in `list-root-trees`. _ie_ I would probably merge a PR along these lines :P)

Issue #355, opened 2018-09-05
rootio: support for krb5 authentication at CERN


sbinet:
this was reported by @pseyfert on the mailing list:
- https://groups.google.com/forum/#!topic/go-hep/PtTAzlHNFWk
in a nutshell, the `krb5` authentication was shunted because the cached krb5 ticket wasn't correctly parsed (because of CERN's `/etc/krb5.conf` realms section that contains subsections not supported by `gokrb5`.)

as promised, I've raised this upstream:
- https://github.com/jcmturner/gokrb5/issues/196

PR #363, opened 2018-09-10
 xrootd/xrdproto/auth/krb5: handle unsupported directive parsing errors

sbinet:
This CL leverages gokrb5.v6 to properly handle unsupported directive
errors that stem from e.g. krb5.conf sections like:

```
[realms]
 CERN.CH  = {
  default_domain = cern.ch
  kpasswd_server = cerndc.cern.ch
  admin_server = cerndc.cern.ch
   kdc = cerndc.cern.ch
    v4_name_convert = {
     host = {
        rcmd = host
     }
   }
 }
```

the v4 parts are not supported by gokrb5.
this used to be fatal (in the sense that xrootd couldn't generate a
proper authentication token) but didn't need to be.

now, gokrb5.v6 returns this new config.UnsupportedDirective error that
we recognize and "handle" (by ignoring it.)

Fixes go-hep/hep#355.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,8 +34,8 @@ require (
 	gonum.org/v1/plot v0.0.0-20180905080458-5f3c436ce602
 	gopkg.in/jcmturner/aescts.v1 v1.0.1 // indirect
 	gopkg.in/jcmturner/dnsutils.v1 v1.0.1 // indirect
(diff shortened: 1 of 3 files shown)


sbinet:
@pseyfert this should resolve the lxplus issue.

pseyfert:
yup. looks good.

PR #372, opened 2018-10-04
xrootd/xrdproto/prepare: implement kXR_prepare request

sbinet:
Update go-hep/hep#170.

### DIFF
--- a/xrootd/xrdproto/prepare/prepare.go
+++ b/xrootd/xrdproto/prepare/prepare.go
@@ -1,0 +1,106 @@
+// Copyright 2018 The go-hep Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package prepare contains the types related to the prepare request.
+// See xrootd protocol specification (http://xrootd.org/doc/dev45/XRdv310.pdf, p. 69) for more details.
+package prepare // import "go-hep.org/x/hep/xrootd/xrdproto/prepare"
+
+import (
+	"strings"
+
+	"go-hep.org/x/hep/xrootd/internal/xrdenc"
+)
+
+// Prepare request options.
+const (
+	Cancel   = 1  // Cancel will cancel a prepare request.
+	Notify   = 2  // Notify will send a message when the file has been processed.
+	NoErrors = 4  // NoErrors will not send a notification for preparation errors.
+	Stage    = 8  // Stage will stage the file to disk if it is not online.
+	Write    = 16 // Write will prepare the file with write access.
+	Colocate = 32 // Colocate will co-locate the staged files, if at all possible.
+	Refresh  = 64 // Refresh will refresh the file access time even when location is known.
+)
+
+// RequestID is the id of the request, it is sent as part of message.
+// See xrootd protocol specification for details: http://xrootd.org/doc/dev45/XRdv310.pdf, 2.3 Client Request Format.
+const RequestID uint16 = 3021
+
+// Request holds the prepare request parameters.
+type Request struct {
+	Options  byte   // Options is a set of flags that apply to each path.
+	Priority byte   // Priority the request will have. 0: lowest priority, 3: highest.
+	Port     uint16 // UDP port number to which a message is to be sent.
+	_        [12]byte
+	Paths    []string
+}
+
+// MarshalXrd implements xrdproto.Marshaler.
+func (req Request) MarshalXrd(w *xrdenc.WBuffer) error {
+	w.WriteU8(req.Options)
+	w.WriteU8(req.Priority)
+	w.WriteU16(req.Port)
+	w.Next(12)
+
+	var raw []byte
+	switch len(req.Paths) {
+	case 0:
+		// no-op
+	default:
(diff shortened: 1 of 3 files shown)


sbinet:
first stab.

needs tests, obviously.

xrootd/xrdproto/prepare/prepare.go
```diff
@@ -0,0 +1,93 @@
...
+// UnmarshalXrd implements xrdproto.Unmarshaler.
+func (req *Request) UnmarshalXrd(r *xrdenc.RBuffer) error {
+	req.Options = r.ReadU8()
+	req.Priority = r.ReadU8()
+	req.Port = r.ReadU16()
+	r.Skip(12)
+	n := r.ReadI32()
+	raw := make([]byte, n)
+	r.ReadBytes(raw)
+	req.Paths = strings.Split(string(raw), "\n")
```
EgorMatirov: Sorry, I haven't checked the specification yet, but that will end up with last element being an empty string if '\n' is present after each path (like it's done in MarshalXrd).
See https://play.golang.org/p/QSI5IqX9p9y.

EgorMatirov: > is the list of new-line separated paths that are to be prepared for access. If
> only one path is supplied, it need not be terminated with a new line
> character (\n). 

Looks like Unmarshal is correct and Marshal doesn't fully follow specification (I suppose that there is no need for the "\n" of `raw` string).

sbinet: you're right. thanks!

xrootd/xrdproto/prepare/prepare.go
```diff
@@ -0,0 +1,93 @@
...
+type Response struct {
+	Data []byte
+}
+
+// RespID implements xrdproto.Response.RespID.
+func (resp *Response) RespID() uint16 { return RequestID }
+
+// MarshalXrd implements xrdproto.Marshaler.
+func (o Response) MarshalXrd(w *xrdenc.WBuffer) error {
+	w.WriteI32(int32(len(o.Data)))
```
EgorMatirov: The `rlen` from the response is a part of the header, actually. 
We need to write only o.Data here.
See `read` response for the example: https://github.com/go-hep/hep/blob/master/xrootd/xrdproto/read/read.go#L27

xrootd/xrdproto/prepare/prepare.go
```diff
@@ -0,0 +1,93 @@
...
+	w.WriteBytes(o.Data)
+	return nil
+}
+
+// UnmarshalXrd implements xrdproto.Unmarshaler.
+func (o *Response) UnmarshalXrd(r *xrdenc.RBuffer) error {
+	n := r.ReadI32()
+	o.Data = make([]byte, n)
+	r.ReadBytes(o.Data)
```
EgorMatirov: Same here, `n` is already exracted as part of header and `r.Len()` equals to it. See https://github.com/go-hep/hep/blob/master/xrootd/xrdproto/read/read.go#L33 for the example.

xrootd/xrdproto/prepare/prepare.go
```diff
@@ -0,0 +1,100 @@
...
+	case 1:
+		raw = []byte(req.Paths[0])
+	default:
+		for i, p := range req.Paths {
+			if i > 0 {
+				raw = append(raw, '\n')
+			}
+			raw = append(raw, []byte(p)...)
+		}
+	}
```
EgorMatirov: Something like:
```go
for i, p := range req.Paths {
    raw = append(raw, []byte(p)...)
    if i + 1 < len(req.Paths) {
        raw = append(raw, '\n')
    }
}
```
looks more compact and readable. However, I'm not sure that it's equal performance-wise in case of `len(req.Paths) == 1`.

The decision is up to you. :)

xrootd/xrdproto/prepare/prepare.go
```diff
@@ -0,0 +1,100 @@
+// Copyright 2018 The go-hep Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package prepare contains the types related to the prepare request.
+// See xrootd protocol specification (http://xrootd.org/doc/dev45/XRdv310.pdf, p. 106) for more details.
```
EgorMatirov: It's page 69 actually. (I actually used that link this time!) :)

sbinet: copy-pasta... :)

(apologies if my answers+PR-updates are a bit asynchronous and "out-of-band" with your helpful comments: I am in a train with really intermittent mobile data, and no wifi...)

sbinet: done

sbinet:
@beer4duke here is the beginning of the support for staging files from tapes :)

xrootd/xrdproto/prepare/prepare.go
sbinet: merged case-1 and default.
should be clearer now and almost as compact :)

Issue #399, opened 2018-11-19
xrootd/cmd/xrd-cp: sub-par performances


sbinet:
trying to copy the following file:
```
$>  xrd-ls -l root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleElectron.root
-r--r--r--	 1841440760	 Oct 16 16:39	 /eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleElectron.root
```

results in:

```
$> time xrd-cp root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleElectron.root go.root

real	27m49.436s
user	0m24.223s
sys	0m36.147s
```

while, with the C++ version, I got:

```
$>  time xrd-cp root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleElectron.root cxx.root

[1.715GB/1.715GB][100%][==================================================][43.9MB/s]   

real	0m40.105s
user	0m0.743s
sys	0m8.754s
```

presumably b/c of 2 factors:
- we haven't implemented `kXR_readv`
- we consequently don't write out the file in non-overlapping buckets

sbinet:
@EgorMatirov want to give this a try?

EgorMatirov:
> @EgorMatirov want to give this a try?

First thing, that I have noticed: `C++` version writes file by buckets of 16 MBytes while `Go` version uses buckets of 16 KBytes (due to https://golang.org/src/io/io.go#L391) which results in much bigger overhead.

Passing a buffer of 16MBytes results in:
`Go` version: `816s`.
`C++` version: `570s`.

Next optimization would be reading from the server and writing to the disk simultaneously.

Something like:
goroutine reads a bucket from the server and puts it to the buffered channel.
another goroutine reads a bucket from the channel and writes it to the disk.

I'll give it a try.

> we haven't implemented kXR_readv

To be honest, I don't see how that can speed up copying here. As far I can tell, the only difference is that it supports reading from several files, but that's unrelated here since we are copying one file. (But it looks like a good idea to check against copying several files later).

sbinet:
>> we haven't implemented kXR_readv
>
> To be honest, I don't see how that can speed up copying here. As far I can tell, the only difference is that it supports reading from several files, but that's unrelated here since we are copying one file. (But it looks like a good idea to check against copying several files later).

[...]

PR #401, opened 2018-11-20
xrootd/cmd/xrd-cp: use a 16MiB buffer for copying files

sbinet:
This CL uses a 16MiB buffer for copying (usually large) files.
We could also make the size available from the command line or try to
automatically decide what would be the best size using some heuristics.
This size turns out to be quite reasonnable.

Updates go-hep/hep#399.


sbinet:
thanks for the `io.CopyBuffer` idea @EgorMatirov :)

sbinet:
basically the same PR than earlier, except I've added a test and some benchmarks:

```
  $> benchstat ./ref.txt new.txt
  name            old time/op    new time/op     delta
  XrdCp_Small-8     52.3ms ± 2%     54.4ms ± 5%      +4.09%  (p=0.000 n=30+29)
  XrdCp_Medium-8     17.2s ±21%      2.5s ±156%     -85.41%  (p=0.000 n=25+26)

  name            old alloc/op   new alloc/op    delta
  XrdCp_Small-8     59.0kB ± 0%  16803.5kB ± 0%  +28371.67%  (p=0.000 n=28+25)
  XrdCp_Medium-8    86.1MB ± 0%    225.8MB ± 0%    +162.37%  (p=0.000 n=29+25)

  name            old allocs/op  new allocs/op   delta
  XrdCp_Small-8        241 ± 0%        242 ± 1%      +0.27%  (p=0.001 n=29+29)
  XrdCp_Medium-8     14.9k ± 0%       0.5k ± 2%     -96.66%  (p=0.000 n=29+28)
```

EgorMatirov:
>Thanks!
I think, I have improved performance a bit more. I'll double-check and make a PR with fresh benchmark results at evening.

> great!

I have tried something like (not totally correct in terms of error handling):
```go

func Copy(dst io.Writer, src io.Reader) (written int64, err error) {
	bucketsCh := make(chan []byte, 3)
	buckets := make([][]byte, 1)
	for i := 0; i<len(buckets); i++ {
		buckets[i] = make([]byte, 16 * 1024 * 1024)
	}
	go func() {
		currentBucket := 0
		for {
			nr, er := src.Read(buckets[currentBucket])
			if nr > 0 {
				bucketsCh <- buckets[currentBucket][:nr]
			}
			if er != nil {
				if er != io.EOF {
					err = er
				}
				break
			}
			currentBucket = (currentBucket + 1) % len(buckets)
		}
		if err != nil {
			log.Println(err)
		}
		close(bucketsCh)
	}()

	for bucket := range bucketsCh {
		nw, ew := dst.Write(bucket)
		if nw > 0 {
			written += int64(nw)
		}
		if ew != nil {
			err = ew
			break
		}
		if len(bucket) != nw {
			err = io.ErrShortWrite
			break
		}
	}

	return written, err
}
```

hoping that this will reduce wait for data being written to disk, but that results in almost 0% difference in time.

PR #404, opened 2018-11-21
groot/cmd/root-srv: prevent double-clicks on plot requests

sbinet:
also add a spinner.

Fixes #403.

PR #405, opened 2018-11-22
groot/{rsrv,cmd/root-srv}: implement a plot-request pipeline + per-request spinner

sbinet:
really fixes #404

### DIFF
--- a/groot/cmd/root-srv/data.go
+++ b/groot/cmd/root-srv/data.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"net/http"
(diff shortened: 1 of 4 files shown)


sbinet:
@vkuznet PTAL.

vkuznet:
The plots looks fine and multiple spinners are nice idea. I also checked that I
can click multiple times on the same attribute and it only produce one plot.

But now Open button does not do anything. I was only able
to load my root file via Upload button.

sbinet:
(I didn't notice it while testing b/c I am mostly using the `Enter` key to select a file from a list of files from history, instead of clicking on "Open". sorry about that.)

PR #421, opened 2018-12-20
groot: add support for TClonesArray

sbinet:


sbinet:
needs:

- support for member-wise streaming
- support for streamer-bypass

PR #429, opened 2019-01-17
xrootd: add ping, mkdir, rm, and rmdir support to the server

EgorMatirov:


xrootd/fshandler_test.go
```diff
@@ -767,3 +767,214 @@ func TestHandler_Rename(t *testing.T) {
...
+				t.Fatalf("could not create client: %v", err)
+			}
+			defer cli.Close()
+
+			if tc.mkdirAll {
+				err = cli.FS().MkdirAll(context.Background(), tc.path, xrdfs.OpenModeOwnerRead|xrdfs.OpenModeOwnerWrite|xrdfs.OpenModeOwnerExecute)
+			} else {
+				err = cli.FS().Mkdir(context.Background(), tc.path, xrdfs.OpenModeOwnerRead|xrdfs.OpenModeOwnerWrite|xrdfs.OpenModeOwnerExecute)
+			}
+			if err != nil {
```
sbinet: perhaps also add a test against the case where `err==nil` and `tc.errCode != 0`

xrootd/fshandler_test.go
```diff
@@ -767,3 +767,214 @@ func TestHandler_Rename(t *testing.T) {
...
+			}
+
+			cli, err := createClient(addr)
+			if err != nil {
+				t.Fatalf("could not create client: %v", err)
+			}
+			defer cli.Close()
+
+			err = cli.FS().RemoveFile(context.Background(), tc.path)
+			if err != nil {
```
sbinet: ditto.

xrootd/fshandler_test.go
```diff
@@ -767,3 +767,214 @@ func TestHandler_Rename(t *testing.T) {
...
+			}
+
+			cli, err := createClient(addr)
+			if err != nil {
+				t.Fatalf("could not create client: %v", err)
+			}
+			defer cli.Close()
+
+			err = cli.FS().RemoveDir(context.Background(), tc.path)
+			if err != nil {
```
sbinet: ditto.

sbinet:
nice PR! thanks.

we should probably add a compile test to make sure `fshandler` implements `xrootd.Handler`...
_e.g.:_
```go
var _ Handler = (*fshandler)(nil)
```

as the code stands now, `fshandler` doesn't implement `xrootd.Handler`: it's missing the new `Ping` method (if I am not mistaken)

EgorMatirov:
> we should probably add a compile test to make sure `fshandler` implements `xrootd.Handler`...
> _e.g.:_
> 
> ```go
> var _ Handler = (*fshandler)(nil)
> ```
> 
> as the code stands now, `fshandler` doesn't implement `xrootd.Handler`: it's missing the new `Ping` method (if I am not mistaken)

`fshandler` does implement `xrootd.Handler` since it embeds one. :)

It uses `xrootd.defaultHandler` that implements the "core" methods like the  `protocol`, `ping`, etc.

Also, it's already tested in:
```go
func NewFSHandler(basePath string) Handler {
	return &fshandler{
		Handler:  Default(),
		basePath: basePath,
		sessions: make(map[[16]byte]*srvSession),
	}
}
```

sbinet:
right... `fshandler` embeds `xrootd.Handler`...

then, could we add a test for `ping` :)

PR #430, opened 2019-01-17
groot: add sensible defaults for AttLine, Axis and H1

sbinet:
So a H1D created with groot can be correctly displayed by ROOT.


sbinet:
@tmitchel PTAL

PR #441, opened 2019-02-12
ci: use Athens GOPROXY in module mode

sbinet:


.travis.yml
```diff
@@ -32,6 +32,7 @@ matrix:
        - TAGS="-tags travis"
        - COVERAGE="-race"
        - GO111MODULE="on"
+       - GOPROXY="https://athens.azurefd.net"
```
jaybatra26: @sbinet, Why do we need a proxy server in travis CI?

sbinet: This isn't for Travis only, per se, but to test modules and the expected new package delivery system:
https://blog.golang.org/modules2019

jaybatra26: okay, Thanks. So, are these mirrors paid? Also, I just learned that these could be a victim to man in middle attack. What are your thoughts on this?

sbinet: athens is supposed to be a mirror operated by the community.
there will be bills to be paid, for sure, but I think the gophers won't have to give a credit card upfront.

https should address most of the MiM attacks (all?).
I am not a security expert, though.

sbinet:
superceded by https://github.com/go-hep/hep/pull/487

Issue #186, opened 2018-05-08
hbook/ntup: add example on how to create histos from CSV


sbinet:
something like this usecase, for example:

- https://root-forum.cern.ch/t/how-to-read-a-dat-file-and-plot-as-histogram/28916

```go
nt, err := ntcsv.Open("testdata/simple.csv", ntcsv.Comma(' '), ntcsv.Comment('#'))
if err != nil {
    log.Fatal(err)
}
defer nt.DB().Close()

h := hbook.NewH2D(100, 0, 100, 50, 0, 50)
_, err = nt.ScanH2D("var1, var2", h)
```

jaybatra26:
@sbinet . Can I take this one?

sbinet:
SGTM. I am sure @sebgadrat won't mind :)

sebgadrat:
@jaybatra26 Sure go ahead! :-)

jaybatra26:
Thanks @sebgadrat 

One quick question, where would this example file go. Would it be a standalone go file or is there a specific folder where this kind of examples reside?

sbinet:
usually, examples are put inside the directory holding the code for a given package.
we also try to use the Go way to create examples.

see:
- https://github.com/go-hep/hep/blob/master/hbook/h1d_test.go#L24

which then translates to:
- https://godoc.org/go-hep.org/x/hep/hbook#example-H1D

the nice thing with `ExampleXYZ` is that these examples are run as part of tests (with `go test`) as soon as they contain:
```go
// Output:
// something to test against.
```

so, in the case at hand, this would involve creating a new file, `ntup_example_test.go`, under `hep/hbook/ntup`:
```go
package ntup_test

func ExampleNtuple_scanH2D() {
    // ...
}
```

sbinet:
> Would this be a part of ReadMe?

no, examples in Go are usually picked up from GoDoc.

> I see similar examples in test files. Can use the csv's in test for examples?

yep :)

jaybatra26:
@sbinet Thanks a lot, This is really helpful. I have tried writing the example, but when I am trying to do `go build -o . ntup_example_test.go`. It says `no packages to build`. I would appreciate if you can help me understand why would this not work.

sbinet:
examples are built and run with `go test`.
_e.g._:
```
$> cd go-hep.org/x/hep/hbook/ntup
$> go test -run=Example -v
```
the above should build and run all tests and examples functions which name start with `Example`.

see:
- https://blog.golang.org/examples

PR #442, opened 2019-02-13
hbook/ntup: add example on how to create histos from CSV

jaybatra26:
This CL adds example to create histogram from a
CSV.

Fixes go-hep/hep#186.

Signed off by: Jay Batra <jaybatra73@gmail.com>

### DIFF
--- a/hbook/ntup/ntup_example_test.go
+++ b/hbook/ntup/ntup_example_test.go
@@ -1,0 +1,49 @@
+// Copyright 2019 The go-hep Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package ntup_test
+
+import (
+	"fmt"
+	"log"
+
+	"go-hep.org/x/hep/hbook/ntup/ntcsv"
+)
+
+func ExampleNtuple_scanH2D() {
+	nt, err := ntcsv.Open(
+		"ntcsv/testdata/simple-with-header.csv",
+		ntcsv.Comma(';'),
+		ntcsv.Header(),
+		ntcsv.Columns("v1", "v2", "v3"),
+	)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer nt.DB().Close()
+
+	h, err := nt.ScanH2D("v1, v2", nil)
+
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Printf("XMean:      %f\n", h.XMean())
+	fmt.Printf("YMean:      %f\n", h.YMean())
+	fmt.Printf("XRMS:       %f\n", h.XRMS())
+	fmt.Printf("YRMS:       %f\n", h.YRMS())
+	fmt.Printf("XStdDev:    %f\n", h.XStdDev())
+	fmt.Printf("YStdDev:    %f\n", h.YStdDev())
+	fmt.Printf("XStdErr:    %f\n", h.XStdErr())
+	fmt.Printf("YStdErr:    %f\n", h.YStdErr())
+
+	// Output:
+	// XMean:      4.500000
+	// YMean:      4.500000
+	// XRMS:       5.338539
+	// YRMS:       5.338539
+	// XStdDev:    3.027650
+	// YStdDev:    3.027650
+	// XStdErr:    0.957427
+	// YStdErr:    0.957427
+}


jaybatra26:
@sbinet. PTAL

hbook/ntup/ntup_test/ntup_example_test.go
```diff
@@ -0,0 +1,58 @@
+package ntup_test
```
sbinet: please add a copyright header.

hbook/ntup/ntup_test/ntup_example_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	// Output:
+	// mean:    0.004393321492241603
+	// rms:     1.0051696196105504
+	// std-dev: 1.0052102803319578
+	// std-err: 0.010052102803319578
+	
+}
+
+func main() {
```
sbinet: please remove the `main` function.
it's not needed.

hbook/ntup/ntup_test/ntup_example_test.go
```diff
@@ -0,0 +1,58 @@
+package ntup_test
+
+import (
```
sbinet: you are missing imports for the following packages:

- `golang.org/x/epx/rand`
- `gonum.org/v1/gonum/stat/distuv`
- `go-hep.org/x/hep/hbook`

and you have packages that are imported but not used:
- `log`, `reflect`, `testing`.

sbinet:
needs https://github.com/go-hep/license/pull/8

hbook/ntup/ntup_test/ntup_example_test.go
```diff
@@ -0,0 +1,55 @@
+// Copyright 2015 The go-hep Authors. All rights reserved.
```
sbinet: `s/2015/2019/`

hbook/ntup/ntup_test/ntup_example_test.go
```diff
@@ -0,0 +1,55 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package ntup_test
+
+import (
+	"fmt"
```
sbinet: we try to distinguish b/w imports from the standard library, and the 3rd party ones.
could you split the imports into 2 lists?

something like:
```go
import (
	"fmt"

	"go-hep.org/x/hep/hbook"
	"golang.org/x/exp/rand"
	"gonum.org/v1/gonum/stat/distuv"
)
```

also: no need to actually import `goimports`.
it's a command that you install once and then run over the code source:
```
$> go get golang.org/x/tools/cmd/goimports
```
and then:
```
$> cd go-hep.org/x/hep
$> goimports -w .
```
(but as I said earlier, it's best to hook it up with your favorite editor so it's run behind the scenes whenever you save the file or modify it.)

jaybatra26: @sbinet, sure. will try to hook it in the editor.

jaybatra26:
@sbinet. Done all the changes. PTAL :)

sbinet:
ah! I overlooked the fact the file is at the wrong place.
it should be under the same directory than the other `ntup` files:
```
$> git mv hbook/ntup/ntup_test/ntup_example_test.go hbook/ntup/ntup_example_test.go
```

this will correct the build failure:
- https://travis-ci.org/go-hep/hep/jobs/492730207#L1025
```
$ go install -v $TAGS ./...
go-hep.org/x/hep/hbook/ntup/ntup_test
go build go-hep.org/x/hep/hbook/ntup/ntup_test: no non-test Go files in /home/travis/gopath/src/go-hep.org/x/hep/hbook/ntup/ntup_test

The command "go install -v $TAGS ./..." exited with 1.
```

jaybatra26:
What does `go install -v $TAGS ./...` do?

jaybatra26:
> ah! I overlooked the fact the file is at the wrong place.
> it should be under the same directory than the other `ntup` files:
> 
> ```
> $> git mv hbook/ntup/ntup_test/ntup_example_test.go hbook/ntup/ntup_example_test.go
> ```

I have fixed this.

sbinet:
- `go get some/import/path/...` downloads the repository pointed by `some/import/path` and recursively downloads all its dependencies, as well as all the dependencies of the packages located "under" `some/import/path` (that's what the ellipsis `...` are for), and then compiles all these packages, and finally installs them under `$GOPATH/pkg/<os>_<arch>`

- `go install ./...` compiles and installs all the packages located under the current directory. it won't fetch dependencies and will fail if there are any missing.

the `$TAGS` environment variable is just an artifact of the CI system, used to test multiple settings (w/ or w/o the race detector enabled, w/ or w/o coverage analysis, ...)

jaybatra26:
@sbinet Thanks for the explanation, this was really helpful.

jaybatra26:
Removing the WIP tag.

jaybatra26:
@sbinet. It is still failing for  Go: 1.11.x. What can be done?

sbinet:
this test is a bit flaky... (I think I need to soften the timeout requirement)
I've relaunched it.

but, ok.

now that you've created the example and the basic support code, what about modifying that example to actually show how to create a 2-dim histogram from a CSV file, via the `ntup.Ntuple` type and its `ScanH2D` method ? :)

jaybatra26:
@sbinet Sure, I will certainly do it. :D

hbook/ntup/ntup_example_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"go-hep.org/x/hep/hbook"
+	
+)
+var (
+	nt *Ntuple
+)
+
+func ExampleNtuple_scanH2D() {
+	const nxpoints = 10000
+	nt, err := ntcsv.Open(
```
jaybatra26: @sbinet, I am not able to access `ntcsv` here. If I do `hbook.ntup.ntcsv.ntcsv.Open` It says `cannot refer to unexported name`. If I import `go-hep.org/x/hep/hbook/ntup/ntcsv` It throws `import cycle error`. Also, If I place this file inside `ntup/ntcsv`. I am not able to access both `hbook` and `ntup` at the same time(If I import `go-hep.org/x/hep/hbook/`, and try to access ntup, it throws `export error`. I am a little dicey about where this file should go. Can you please help me with that?

hbook/ntup/ntup_example_test.go
```diff
@@ -0,0 +1,71 @@
+// Copyright 2019 The go-hep Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package ntup
```
sbinet: please put examples into `_test` packages so the example code shows the exported API (and doesn't use internal bits.)
_i.e.:_
`s/package ntup/package ntup_test/`

hbook/ntup/ntup_example_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"fmt"
+	"log"
+	"go-hep.org/x/hep/hbook"
```
sbinet: please put an empty line before the `hbook` import to separate standard library imports from 3rd-party imports.
you'll probably also have to import `"go-hep.org/x/hep/hbook/ntup/ntcsv"` to be able to use it inside the example code.

hbook/ntup/ntup_example_test.go
```diff
@@ -0,0 +1,71 @@
...
+	
+)
+var (
+	nt *Ntuple
+)
+
+func ExampleNtuple_scanH2D() {
+	const nxpoints = 10000
```
sbinet: please remove.

hbook/ntup/ntup_example_test.go
```diff
@@ -0,0 +1,71 @@
...
+    }
+	
+
+	// Output:
+	// mean:    0.004393321492241603
+	// rms:     0.004393321492241603
+	// std-dev: 1.0052102803319578
+	// std-err: 0.010052102803319578
+
+}
```
sbinet: please add a final `\n`.

hbook/ntup/ntup_example_test.go
```diff
@@ -0,0 +1,71 @@
...
+package ntup
+
+import (
+	"fmt"
+	"log"
+	"go-hep.org/x/hep/hbook"
+	
+)
+var (
+	nt *Ntuple
```
sbinet: please remove.

hbook/ntup/ntup_example_test.go
sbinet: you have to import `"go-hep.org/x/hep/hbook/ntup/ntcsv"` before being able to use `ntcsv.Open`.
if you put this example in the `ntup_test` package (instead of the `ntup` package as is currently done), you'll be able to break the import cycle (`ntcsv` import `ntup`, so `ntup` can't import `ntcsv`.)

see my comment above, at line 5.

hbook/ntup/ntup_example_test.go
```diff
@@ -0,0 +1,48 @@
...
+		f float64
+		s string
+	}
+
+	h := hbook.NewH2D(10, 0, 10, 10, 0, 10)
+	got, err :=  nt.ScanH2D("v1,v2",h)
+	if err != nil {
+		log.Fatalf("%s: error scanning: %v", "v1, v2, v3 ", err)
+	}
+
```
jaybatra26: @sbinet, I am clueless what is to be done after this. Can you shed some light on it.

Please ignore any linter or style errors, once patch is complete, I will run the checks and correct accordingly.

Thanks in advance.

sbinet: according to the doc of `ntup.Ntuple.ScanH2D`:
```
$> go doc go-hep.org/x/hep/hbook/ntup.Ntuple.ScanH2D 
func (nt *Ntuple) ScanH2D(query string, h *hbook.H2D) (*hbook.H2D, error)
    ScanH2D executes a query against the ntuple and fills the histogram with the
    results of the query. If h is nil, a (100-bins, xmin, xmax+ULP) (100-bins,
    ymin, ymax+ULP) 2d-histogram is created, where xmin, xmax and ymin,ymax are
    inferred from the content of the underlying database.
```

so, I would drop the definition of `h` and just use the simpler form:
```go
h, err := nt.ScanH2D("v1, v2", nil)
```

test for the error (as you did) and then printout the mean, rms, stddev and std-err of `h`, for both X and Y (as `h` is a 2D distribution.)
see:
- https://github.com/go-hep/hep/blob/master/hbook/h2d_test.go#L78

hbook/ntup/ntup_example_test.go
```diff
@@ -0,0 +1,48 @@
...
+		ntcsv.Header(),
+		ntcsv.Columns("v1", "v2", "v3"),
+	)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	defer nt.DB().Close()
+
+	type dataType struct {
```
sbinet: please remove.

hbook/ntup/ntup_example_test.go
```diff
@@ -0,0 +1,48 @@
...
+		i int64
+		f float64
+		s string
+	}
+
+	h := hbook.NewH2D(10, 0, 10, 10, 0, 10)
+	got, err :=  nt.ScanH2D("v1,v2",h)
+	if err != nil {
+		log.Fatalf("%s: error scanning: %v", "v1, v2, v3 ", err)
```
sbinet: `s/v1, v2, v3/v1, v2/`

hbook/ntup/ntup_example_test.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"fmt"
+	"log"
+
+	"go-hep.org/x/hep/hbook/ntup/ntcsv"
+)
+
+func ExampleNtuple_scanH2D() {
+	nt, err := ntcsv.Open(
+		"testdata/simple-with-header.csv",
```
sbinet: you could use directly the file from `ntcsv/testdata`:

```go
nt, err := ntcsv.Open(
 	"ntcsv/testdata/simple-with-header.csv",
	ntcsv.Comma(';'),
	ntcsv.Header(),
	ntcsv.Columns("v1", "v2", "v3"),
)
```

so we don't duplicate too many files :)

jaybatra26: I am getting `open ntcsv/testdata/simple-with-header.csv: no such file or directory
exit status 1
`

jaybatra26: I think, because ntup_example_test.go is not in the ntup package. That's why this is happening.

sbinet: hum... worked for me.
also, from `ExampleNtuple_scanH2D` point of view, `testdata/...` isn't either in the `ntup` package :)

jaybatra26: @sbinet, I think ntcsv should be outside ntup, something like this 
 https://github.com/jaybatra26/hep/commit/49ade4a87ffee544cf626ed5c0e4becf0a000ad4

In this commit, I moved the example file to ntcsv. WDYT?

jaybatra26: umm, why didn't it work for me? Maybe because of  wrong `GOPATH` and `GOROOT` :(

Yep, I realized it as soon as I wrote :)

jaybatra26: My env variables are as follows. 
`GOPATH=/home/jbatra/go/
GOROOT=/usr/local/go
` 
and the project is in `/home/jbatra/go/src/go-hep.org/` which technically lies in `GOPATH`. ANy idea what could have gone wrong?

jaybatra26: Woooah!! It worked.

sbinet: cool. what did you change?
if I am not mistaken, one shouldn't set `$GOROOT` if it already points to the location `go` expects (IIRC, `/usr/local/go`)

hbook/ntup/ntup_example_test.go
```diff
@@ -0,0 +1,50 @@
...
+	nt, err := ntcsv.Open(
+		"ntcsv/testdata/simple-with-header.csv",
+		ntcsv.Comma(';'),
+		ntcsv.Header(),
+		ntcsv.Columns("v1", "v2", "v3"),
+	)
+	if err != nil {
+		log.Fatal(err)
+	}
+
```
sbinet: please remove this empty line.

hbook/ntup/ntup_example_test.go
```diff
@@ -0,0 +1,50 @@
...
+	defer nt.DB().Close()
+
+	got, err := nt.ScanH2D("v1, v2", nil)
```
sbinet: please rename `got` into `h` or `h2`.

hbook/ntup/ntup_example_test.go
jaybatra26: I changes nothing. Just reset the branch. Any way it worked :)

hbook/ntup/ntup_example_test.go
jaybatra26: @sbinet Done :)

PR #495, opened 2019-05-21
groot/rarrow: implement ROOT/Tree to Arrow Table conversion

sbinet:
Fixes go-hep/hep#493


gedw99:
Wow this is awesome. I was always wondering what would happen with Root and Arrow.
Got to start getting into this more now

PR #596, opened 2020-03-26
Groot rtree reader

sbinet:


groot/rtree/example_reader_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+import (
+	"fmt"
+	"log"
+
+	"go-hep.org/x/hep/groot"
+	"go-hep.org/x/hep/groot/rtree"
+)
+
+func ExampleReader() {
```
sbinet: @rmadar, what do you think about this new API to crawl through a tree?

rmadar: That's quite nice! I am just wondering if this is just to crawl the tree (like printing-based investigation) or if we could actually compute things? And if we can, could it be possible to change the type of what `func(ctx rtree.RCtx)` returns? We could think to create an histogram or a in a rather quick way (a bit a la numpy) ... But I don't know if this is doable (and if this is what you had in mind initially).

sbinet: this API allows that kind of thing.
```go
h1 := hbook.NewH1D(100, -10, 10)
v1 := 0.0
rvars := []rtree.ReadVar{{Name: "f64", Value:&v1}}

r, err := rtree.NewReader(...)
err = r.Read(func(ctx rtree.RCtx) error {
    h1.Fill(v1, 1.0)
    return nil
})
```

rmadar: Ha ok, yeah, that's definitely really nice: I would totally use it!

A side question: what the type `RCtx` does mean?

sbinet: read context. I probably/definitely should add some documentation to that type.

that said, there's also:
- https://godoc.org/go-hep.org/x/hep/groot/rsql#ScanH1D

and:
- https://godoc.org/go-hep.org/x/hep/hbook/ntup#example-Ntuple--ScanH1D
(together with this ROOT loader:
https://godoc.org/go-hep.org/x/hep/hbook/ntup/ntroot#example-Open)

rmadar: I see! Although I might not get all the features of the above links, I think the reader you propose here might be better because:
 + it can fill several histograms at once with possible short operations (maybe last link above `ntroot` loader can also do this)
 + apply various set of cuts and fill every histos for each of them (which is very good for event cut-flow for e.g.)

It seems that you can do small analysis with this reader, only with few lines of codes, I like it.

PR #626, opened 2020-04-08
hplot/huntex: first import

sbinet:


sbinet:
@rmadar
well, that's a start.
unfortunately, we'd need a few adjustements in `gonum/plot` to make this work correctly: strings are "measured" by `gonum/plot`'s machinery _before_ they are massaged by `huntex.Canvas`.
so the lengths differ before/after massaging and you get these off-aligned placements.

hplot/htex/parser.go
```diff
@@ -0,0 +1,11 @@
+// Copyright ©2020 The go-hep Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package htex
+
+import "io"
+
+type parser struct {
+	r io.Reader
```
golangcibot, line 10: `r` is unused (from `structcheck`)

hplot/htex/parser.go
```diff
@@ -0,0 +1,11 @@
+// Copyright ©2020 The go-hep Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package htex
+
+import "io"
+
+type parser struct {
```
golangcibot, line 9: `parser` is unused (from `deadcode`)

rmadar:
Honestly, I find it not so bad for a first pass. I will already help a lot - looking forward trying it!

PR #636, opened 2020-04-14
hplot: use y-aligned legend entries, update reference plots

sbinet:


sbinet:
@rmadar: this should close and handle all your issues (except for those related to LaTeX handling)

thanks for bringing them up :)

rmadar:
Excellent, thanks for reacting fast! I just tried out and it does solve everything I saw:

image

In the meanwhile, I have seen something else going-on with LaTeX: I'll open another issue (it has been a long time ;-) )

Issue #634, opened 2020-04-11
Suggestion for rtree: convert a string into a formula a la TTreeFormula


rmadar:
When building a small tool on `rtree` module, I realized that it would be very convenient to have an equivalent of [`TTreeForumula()` in ROOT](https://root.cern.ch/doc/master/classTTreeFormula.html) in order to do things like:
```go
// Declare a string expression
equation_str := "var1+var2"
f := rtree.TreeFormula(tree, equation_str)

// When looping over event, access the value, even if 'reader' 
// wasn't explicitly initialized with 'var1' and 'var2' branches
err = reader.Read(func(ctx rtree.RCtx) error {			
         x := f.Eval()
         return nil
      })
```

Would that be feasible? Would there be a performance cost compared to explicitly declare a reader with the variables `var1` and `var2` ?

sbinet:
it should be a relatively easy thing to implement, using _e.g._ the [go/scanner](https://godoc.org/go/scanner) and/or [go/parser](https://godoc.org/go/parser) packages (or even directly the [yaegi](https://godoc.org/github.com/containous/yaegi) interpreter or the [gomacro](https://github.com/cosmos72/gomacro) one).

rmadar:
Thanks for the update! I tried it and I like it. But ... :) there is something which is a bit annoying from the user side: the type assertion needed in the event loop. I am not sure if that can be shortcut, but it practically prevents the user of having *just a string* to declare in order to have the value.

I just based this statement on [this example](https://godoc.org/go-hep.org/x/hep/groot/rtree#ex-Reader--WithFormula), but maybe there are other syntax (?)

sbinet:
well... Go is a statically typed language, w/o generics (think, `template<class T> MyClass;`).
so the only way to be able to work with *any* type is to return `interface{}`.

one way around it is to wrap a closure around your `rtree.Formula`:

```go
r, err := rtree.NewReader(...)
form, err := r.Formula("float64(br1) + br2", nil)
eval := func() float64 { return form.Eval().(float64) }

var sum float64
err = r.Read(func(ctx rtree.RCtx) error {
    sum += eval()
    return nil
})
```

rmadar:
[...]

PR #639, opened 2020-04-15
groot/rtree: implement a Formula interpreter

sbinet:
Fixes go-hep/hep#634.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/apache/arrow/go/arrow v0.0.0-20200403134915-89ce1cadb678
 	github.com/astrogo/fitsio v0.1.0
 	github.com/campoy/embedmd v1.0.0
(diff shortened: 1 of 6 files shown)


sbinet:
@rmadar
would that suit you?

note that only expressions are allowed, _i.e.:_ this wouldn't be allowed:
```go
f, err := reader.Formula(`if br1 { return br2*42 } else { return br3 }`, nil)
```

rmadar:
Yeah that looks good! I didnt see examples in the commit, so i need try it in practice, but from what i understood from the test function, this should be all good!

sbinet:
there is an example :)

- https://github.com/go-hep/hep/pull/639/files#diff-6c7e0358952f3a4b4588a948f2605785R208

sbinet:
(not sure about the performances of this whole thing though... happy to get some feedback on that.)

Issue #644, opened 2020-04-16
groot/rtree: improve rtree.Formula performances


sbinet:
as noted in https://github.com/go-hep/hep/issues/634#issuecomment-614648197, we could shave off a few type-assertions by generating a function with a signature with the exact return type, by analyzing the provided expression.

this could probably be done by generating a little program through `go/types`:
```go
func f() {
    var v1 %T // filled by us
    var v2 %T
    ...
    var out = (%expr%)
    _ = out
    %typeof(out)%?
}
```

PR #646, opened 2020-04-16
groot/rtree: remove some type assertions in Formula generation

sbinet:
Fixes go-hep/hep#644.


sbinet:
@rmadar, could you try this one out and report whether this actually improves things for you?

steps to reproduce:
```
$> cd /some/where
$> go get go-hep.org/x/hep
$> cd /some/where/src/go-hep.org/x/hep
$> git remote add sbinet git@github.com:sbinet-hep/hep
$> git checkout -b issue-644 sbinet/issue-644
```
(and then point your `go.mod` to that location via a `replace` directive.)

thanks.

rmadar:
Hum when I do `go get go-hep.org/x/hep` I don't have any directory in my `some/where`.

I tried to clone directly the github 
```
git clone https://github.com/go-hep/hep.git
cd hep
git remote add sbinet git@github.com:sbinet-hep/hep
git checkout -b issue-644 sbinet/issue-644
```

And I get
```
fatal: 'sbinet/issue-644' is not a commit and a branch 'issue-644' cannot be created from it
```

So I don't know :)

sbinet:
sorry, forgot:
```
$> git fetch --all -p
```
and then:
```
$> git checkout -b issue-644 sbinet/issue-644
```

rmadar:
Ha ok, yes it worked! At least, my `go run main.go` lead to 
```
go: downloading github.com/containous/yaegi v0.8.2
```

**Without the modifications**
```
 Processing report:
    - 300 histograms filled over 3120 kEvts (5 samples, 15 variables, 4 selections)
    - running time: 34 ms/kEvt (00:01:45 for 3120 kEvts)
    - time fraction: 100% (event loop), 0% (plotting)
```

**With the modifications**
```
 Processing report:
    - 300 histograms filled over 3120 kEvts (5 samples, 15 variables, 4 selections)
    - running time: 30 ms/kEvt (00:01:32 for 3120 kEvts)
    - time fraction: 100% (event loop), 0% (plotting)
```

** Without formula**
```
 Processing report:
    - 300 histograms filled over 3120 kEvts (5 samples, 15 variables, 4 selections)
    - running time: 9 ms/kEvt (00:00:27 for 3120 kEvts)
    - time fraction: 99% (event loop), 1% (plotting)
```

So it is a bit better, but not too much. I let judge if this is reasonable or not :)

sbinet:
I'll call it a day for the moment :)
there's another possible improvement: slamming all formulae into a single one (behind the scenes), to pay the price of interpretation only once.

another option would be to require the user to provide a `*T` where to store the return value:

```go
r, err := rtree.NewReader(...)

// access result via 'v1' or form1.Eval().(float64)
var v1 float64
form1, err := r.Formula(&v1, "br1+br2", nil)

// access result via 'v2' or form2.Eval().(bool)
var v2 bool
form2, err := r.Formula(&v2, "br1 > 42", nil)

// infer type from expression.
// access result via form3.Eval().(float64)
form3, err := r.Formula(nil, "br1+br3", nil)
```

this removes another `reflect` call per event iteration.

what do you think?
(I'd delay option a) for later. I could do option b) earlier b/c of the API change -- that'd need to be merged fairly early on)

rmadar:
Yes, I think that's nice (especially to have the choice of doing it or not). I'd happy to try it :)

PR #648, opened 2020-04-17
hplot: add RatioPlot

sbinet:


sbinet:
@rmadar: what do you think about this kind of API?

sbinet:
note that in order to be able to seamlessly handle borders and latex compilation for all plot types (normal, tiled, ratio), I had to remove the `.Border` and `.Latex` fields of `hplot.Plot`.

the same can be done via a call to `hplot.Wrap` with the according `hplot.WithXYZ` plot option.

alternatively, (if the `hplot.Wrap` approach is felt too cumbersome), I could expose the "wrapper" plot type:

```go
type P struct {
    Plot   Drawer
    Border Border
    Latex  htex.Handler
}

p := hplot.New()
p.Title.Text = "title"

wp := hplot.P{Plot:p, Latex:htex.Handler}
wp.Border.Left = 10

err = wp.Save(-1, -1, "foo.tex")
```

rmadar:
I think it's quite nice, yes! 

I would maybe say than another user-friendly way to add (or not) a ratio plot could be the following (but not sure if this doable in practice, just tell me):

```go
p = hplot.New()
p.Add(h1, "Label")

p.SubPlot = hplot.New() // not sure you can have a field x in struct x ...?
p.SubPlot.Add(hratio, "Ratio")
p.SubPlot. ... // all cosmetic you want
```

sbinet:
hum... not sure I completely understand what is the expected behaviour of your approach.
could you clarify a bit?
thanks.

rmadar:
As a user, the best would be to simply say: "I want ratio plot on this one, with these histograms in it". So I was thinking:
 1. create the main plot (with axis, legend and everything)
 2. with a simple command, get the bottom plot with specified data (legend, axis etc ...)

```go
// --> Good looking plot without ratio at this stage
p := hplot.New()
applySomeStyle(p)
p.Add(hist1)
p.Add(hist2) 

// Add a ratio if wanted
if addRatio {
   p.RatioPlot.Add(h3)
   p.RatioPlot.Legend.Add("label", h3)
}
```

But that's maybe a bit too simplistic (e.g. size of plot w or w/o ratio, axis w or w/o etc ..) or focused for a general tool, and might have to be implemented at the user-level (?) I hope this, at least clarify it.

sbinet:
ok.

then it would be something like that:

```go
var plt hplot.Drawer

// --> Good looking plot without ratio at this stage
p := hplot.New()
applySomeStyle(p)
p.Add(hist1)
p.Add(hist2) 
plt = p

// Add a ratio if wanted
if addRatio {
   rp := hplot.NewRatio()
   rp.Top = p
   rp.Bottom.Add(h3)
   rp.Bottom.Legend.Add("label", h3)
   plt = rp
}

err = hplot.Save(plt, -1, -1, "foo.png")
```

rmadar:
PERFECT !!! Eager to try it !

sbinet:
cool.
note you'll probably have to modify a bit your `applySomeStyle` function because of the "slight" refactoring wrt borders, latex and DPI.

rmadar:
Yes, it seems relatively straighforward. Cool!

rmadar:
I am trying this now, and I was thinking about maybe re-naming the type `P` into something more suggestive for the user. Maybe `Figure` could be a nice one. Type Figure could be related to everything which gives the "final visual" (border & latex - for now). Well, that's just a suggestion.

sbinet:
I like `Figure`, thanks. (I was racking my brain for some nicer name... naming is tough)

PR #653, opened 2020-04-18
hplot: rename P+Wrap into Fig+Figure

sbinet:


sbinet:
@rmadar: thanks for the naming suggestion.

Issue #650, opened 2020-04-17
hep/sliceop/f64s: new package for operations on []float64


sbinet:
`gonum/floats.Filter` was rejected in https://github.com/gonum/gonum/issues/1310.

perhaps we could consider adding it to a new `f64s` package (tentative name):

```go
package f64s

func Filter(dst, src []float64, f func(v float64) bool) []float64 {}
func Map(dst, src []float64, f func(v float64) float64) []float64 {}
func Find(dst []int, src []float64, f func(v float64) bool) []int {}
```

ex:
```go
o := f64s.Filter(nil, src, fct)
o := f64s.Filter(make([]float64, len(src)), src, fct)

o := f64s.Map(nil, src, func(v float64) float64 { return v*2 })
o := f64s.Map(make([]float64, len(src)), src, func(v float64) float64 { return v*2 })
```

@rmadar, do you want to take a stab at it?

rmadar:
Yes, I am happy to implement this new package. I'll just need some instruction on how to proceed git-wise to add the corresponding file and ask for a merge request.

sbinet:
you'll need to "fork" `go-hep/hep` into your `rmadar` namespace (just click on the `"Fork"` button of this page, upper right corner).

then:
```
$> cd somewhere
$> git clone git@github.com:rmadar/hep && cd hep
$> git remote add upstream git@github.com:go-hep/hep
$> git fetch --all -p
$> git checkout -b issue-650 upstream/master
$> mkdir -p ./sliceop/f64s
$> touch ./sliceop/f64s/f64s.go
```

also, please read the [/contribute](https://go-hep.org/contributing/#code-contribution) document.

rmadar:
OK perfect, I did a first commit with the three functions. For the push, I am asked the following:
```
fatal: The upstream branch of your current branch does not match
the name of your current branch.  To push to the upstream branch
on the remote, use

    git push upstream HEAD:master

To push to the branch of the same name on the remote, use

    git push upstream issue-650
```

What is recommended when doing such developments?

sbinet:
you need to push to your fork's branch:

```
$> git push origin issue-650
```

this will then propose to send a pull request.
(yeah, github is a bit cumbersome with insisting to fork a repo to be able to send PR against it...)

sbinet:
https://godoc.org/go-hep.org/x/hep/sliceop/f64s

rmadar:
very cool!

PR #654, opened 2020-04-19
First version of the f64s package

rmadar:
I have implemented the three functions discussed in #650 with some tentatively clear comment for the doc. I don't know if we need to add examples for it.

### DIFF
--- a/sliceop/f64s/example_f64s_test.go
+++ b/sliceop/f64s/example_f64s_test.go
@@ -1,0 +1,39 @@
+// Copyright ©2020 The go-hep Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package f64s_test
+
+import (
+	"fmt"
+)
+
+// An example of slice filtering
+func ExampleFilter() {
+	slice := []float64{1, 2, -99, 4, 5, -99, 7}
+	condition := func(x float64) bool { return x > 0 }
+	fmt.Println(f64s.Filter(nil, slice, condition))
+
+	// Output:
+	// [1 2 4 5 7]
+}
+
+// An example of slice mapping
+func ExampleFind() {
+	slice := []float64{1, 2, -99, 4, 5, -99, 7}
+	operation := func(x float64) float64 { return x*x }
+	fmt.Println(f64s.Map(nil, slice, operation))
+
+	// Output:
+	// [1, 4, 9801, 16, 25, 9801, 49]
+}
+
+// An example of slice finding
+func ExampleFind() {
+	slice := []float64{1, 2, -99, 4, 5, -99, 7}
+	condition := func(x float64) bool { return x == -99 }
+	fmt.Println(f64s.Find(nil, slice, condition))
+
+	// Output:
+	// [2 5]
+}
--- a/sliceop/f64s/example_f64s_test.go
+++ b/sliceop/f64s/example_f64s_test.go
@@ -21,7 +21,7 @@ func ExampleFilter() {
 // An example of slice mapping
 func ExampleFind() {
 	slice := []float64{1, 2, -99, 4, 5, -99, 7}
-	operation := func(x float64) float64 { return x*x }
+	operation := func(x float64) float64 { return x * x }
 	fmt.Println(f64s.Map(nil, slice, operation))
 
 	// Output:
--- a/sliceop/f64s/example_f64s_test.go
+++ b/sliceop/f64s/example_f64s_test.go
@@ -27,7 +27,7 @@ func ExampleMap() {
 	fmt.Println(f64s.Map(nil, slice, operation))
 
 	// Output:
-	// [1, 4, 9801, 16, 25, 9801, 49]
+	// [1 4 9801 16 25 9801 49]
 }
 
 // An example of slice finding
--- a/sliceop/f64s/example_f64s_test.go
+++ b/sliceop/f64s/example_f64s_test.go
@@ -6,6 +6,8 @@ package f64s_test
 
 import (
 	"fmt"
+
+	"go-hep.org/x/hep/sliceop/f64s"
 )
 
 // An example of slice filtering
@@ -19,7 +21,7 @@ func ExampleFilter() {
 }
 
 // An example of slice mapping
-func ExampleFind() {
+func ExampleMap() {
 	slice := []float64{1, 2, -99, 4, 5, -99, 7}
 	operation := func(x float64) float64 { return x * x }
 	fmt.Println(f64s.Map(nil, slice, operation))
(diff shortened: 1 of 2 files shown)


sliceop/f64s/f64s.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	for _, x := range src {
+		if f(x) {
+			dst = append(dst, x)
+		}
+	}
+	return dst	
+}
+
+// Return a slice of {f(x_i)} when src = {x_i}
```
sbinet, line 22: ditto.

but I think I would add this requirement:
```
// Map will panic if the lengths of src and dst differ.
```

sliceop/f64s/f64s.go
```diff
@@ -0,0 +1,46 @@
...
+func Map(dst, src []float64, f func(v float64) float64) []float64 {
+	if dst == nil {
+		dst = []float64{}
```
sbinet, line 25:
```suggestion
		dst = make([]float64, len(src))
```

sliceop/f64s/f64s.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	for _, x := range src {
```
sbinet, line 27: add, before the loop:
```go
if len(src) != len(dst) {
    panic("f64s: length mismatch")
}
```

this should also help the compiler to automatically remove a few bound-check tests.

then, I'd rewrite the for-loop like so:
```go
for i, v := range src {
    dst[i] = f(v)
}
```

sliceop/f64s/f64s.go
```diff
@@ -0,0 +1,46 @@
...
+		dst = append(dst, f(x))
+	}
+	return dst
+}
+
+// Return a slice of all indices corresponding to elements for which f(x) is true
```
sbinet, line 33: ditto.

sliceop/f64s/f64s.go
```diff
@@ -0,0 +1,46 @@
...
+func Find(dst []int, src []float64, f func(v float64) bool) []int {
+	if dst == nil {
+		dst = []int{}
```
sbinet, line 36:
```suggestion
		dst = make([]int, 0, len(src))
```

sliceop/f64s/f64s.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	for i, x := range src {
```
sbinet, line 38: and here I'd add:
```go
dst = dst[:0]
```
(with a matching comment in the doc-string of `Find`.)

sbinet:
also:

- yes, it would be nice to provide a little example for those (also because it will be counted in the code coverage :P)
- the top-level `TestGoFmt` test (in the `hep` directory) is failing because the file isn't properly `gofmt`ed. please run:
 ```
 $> gofmt -w ./f64s.go
 ```

sliceop/f64s/f64s.go
rmadar: OK I am not sure to understand the logic of this one ... After some googling, I understand this is the clean the slice. But then, shouldn't we do this for the three functions (and *before* the loop)?

sbinet: yes, `dst = dst[:0]` will modify `dst` to have `len(dst)==0` while still keeping its underlying capacity unchanged.

and, yes, it should be put just before the loop.
we use it only for `Filter` and `Find`, because we don't know the final number of elements in `dst`.

but for `Map`, we know that `len(dst)==len(src)`, so we don't have to use `append` (just random access via `dst[i]`)

sliceop/f64s/f64s.go
```diff
@@ -0,0 +1,64 @@
+// Copyright ©2020 The go-hep Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package f64s provide common operations on float64 slices.
```
sbinet, line 5:
```suggestion
// Package f64s provides common operations on float64 slices.
```

sliceop/f64s/f64s.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Map creates a slice with all the elements f(x_i) where x_i are elements from src.
+// Map uses dst as work buffer, storing elements at the start of the slice.
+// Map allocates a new slice if dst is nil.
+// Map will panic if the lengths of src and dst differ.
+func Map(dst, src []float64, f func(v float64) float64) []float64 {
+
+	if dst == nil {
+		dst = make([]float64, 0, len(src))
```
sbinet, line 34:
```suggestion
		dst = make([]float64, len(src))
```
here we know the final size (and we index into it inside the loop)

sliceop/f64s/example_f64s_test.go
```diff
@@ -0,0 +1,39 @@
...
+	slice := []float64{1, 2, -99, 4, 5, -99, 7}
+	condition := func(x float64) bool { return x > 0 }
+	fmt.Println(f64s.Filter(nil, slice, condition))
+
+	// Output:
+	// [1 2 4 5 7]
+}
+
+// An example of slice mapping
+func ExampleFind() {
```
sbinet, line 22:
```suggestion
func ExampleMap() {
```

sliceop/f64s/example_f64s_test.go
```diff
@@ -0,0 +1,39 @@
+// Copyright ©2020 The go-hep Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package f64s_test
+
+import (
+	"fmt"
+)
```
sbinet, line 9: you're also missing an import of `"go-hep.org/x/hep/sliceop/f64s"`

```suggestion

	"go-hep.org/x/hep/sliceop/f64s"
)
```

sliceop/f64s/example_f64s_test.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// An example of slice mapping
+func ExampleMap() {
+	slice := []float64{1, 2, -99, 4, 5, -99, 7}
+	operation := func(x float64) float64 { return x * x }
+	fmt.Println(f64s.Map(nil, slice, operation))
+
+	// Output:
+	// [1, 4, 9801, 16, 25, 9801, 49]
```
sbinet, line 30:
```suggestion
	// [1 4 9801 16 25 9801 49]
```

sbinet:
(that would be against that [license](https://github.com/go-hep/license) repository, "obviously".)

rmadar:
ok, this is done: https://github.com/go-hep/license/pull/9

sbinet:
thanks, and welcome to the (HEP) world Gomination!

Issue #659, opened 2020-04-20
Possibly strange behaviour of S2D


rmadar:
I am using `hbook.DivideH1D(hData, hMC)` to create ratio plot and it seems to "freeze" the program once I add this obtained corresponding `hplot.S2D` to a `hplot.Plot`. The problem can (hopefully) be reproduced using https://github.com/rmadar/tree-gonalyzer/commit/b039fc00d21eae013cbebbc3a753b85ffcca4688 .

On my computer, the program doesn't end - it keeps running at least for a *very* long time - unless I comment the line https://github.com/rmadar/tree-gonalyzer/blob/master/ana/maker.go#L340
```go
 rp.Bottom.Add(s2d_ratio)
```

I don't know if I did something wrong or not. Sorry if this is a stupid mistake on my side.

sbinet:
running it and then hitting `^C` while it seemed stuck gave:

```
$> ./ana-show 
before division
after division
^\SIGQUIT: quit
PC=0x470ca1 m=0 sigcode=128

goroutine 0 [idle]:
runtime.futex(0x20130c8, 0x80, 0x0, 0x0, 0x7ffc00000000, 0x4415da, 0x20121f8, 0x100000002, 0x7ffcc4a9e648, 0x41000f, ...)
	/home/binet/sdk/go/src/runtime/sys_linux_amd64.s:567 +0x21
runtime.futexsleep(0x20130c8, 0x7ffc00000000, 0xffffffffffffffff)
	/home/binet/sdk/go/src/runtime/os_linux.go:45 +0x46
runtime.notesleep(0x20130c8)
	/home/binet/sdk/go/src/runtime/lock_futex.go:159 +0x9f
runtime.stopm()
	/home/binet/sdk/go/src/runtime/proc.go:1881 +0xc4
runtime.findrunnable(0xc000046000, 0x0)
	/home/binet/sdk/go/src/runtime/proc.go:2413 +0xa0b
runtime.schedule()
	/home/binet/sdk/go/src/runtime/proc.go:2613 +0x324
runtime.park_m(0xc000001e00)
	/home/binet/sdk/go/src/runtime/proc.go:2783 +0x9d
runtime.mcall(0x0)
	/home/binet/sdk/go/src/runtime/asm_amd64.s:318 +0x5b

goroutine 1 [runnable]:
[...]

PR #662, opened 2020-04-21
Option to discard NaNs in hbook.DivideH1D()

rmadar:
Following https://github.com/go-hep/hep/issues/659, and ooking at https://github.com/go-hep/hep/blob/a50ee1ad8cc2d7923eec6c476508f67758d52379/hbook/ops.go#L35                                               
                                                                                                                                                      
I think we could just implement this way, *i.e.* explicitly ask to ignore NaN points. This could be done simply by                                    
```go
func DivideH1D(num, den *H1D, ignoreNaN bool) (*S2D, error) {
   for i := range bins1 { 
      // ... 
      switch {                                                                                                                                        
                case b2h == 0 || (b1h == 0 && b1herr != 0): // TODO(sbinet): is it OK?                
                        y = math.NaN()
                        ey = math.NaN() 
                        if ignoreNane {
                                continue
                        }
                //...
      }
     //...
     s2d.Fill(Point2D{X: x, Y: y, ErrX: Range{Min: exm, Max: exp}, ErrY: Range{Min: ey, Max: ey}}) 
   }
}                      
``` 
I have implemented a first version (modifying the corresponding test and `../fads/cmd/fads-rivet-mc-generic/mcalg.go`.

### DIFF
--- a/hbook/ops.go
+++ b/hbook/ops.go
@@ -11,8 +11,13 @@ import (
 
 // DivideH1D divides 2 1D-histograms and returns a 2D scatter.
 // DivideH1D returns an error if the binning of the 1D histograms are not compatible.
-// DivideH1D ignores points with NaN values if ignoreNaN is true.
-func DivideH1D(num, den *H1D, ignoreNaN bool) (*S2D, error) {
+func DivideH1D(num, den *H1D, opts ...DivOptions) (*S2D, error) {
+
+	cfg := newDivConfig()
+	for _, opt := range opts {
+		opt(cfg)
+	}
+
 	var s2d S2D
 
 	bins1 := num.Binning.Bins
@@ -43,10 +48,12 @@ func DivideH1D(num, den *H1D, ignoreNaN bool) (*S2D, error) {
 
 		switch {
 		case b2h == 0 || (b1h == 0 && b1herr != 0): // TODO(sbinet): is it OK?
-			y = math.NaN()
-			ey = math.NaN()
-			if ignoreNaN {
+			if cfg.ignoreNaN {
 				continue
+			} else {
+				y = cfg.replaceNaN
+				ey = 0.0 // TODO(rmadar): I guess this is the most senstive case
+				// but another field could be added to divConfig
 			}
 		default:
 			y = b1h / b2h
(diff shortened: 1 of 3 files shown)


hbook/ops.go
```diff
@@ -11,7 +11,8 @@ import (
 
 // DivideH1D divides 2 1D-histograms and returns a 2D scatter.
 // DivideH1D returns an error if the binning of the 1D histograms are not compatible.
-func DivideH1D(num, den *H1D) (*S2D, error) {
+// DivideH1D ignores points with NaN values if ignoreNaN is true.
+func DivideH1D(num, den *H1D, ignoreNaN bool) (*S2D, error) {
```
sbinet, line 15: actually, I think I'd rather use the "functional option" pattern [1,2].

with having a default for `DivOptions` that matches the current behaviour (keeping the NaNs).
I could imagine the following definitions:
```go
func DivideH1D(num, den *H1D, opts ...DivOptions) (*S2D, error) {
    cfg := newDivConfig()
    for _, opt := range opts {
        opt(cfg)
    }

    // ... ignore if cfg.ignoreNaNs, use cfg.replace() otherwise ...
}

type DivOptions func(c *divConfig)
type divConfig struct {
    ignoreNaNs bool
    replace func() float64
}

func newDivConfig() *divConfig {
    return &divConfig{replace: func() float64 { return math.NaN() }}
}

func DivIgnoreNaNs() DivOptions {
    return func(c *divConfig) {
        c.ignoreNaNs = true
    }
}

func DivReplaceNaNs(v float64) DivOptions {
    return func(c *divConfig) {
        c.replace = func() float64{ return v }
    }
}
```

it's a fair amount of boilerplate code, granted.
but neat.

do you want to try ironing it out?

[1]: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis
[2]: https://godoc.org/go-hep.org/x/hep/hplot#FigOption

rmadar: OK, yes this is nice. I need to think about the logic (not too used to it). Let me give a shot and we'll see if I manage to do it.

hbook/ops.go
```diff
@@ -42,8 +48,13 @@ func DivideH1D(num, den *H1D) (*S2D, error) {
 
 		switch {
 		case b2h == 0 || (b1h == 0 && b1herr != 0): // TODO(sbinet): is it OK?
-			y = math.NaN()
-			ey = math.NaN()
+			if cfg.ignoreNaN {
+				continue
+			} else {
+				y = cfg.replaceNaN
+				ey = 0.0 // TODO(rmadar): I guess this is the most senstive case
```
sbinet, line 55:
```suggestion
				ey = 0.0 // TODO(rmadar): I guess this is the most sensitive case
```

also, please remove the TODO at line 41, above.
(as this PR is meant to fix it)

hbook/ops.go
```diff
@@ -67,6 +78,39 @@ func DivideH1D(num, den *H1D) (*S2D, error) {
 	return &s2d, nil
 }
 
+// DivOptions is type handling options of DivideH1D()
```
sbinet, line 81:
```suggestion
// DivOptions allows to customize the behaviour of DivideH1D.
```

hbook/ops.go
```diff
@@ -11,7 +11,13 @@ import (
 
 // DivideH1D divides 2 1D-histograms and returns a 2D scatter.
 // DivideH1D returns an error if the binning of the 1D histograms are not compatible.
-func DivideH1D(num, den *H1D) (*S2D, error) {
+func DivideH1D(num, den *H1D, opts ...DivOptions) (*S2D, error) {
```
sbinet, line 14: please also document the default behaviour, w/o any DivOptions.

hbook/ops.go
```diff
@@ -67,6 +78,39 @@ func DivideH1D(num, den *H1D) (*S2D, error) {
...
+	replaceNaN float64
+}
+
+// newDivConfig function builds the default configuration
+// for DivideH1D() option.
+func newDivConfig() *divConfig {
+	return &divConfig{replaceNaN: math.NaN()}
+}
+
+// DivIgnoreNaNs function returns an option of DivideH1D()
```
sbinet, line 98:
```suggestion
// DivIgnoreNaNs configures DivideH1D to ignore data points with NaNs.
```

hbook/ops.go
```diff
@@ -67,6 +78,39 @@ func DivideH1D(num, den *H1D) (*S2D, error) {
...
+// where NaN values are ignored.
+func DivIgnoreNaNs() DivOptions {
+	return func(c *divConfig) {
+		c.ignoreNaN = true
+	}
+}
+
+// DivReplaceNaNs function returns an option of DivideH1D()
```
sbinet, line 106:
```suggestion
// DivReplaceNaNs configures DivideH1D to replace NaN raised during divisions with the provided value.
```

Issue #671, opened 2020-04-23
f64s: add a function Take(indices[]int)?


rmadar:
I think having a function that return a sub-array corresponding to a slice of indices would be useful - especially in association with `floats.Find()`. This would imply to allocate, so I think it cannot be done in `gonum/floats` package. But we could do it in `f64s`. 

This would typically generalize easy filtering of a given slice, say `s1`, using another slice `s2`:
```go
filtered_slice := f64s.Take(s1, floats.Find(s2, cond))
```

What do you think?

sbinet:
SGTM.

it should probably have that kind of signature (to allow for user-defined allocations):
```go
func Take(dst, src []float64, indices []int) []float64 { ... }
```

PR #672, opened 2020-04-23
Adding Take() function

rmadar:
Fixes go-hep/hep#671.

### DIFF
--- a/sliceop/f64s/f64s.go
+++ b/sliceop/f64s/f64s.go
@@ -62,3 +62,24 @@ func Find(dst []int, src []float64, f func(v float64) bool) []int {
 
 	return dst
 }
+
+// Take creates a slice with all elements of src indiced by indices.
+// Find uses dst as work buffer, storing elements at the start of the slice.
+// Find clears dst if a slice is passed, and allocates a new slice if dst is nil.
+func Take(dst, src []float64, indices []int) []float64  {
+
+	if dst == nil {
+		dst = make([]float64, 0, len(indices))
+	}
+	
+	if len(indices) > len(src) {
+		panic("f64s: length mismatch")
+	}
+
+	dst = dst[:0]
+	for _, i := range indices {
+		dst = append(dst, src[i])
+	}
+
+	return dst
+}
--- a/sliceop/f64s/f64s.go
+++ b/sliceop/f64s/f64s.go
@@ -5,6 +5,10 @@
 // Package f64s provides common operations on float64 slices.
 package f64s
 
+import (
+	"sort"
+)
+
 // Filter creates a slice with all the elements x_i of src for which f(x_i) is true.
 // Filter uses dst as work buffer, storing elements at the start of the slice.
 // Filter clears dst if a slice is passed, and allocates a new slice if dst is nil.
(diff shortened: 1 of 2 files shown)


sliceop/f64s/f64s.go
```diff
@@ -62,3 +62,24 @@ func Find(dst []int, src []float64, f func(v float64) bool) []int {
 
 	return dst
 }
+
+// Take creates a slice with all elements of src indiced by the provided indices.
+// Take uses dst as work buffer, storing elements at the start of the slice.
+// Take clears dst if a slice is passed, and allocates a new slice if dst is nil.
+func Take(dst, src []float64, indices []int) []float64 {
```
sbinet, line 69: just thinking out loud, but:
- do we want - for performances reasons (well, I guess) - to require the indices to be sorted?
- what should we do about duplicate indices? is it worrisome? do we care? (we should probably document that as well)

sliceop/f64s/f64s.go
```diff
@@ -62,3 +62,24 @@ func Find(dst []int, src []float64, f func(v float64) bool) []int {
...
+
+	if dst == nil {
+		dst = make([]float64, 0, len(indices))
+	}
+
+	if len(indices) > len(src) {
+		panic("f64s: length mismatch")
+	}
+
+	dst = dst[:0]
```
sbinet, line 79:
```suggestion
	dst = dst[:len(indices)]
```
(we should also probably add a test on the length of `dst` before that line)

sliceop/f64s/f64s.go
```diff
@@ -66,8 +66,14 @@ func Find(dst []int, src []float64, f func(v float64) bool) []int {
 // Take creates a slice with all elements of src indiced by provided indices.
 // Take uses dst as work buffer, storing elements at the start of the slice.
 // Take clears dst if a slice is passed, and allocates a new slice if dst is nil.
+// Take will panic if length of indices is larger than lenght of src.
+// Take will panic if length of indices is different from lenght of dst.
 func Take(dst, src []float64, indices []int) []float64 {
 
+	// TO BE DONE (rmadar)
+	// - require that indices is sorted (for perf reasons) + doc
```
sbinet, line 74: for that, there's [sort.IntsAreSorted](https://godoc.org/sort#IntsAreSorted)

sliceop/f64s/f64s.go
rmadar: Concerning the duplicated indices, I think it's better to require no duplicate - to keep it simple. Otherwise, `Take()` would have to behave more like the equivalent `numpy` function, in which the returned array has the dimension of the indices (where - obviously - arbitrary size, dimension and duplicates are allowed for `indices`). I'll make it clear in the doc that Take returns a sub-slice of `src`.

sliceop/f64s/f64s.go
```diff
@@ -62,3 +66,42 @@ func Find(dst []int, src []float64, f func(v float64) bool) []int {
...
+	}
+
+	dst = dst[0:len(indices)]
+
+	m := make(map[int]int, len(indices))
+	for i, v := range indices {
+		if _, there := m[v]; !there {
+			m[v] = v
+		} else {
+			panic("f64s: ducplicate indices")
```
sbinet, line 101:
```suggestion
			panic("f64s: duplicate indices")
```

sliceop/f64s/f64s.go
```diff
@@ -62,3 +66,42 @@ func Find(dst []int, src []float64, f func(v float64) bool) []int {
...
+
+	if len(dst) != len(indices) {
+		panic("f64s: length mismatch")
+	}
+
+	dst = dst[0:len(indices)]
+
+	m := make(map[int]int, len(indices))
+	for i, v := range indices {
+		if _, there := m[v]; !there {
```
sbinet, line 98: I am not sure it wouldn't be faster to do this in 2 loops:
- one to fill the "set" of indices (we tend to use a `map[T]struct{}` for this: `struct{}` does not take up any memory),
- then to fill `dst`.

this can be part of a later PR (I'll play a bit with benchmarks)

rmadar: OK sounds good to me! I have replaced the `int` by en empty structure in the meanwhile.

PR #674, opened 2020-04-24
sliceop/f64s: improve performances of Take

sbinet:
Reduce memory allocation:
- from populating the map and
- from using sort.IntsAreSorted.

also removed a loop over indices by removing sort.IntsAreSorted.

```
name                old time/op    new time/op    delta
Take/Len=2-4          63.5ns ± 3%     6.1ns ± 1%   -90.45%  (p=0.000 n=18+18)
Take/Len=4-4          82.7ns ± 2%     6.9ns ± 2%   -91.64%  (p=0.000 n=19+20)
Take/Len=8-4           116ns ± 2%       8ns ± 2%   -93.33%  (p=0.000 n=18+20)
Take/Len=128-4        4.03µs ± 1%    0.07µs ± 2%   -98.24%  (p=0.000 n=20+19)
Take/Len=1024-4       31.8µs ± 1%     0.5µs ± 3%   -98.44%  (p=0.000 n=17+19)
Take/Len=1048576-4    57.2ms ± 2%     3.1ms ± 4%   -94.61%  (p=0.000 n=19+20)

name                old alloc/op   new alloc/op   delta
Take/Len=2-4           32.0B ± 0%      0.0B       -100.00%  (p=0.000 n=20+20)
Take/Len=4-4           32.0B ± 0%      0.0B       -100.00%  (p=0.000 n=20+20)
Take/Len=8-4           32.0B ± 0%      0.0B       -100.00%  (p=0.000 n=20+20)
Take/Len=128-4        1.49kB ± 0%    0.00kB       -100.00%  (p=0.000 n=20+20)
Take/Len=1024-4       11.0kB ± 0%     0.0kB       -100.00%  (p=0.000 n=20+20)
Take/Len=1048576-4    11.2MB ± 0%     0.0MB       -100.00%  (p=0.000 n=18+20)

name                old allocs/op  new allocs/op  delta
Take/Len=2-4            1.00 ± 0%      0.00       -100.00%  (p=0.000 n=20+20)
Take/Len=4-4            1.00 ± 0%      0.00       -100.00%  (p=0.000 n=20+20)
Take/Len=8-4            1.00 ± 0%      0.00       -100.00%  (p=0.000 n=20+20)
Take/Len=128-4          3.00 ± 0%      0.00       -100.00%  (p=0.000 n=20+20)
Take/Len=1024-4         6.00 ± 0%      0.00       -100.00%  (p=0.000 n=20+20)
Take/Len=1048576-4      19.0 ± 0%       0.0       -100.00%  (p=0.000 n=20+20)
```

### DIFF
--- a/sliceop/f64s/f64s.go
+++ b/sliceop/f64s/f64s.go
@@ -7,7 +7,6 @@ package f64s
 
 import (
 	"fmt"
-	"sort"
 )
 
 var (
@@ -82,10 +81,6 @@ func Find(dst []int, src []float64, f func(v float64) bool) []int {
 // Take will panic if length of indices is different from length of dst.
 func Take(dst, src []float64, indices []int) []float64 {
 
-	if !sort.IntsAreSorted(indices) {
-		panic(errSortedIndices)
-	}
-
 	if len(indices) > len(src) {
 		panic(errLength)
 	}
@@ -98,15 +93,21 @@ func Take(dst, src []float64, indices []int) []float64 {
 		panic(errLength)
 	}
 
-	dst = dst[:len(indices)]
+	if len(indices) == 0 {
+		return dst
+	}
 
-	set := make(map[int]struct{}, len(indices))
-	for i, v := range indices {
-		if _, dup := set[v]; dup {
+	dst[0] = src[indices[0]]
+	for i := 1; i < len(indices); i++ {
+		v0 := indices[i-1]
+		v1 := indices[i]
+		switch {
+		case v0 == v1:
 			panic(errDuplicateIndices)
+		case v0 > v1:
+			panic(errSortedIndices)
 		}
-		set[v] = struct{}{}
-		dst[i] = src[v]
+		dst[i] = src[v1]
 	}
 
 	return dst
--- a/sliceop/f64s/f64s_test.go
+++ b/sliceop/f64s/f64s_test.go
@@ -116,7 +116,7 @@ var takeSink []float64
 
 func BenchmarkTake(b *testing.B) {
 	for _, size := range []int{2, 4, 8, 128, 1024, 1024 * 1024} {
-		b.Run(fmt.Sprintf("%d", size), func(b *testing.B) {
+		b.Run(fmt.Sprintf("Len=%d", size), func(b *testing.B) {
 			src := make([]float64, size)
 			ind := make([]int, 0, len(src))
 			rnd := rand.New(rand.NewSource(0))
@@ -127,10 +127,9 @@ func BenchmarkTake(b *testing.B) {
 				}
 			}
 			dst := make([]float64, len(ind))
-
 			b.ReportAllocs()
-			b.ResetTimer()
 
+			b.ResetTimer()
 			for i := 0; i < b.N; i++ {
 				takeSink = Take(dst, src, ind)
 			}


sbinet:
pretty happy with myself :)
@rmadar

rmadar:
great! the way to handle the duplicates and ordering is nice (and indeed removing two loops is quite cool!). I guess the tests are ok, but I have the impression you don't fill `src[index[0]]`, and then you miss the first element, no?

sbinet:
FYI, testing with this change:
```diff
diff --git a/sliceop/f64s/f64s.go b/sliceop/f64s/f64s.go
index 347e2f36..b45a9961 100644
--- a/sliceop/f64s/f64s.go
+++ b/sliceop/f64s/f64s.go
@@ -97,18 +97,19 @@ func Take(dst, src []float64, indices []int) []float64 {
                return dst
        }
 
-       dst[0] = src[indices[0]]
-       for i := 1; i < len(indices); i++ {
-               v0 := indices[i-1]
-               v1 := indices[i]
+       n := len(indices) - 1
+       for i := 0; i < n; i++ {
+               v0 := indices[i]
+               v1 := indices[i+1]
                switch {
                case v0 == v1:
                        panic(errDuplicateIndices)
                case v0 > v1:
                        panic(errSortedIndices)
                }
-               dst[i] = src[v1]
+               dst[i] = src[v0]
        }
+       dst[n] = src[indices[n]]
 
        return dst
 }
```
(compared to the proposed change in 7e55fec)

lead to this performance improvement:
```
$> benchstat new4.log new7.log
name                old time/op    new time/op    delta
Take/Len=2-4          6.06ns ± 1%    5.86ns ± 1%   -3.34%  (p=0.000 n=18+20)
Take/Len=4-4          6.92ns ± 2%    6.83ns ± 1%   -1.27%  (p=0.000 n=20+17)
Take/Len=8-4          7.76ns ± 2%    7.56ns ± 1%   -2.60%  (p=0.000 n=20+20)
Take/Len=128-4        71.1ns ± 2%    82.9ns ± 1%  +16.59%  (p=0.000 n=19+20)
Take/Len=1024-4        495ns ± 3%     587ns ± 2%  +18.74%  (p=0.000 n=19+19)
Take/Len=1048576-4    3.08ms ± 4%    1.95ms ± 2%  -36.76%  (p=0.000 n=20+18)

name                old alloc/op   new alloc/op   delta
Take/Len=2-4           0.00B          0.00B          ~     (all equal)
Take/Len=4-4           0.00B          0.00B          ~     (all equal)
Take/Len=8-4           0.00B          0.00B          ~     (all equal)
Take/Len=128-4         0.00B          0.00B          ~     (all equal)
Take/Len=1024-4        0.00B          0.00B          ~     (all equal)
Take/Len=1048576-4     0.00B          0.00B          ~     (all equal)

name                old allocs/op  new allocs/op  delta
Take/Len=2-4            0.00           0.00          ~     (all equal)
Take/Len=4-4            0.00           0.00          ~     (all equal)
Take/Len=8-4            0.00           0.00          ~     (all equal)
Take/Len=128-4          0.00           0.00          ~     (all equal)
Take/Len=1024-4         0.00           0.00          ~     (all equal)
Take/Len=1048576-4      0.00           0.00          ~     (all equal)
```

_ie:_ it worsened for size=[128,1024]. not sure why.
but I _expect_ those sizes to be the more common for HEP workloads.
so I'd be enclined to leave the perf-improvements for sizes=[2,4,8,1024*1024] on the table.

what do you think @rmadar?

rmadar:
ok let me re-phrase to make sure I understand:
 + option (1): we first fill `dst[0]` with `src[indices[0]]` and then do the loop starting from 1
 + option (2): we first do the loop up to `n` and then fill `dst[n]` with `src[indices[n]]` 

with:
 + option (1) is faster by ~ 100% wrt the very first implementation
 + option (2) is faster by ~15-20% wrt option (1) only for n = [128, 1024], but 30% worth for higher n (and actually similarish for n<128).

In that case, I would agree with you. Alternatively we could also use both calculation depending on n (with two internal functions which can be called in the proper case, ie n<1028 or n>1028 - I neglect the few % improvement for smaller sizes). But I don't know if this a "clean" option for an official software.

sbinet:
> [rephrasing]

yes. you got it.

and picking various algorithm strategies depending on the inputs is what is generally done in various libraries, but one does really want the benefits to outweight the added "sophistication".

I'll leave it like is (_ie:_ sticking with the "i=1->n-1" iteration") and adding a new issue linking to this PR.

Issue #669, opened 2020-04-22
Proposition to add a plotting style for hplot.S2D


rmadar:
Trying to play a bit with ratio plots and histogram ratio as `S2D`, I was wondering if it would be possible to add a histogram-like plotting style fo `S2D`. Indeed, as you can see on 

![image](https://user-images.githubusercontent.com/30081646/79964771-77981e80-848b-11ea-8746-bd970f3b43d5.png)

the ratio plot is not done "per bin" while this would be more suggestive from the visual point of view (and this is also how histogram ration are usually displayed). In practice: one line as a histogram and the band around it.

Thinking about it, I would see two possible options:
 1. have a function `func NewH1D(s S2D)` but the returned `H1D` wouldn't have the right statistical attributes, *and* we would loose the super nice band feature (which I think should be added to histo for uncertainty plotting actually - but this could be discussed in another issue).
 2. have a new functional option like `hplot.NewS2D(s, DrawLikeHist(true))` (we can think of a better name) which would artificially plot only horizontal lines (connecting `(x[i]-xerr[i], y[i])` and `x[i]+xerr[i], y[i])` points) and vertical lines (to connect `(x[i]+xerr[i], y[i])` and `(x[i+1]-xerr[i+1], y[i+1])` points).

sbinet:
item 2) SGTM.

rmadar:
OK. I can try have a look but probably need some help/guidance on how/where this implementation should be done exactly. Should we put this plotting behavior in `hplot.Add(s)` function or in `hplot.NewS2D()`?

sbinet:
the way I'd go with it (subject to be refined as the plot thickens or unfold, puns intended), would be the following:

- add a new functional option, say, `WithStepsKind(HiSteps|Steps|FSteps)` (like in gnuplot [steps](https://stackoverflow.com/questions/58697529/gnuplot-how-to-stop-histeps-from-terminating-to-zero-y-0-at-start-and-endp))
  https://github.com/go-hep/hep/blob/master/hplot/s2d.go#L102

- handle it in the `withBand()` method (probably `switch`-ing on the value of this newly added `steps` field value, between various ways of creating the `top` and `bot(tom)` "perimeter" lines)
  https://github.com/go-hep/hep/blob/master/hplot/s2d.go#L77

sbinet:
and introducing a new `type`:
```go
type StepsKind byte

const (
    NoSteps StepsKind = iota
    HiSteps
    Steps
    FSteps
[...]

PR #678, opened 2020-04-25
Adding histStep plotting style for S2D

rmadar:
This is the first pass. I wasn't able to test it because no png file is save and I run the example (I am not sure why). The behavior of the band has to be defined, which will be done in the next commit(s).

Fixes go-hep/hep#669.


hplot/example_s2d_test.go
```diff
@@ -123,3 +123,35 @@ func ExampleS2D_withBand() {
 		log.Fatal(err)
 	}
 }
+
+// ExampleS2D_withHiSteps draws some scatter points
+// with their error bars, using a step-like style
+func ExampleS2D_WithStepKind() {
```
sbinet, line 129:
```suggestion
func ExampleS2D_withStepKind() {
```

sbinet:
could you add the `testdata/s2d_steps.png` file (but as `testdata/s2d_steps_golden.png`) ?

and add to the `s2d_test.go` file:
```go
func TestScatter2DWithStepKind(t *testing.T) {
	checkPlot(cmpimg.CheckPlot)(ExampleS2D_withStepKind, t, "s2d_steps.png")
}
```

rmadar:
OK, I will for the name of the file. I did a small local code: it seems to work as expected :) 
image

hplot/s2d.go
```diff
@@ -141,10 +161,29 @@ func (pts *S2D) Plot(c draw.Canvas, plt *plot.Plot) {
...
+
+		// rmadar: a switch was suggested but I'd say a if seems more suitable
+		if pts.Steps == HiSteps {
+
+			xerr, ok := pts.Data.(plotter.XErrorer)
+			if !ok {
+				panic("s2d: cannot get X errors during HiSteps plotting")
+			}
+
+			data_step := plotter.XYs{}
```
sbinet, line 178: we don't like snake_case too much in Go.
`dsteps` ?
`dataSteps` ?

hplot/example_s2d_test.go
```diff
@@ -123,3 +123,35 @@ func ExampleS2D_withBand() {
 		log.Fatal(err)
 	}
 }
+
+// ExampleS2D_withHiSteps draws some scatter points
```
sbinet, line 127:
```suggestion
// ExampleS2D_withStepsKind draws some scatter points
```

hplot/example_s2d_test.go
```diff
@@ -123,3 +123,35 @@ func ExampleS2D_withBand() {
...
+	}
+	s2d := hbook.NewS2D(pts...)
+
+	p := hplot.New()
+	p.Title.Text = "Scatter-2D (with steps)"
+	p.X.Label.Text = "X"
+	p.Y.Label.Text = "Y"
+	p.Add(plotter.NewGrid())
+
+	var stepsKind hplot.StepsKind = 1
```
sbinet, line 144:
```suggestion
	var stepsKind = hplot.HiSteps
```
(or just pass that value directly to `hplot.WithStepsKind`, below.)

hplot/s2d.go
```diff
@@ -108,6 +126,10 @@ func NewS2D(data plotter.XYer, opts ...Options) *S2D {
 
 	cfg := newConfig(opts)
 
+	if cfg.steps > 0 {
```
sbinet, line 144: actually, that may not be necessary (the zero value being `NoSteps`...)

hplot/s2d.go
```diff
@@ -100,16 +103,28 @@ func (pts *S2D) withBand() error {
...
-			x, y := pts.Data.XY(i)
+		top = make(plotter.XYs, 2*pts.Data.Len())
+		bot = make(plotter.XYs, 2*pts.Data.Len())
+		xerr, ok := pts.Data.(plotter.XErrorer)
+		if !ok {
+			panic("s2d: cannot get X errors, required for HiSteps plotting")
+		}
+
+		// rmadar: the loop here doesn't really need a copy of XYs() but I am not sure
+		// how to handle indices in the loop if only i is defined.
```
sbinet, line 116: what do you mean?

hplot/s2d.go
```diff
@@ -100,16 +103,28 @@ func (pts *S2D) withBand() error {
...
 	case HiSteps:
-		for i := range top {
-			// WIP(rmadar): implement the proper band in case of step
-			//              this might actually involve 2xn points - need to think
-			x, y := pts.Data.XY(i)
+		top = make(plotter.XYs, 2*pts.Data.Len())
+		bot = make(plotter.XYs, 2*pts.Data.Len())
+		xerr, ok := pts.Data.(plotter.XErrorer)
+		if !ok {
+			panic("s2d: cannot get X errors, required for HiSteps plotting")
```
sbinet, line 112: perhaps this check could be performed earlier?
(in `NewS2D`)

hplot/s2d.go
rmadar: Let say I don't want to copy data, then I'd have to go with:
```go
for i := range top { // running from 0 to 2n - 1
                       idata: = f(i)  
                       x, y := pts.Data.XY(idata)   
                        xmin, xmax := xerr.XError(idata)
                        ymin, ymax := yerr.YError(idata)
                        top[i] = ...
                        bot[i] = ... 
                }
```
Actually, now I write it, I could probably use something like `f(i) = i // 2`

hplot/example_s2d_test.go
```diff
@@ -123,3 +123,67 @@ func ExampleS2D_withBand() {
...
+	s := hplot.NewS2D(s2d, hplot.WithStepsKind(stepsKind), hplot.WithYErrBars(true), hplot.WithBand(true))
+	s.GlyphStyle.Color = color.Black
+	s.GlyphStyle.Radius = vg.Points(4)
+	s.LineStyle.Width = 1
+	s.LineStyle.Dashes = plotutil.Dashes(2)
+
+	p.Add(s)
+
+	err := p.Save(10*vg.Centimeter, 10*vg.Centimeter, "testdata/s2d_steps_bands.png")
```
sbinet, line 185:
```suggestion
	err := p.Save(10*vg.Centimeter, 10*vg.Centimeter, "testdata/s2d_steps_band.png")
```

hplot/s2d.go
rmadar: Actually this is not strictly speaking need: the way the step are done will still work without X errors, but will give the same result as NoSteps. So I would tend to remove this `panic` ... What do you think?

Here is an example for those points:
```go
pts := []hbook.Point2D{                                                                            
          {X: 1, Y: 1, ErrY: hbook.Range{Min: 2, Max: 3}},                                           
          {X: 2, ErrX: hbook.Range{Min: 0.5, Max: 0.5}, Y: 2, ErrY: hbook.Range{Min: 5, Max: 2}},    
          {X: 3, Y: 3, ErrY: hbook.Range{Min: 2, Max: 2}},                                           
          {X: 4, ErrX: hbook.Range{Min: 0.5, Max: 0.5}, Y: 4, ErrY: hbook.Range{Min: 1.2, Max: 2}},  
}
```

image

sbinet: sure, no panics is better :)

hplot/s2d.go
```diff
@@ -82,18 +86,45 @@ func (pts *S2D) withBand() error {
...
+	case HiSteps:
+		top = make(plotter.XYs, 2*pts.Data.Len())
+		bot = make(plotter.XYs, 2*pts.Data.Len())
+		xerr := pts.Data.(plotter.XErrorer)
+		for i := range top {
+			idata := i / 2
+			x, y := pts.Data.XY(idata)
+			xmin, xmax := xerr.XError(idata)
+			ymin, ymax := yerr.YError(idata)
+			top[i].X = x - math.Abs(xmin)
```
sbinet, line 116: perhaps use a `switch` instead of setting those fields twice in the case of `i%2 != 0`?
_e.g.:_
```go
switch {
case i%2 != 0:
     top[i].X = x + math.Abs(xmax)
     // ...
default:
     top[i].X = x - math.Abs(xmin)
    // ...
}
```

Issue #685, opened 2020-04-26
Implement uncertainty band to hstack


rmadar:
Following the recent addition of `Band` to H1D https://github.com/go-hep/hep/issues/679, I realized it wasn't propagated to `HStack` type. The expected behavior could be the band of the total histogram when using `hplot.HStackOn` (also band on individual stacked histo could be possible as well, but less readable) and individual histograms band when `hplot.HStackOff`. 

I'll try to give a shot.

PR #686, opened 2020-04-26
Propagate band option to HStack

rmadar:
Fixes https://github.com/go-hep/hep/issues/685


rmadar:
At this stage, it works only for the `HStackOff` mode.

rmadar:
Thanks! Please wait a bit before merging to master, I'd like to make another check

rmadar:
Ok now it properly works when `HStackOff` and don't display anything when `HStackOn` (while it displayed individual histogram bands in https://github.com/go-hep/hep/pull/686/commits/bebc237d29fbcac7a5b450cb3e19ca1f854337e1)

Now, here is the problem for the band of the fully stacked histogram: we would need to plot the band of the total histogram, *i.e.* after a `hbook.Add(hTot, h)` running over all `hbook.H1D` of the `HStack`. But we don't have the `hbook` object a this stage. 

Would you have a suggestion? (note that it's far from being critical - we can sum the `hbook` at the user level - but it would be nice to have I think).

rmadar:
Maybe one possibility would be to change 
```go
func yoffs(i int, ys []float64, v float64) {
...
}
```
into 

```go
func yoffs(i int, ys, yerrs []float64, v, verr float64) {
...
}
```
to quadratic ally sum up the stat error too. I'll try.

sbinet:
actually, we do have access to the original/underlying `hbook.H1D` value:
- https://github.com/go-hep/hep/blob/master/hplot/h1d.go#L24

rmadar:
Thanks! I have manage to make it work as (I think) it should *i.e.*:
  + no stack: display individual histogram band, when exist
  + stack: don't display individual band (even if they exist) but display a band corresponding to the total histogram. This band is a field of the stack so that its syle can be tuned independently.
(I have updated the documentation accordingly)

There is one issue which remains: the plotting of the full band must be done between the fill and the line of the last histogram. I didn't find how to do this properly, so right now I plot the same band n times - which is not so nice.

hplot/hstack.go
```diff
@@ -222,6 +252,13 @@ func (hs *HStack) hplot(c draw.Canvas, p *plot.Plot, h *H1D, yoffs []float64, hs
 		c.FillPolygon(h.FillColor, c.ClipPolygonXY(poly))
 	}
 
+	// FIX-ME (rmadar): this is plot many times but it has to be here
```
sbinet, line 255: we could add an extra argument to `func (hs *HStack) hplot(...)` to indicate the position in the stack.
either a boolean to indicate last, or just the current index in the stack (we know the length via the `hs *HStack` receiver).

hplot/hstack.go
```diff
@@ -172,6 +197,11 @@ func (hs *HStack) hplot(c draw.Canvas, p *plot.Plot, h *H1D, yoffs []float64, hs
 		}
 	}
 
+	// Plot individual histo band when not stacked
+	if h.Band != nil && hsk == HStackOff {
```
sbinet, line 201: assuming we go with the `boolean/index` way from below, shouldn't we move this at the same place than for `HStackOff`?

rmadar: Actually, I think the one of `HStackOff` should move down (and the same for `hplot.H1D`). I think it's better to have the following order: fill, band and then line.

sbinet:
(the failure for `go-master` is coming from the yaegi interpreter...)

Issue #683, opened 2020-04-26
Legend improvements


rmadar:
Now we have the nice band feature for histogram (which actually don't show up in the legend right now) , I think we should revisit a bit the legend for `H1D`. I would propose something like this (let me now if this is feasible or not), at least for cases without `fill` and with `fill` but no markers.

| line  | fill  |  marker  | band  | legend thumb.|
|---|---|---|---|---|
|  1 | 0  | 0/1  | 0/1  |  sky-line w/ marker, w/ box for the band |
|  0/1 |  1 | 0  | 0  | box w/ or w/o a line around (not the current sky-line - that I know I have asked for) |

I am aware that cases like `fill+markers` would then look strange (a box with a line around and a marker in the middle), as well as `fill+band` which would be even not defined (since the same box would have to carry two colors). But these configuration shouldn't be selected in principle (at least I wouldn't use them).

sbinet:
it certainly is doable.
tentatively assigning it to you :)

rmadar:
Thinking about this again, I think we should also probably add a legend for a band (individually, I mean). So Edited the title of the issue and the different points to address would be:
 - [ ] add a thumbnailer for band 
 - [ ] add band in the thumbnail of `H1D` (not present right now)
 - [ ] tune the thumbnail style of `H1D` depending on the option (band, marker, fill, line, etc ...)

PR #689, opened 2020-05-03
Improving H1D legend style

rmadar:
Fixes the two last items of https://github.com/go-hep/hep/issues/683

This is a first pass, I'll improve all cases and add the corresponding demo plot.


rmadar:
What do you think? (I expect cases with both `fill` and `band` to not be used so much in practice)

image

Not that some test are failing because the change that moved the plotting of `h.Band` between the fill and the line done in https://github.com/go-hep/hep/pull/686 didn't propagate somehow (most likely, I forgot to push/pull something at some point).

sbinet:
there are also tests that are failing b/c their "golden" file needs to be updated (as the way the legend is drawn has changed)

to be honest, I feel like the legend starts to be a bit cramped with all these details.
but ok.
(the alternative is to give the user control about what is displayed in the legend, via a mechanism like ROOT's and its `SetOptStat`)

I guess we can live with that until somebody really wants finer control on their legend.

rmadar:
Ok, looking at the plot again, I tend to agree. Let me try to improve this.

rmadar:
Thinking a bit more about this, I don't think there is a way to define a proper thumbnail behavior in all cases. For instance in ROOT, one needs to specify which property to display in the thumbnail (fill, line, marker, etc ...). One way to have the same flexibility could be to add options in 
```go
p.Legend.Add("Label", h, 
              hplot.LegWithMarkers(true), 
              hplot.LegWithFill(true), 
              hplot.LegWithLine(true),
              hplot.LegWithBand(false),              
)
```
But I'd propose this for a futrur PR, as it's really a "fine-tuning" feature. For the time being, I will only consider the three first cases on the plot above and define a priority when several "incompatible" attributes are enabled together (in order to keep the legend simple and clear).

sbinet:
SGTM

sbinet:
please rebase off the latest `master` (`75db7043`).
this should also probably fix your other band issue, you seem to have branched `issue-683` off an old revision:
```
* 75db7043 (HEAD -> master, sbinet/master, origin/master, origin/HEAD) groot/rtree: improve performances of Formula
* 84104945 groot/rtree: migrate to cosmos72/gomacro as interpreter
* 484efca1 groot/rtree: add formula benchmarks
* 8b07e3fc hplot: add hstack+band example to gallery
* f4f61ec4 hplot: propagate band option to HStack
* 37b989af hplot: export x/y error bar plotters of S2D
| * f8bf678c (rmadar/issue-683) Preparing the priority for thumbnail plotting
| * c6302f92 A more complete version
| * f9530ab1 gofmt
| * 52e81912 Improving H1D legend style
|/  
| * 6e3a88d5 (rmadar/issue-685) Adding an example and a test
| * 3325c866 Move the band plotting of histo after the fill
| * 9be1203a Implementing comments, adding clean solution for total band plotting
| * ad0812b2 Updating the doc to describe the band behaviour
| * 6845ee40 Adding a first version that properly works in all cases
| * 36f76319 Plot individual band only when not stacking
| * bebc237d Passing gofmt
| * c25da068 Band option in HTstack
|/  
| * 47a52649 (rmadar/issue-684) Adding documentation
| * 40439366 Exporting XErrs and YErrs
|/  
* dc78d321 hplot: add S2D plots with steps and band to gallery
* d7f9e177 hplot: add HiSteps plotting style for S2D
```

sbinet:
superceded by https://github.com/go-hep/hep/pull/703

PR #703, opened 2020-05-10
Improving H1D legend style - 2

rmadar:
PR following https://github.com/go-hep/hep/pull/689 (fixes https://github.com/go-hep/hep/issues/683)


rmadar:
OK I have simplified the legend and gave priority according to the following rules:
 + with fill: boxline and markers but disregard band
 + w/o fill: skyline, band and markers

image

If you think the last one is too crowded, we can also remove the marker when the histo is filled (I don't have a strong opinion here - and this setup shouldn't be used I think).

I'll update the `*_golden.png` files once we settle on the legend choice.

sbinet:
it looks fine.
and we can always adapt a bit after some experience has been gained.

thanks.

Issue #732, opened 2020-05-26
hplot: bug in HStack band display


rmadar:
The logic of the `hplot.HStack` band plotting is not exactly correct. When `HStackON` is enabled, it should display the full band whatever the individual histogram options, while it doesn't.

I think, the following piece of code
```go
// Plot individual histo band when not stacked or total band
if h.Band != nil {
	switch hsk {
	case HStackOff:
		h.Band.Plot(c, p)
	case HStackOn:
		if ih == len(hs.hs)-1 {
			hs.Band.Plot(c, p)
		}
	}
}
```
should be changed into
```go
// Plot individual histo band when not stacked or total band
switch hsk {
   case HStackOff:
        if h.Band != nil {
		h.Band.Plot(c, p)
        }
   case HStackOn:
        if ih == len(hs.hs)-1 && hs.Band != nil {
		hs.Band.Plot(c, p)
	}
}
```

I'll submit a PR for this.

PR #733, opened 2020-05-26
Fix the ordering of band plotting for individual histos or total stack

rmadar:
Fixes https://github.com/go-hep/hep/issues/732


rmadar:
What's about this ?

image

sbinet:
SGTM :)

sbinet:
it seems like the golden file might not have been correctly updated.

PR #734, opened 2020-05-27
groot/cmd/root-gen-rfunc: first import

### LINKED ISSUE #729: groot/rtree/rfunc: consider using go/types to generate rfunc.Formula implementations
right now, we use a ad hoc type (`Func`) to gather facts about the `Formula` implementation to generate (input types, output types, arity, etc...)

we could in fact generate the code from a `go/types.Func` value.
we could also perhaps expose a little command, taking a `func` signature as input, to generate the `Formula` implementation.

_e.g.:_
```
$> gen-rfunc-formula -pkg mypkg 'func(int32, float32, string) float64'
```

or, directly pointing the tool at the function to be wrapped:
```
$> gen-rfunc-formula -func 'example.com/mypkg.MyFunc'
```

- [ ] use `go/types` to generate code
- [ ] export a generation command

sbinet:
This CL adds a new command to automatically generate the rfunc.Formula
implementation corresponding to a func signature (or a function in a
package (ex: math.Abs)).

This CL also drops the automagical registration of rfunc.Formula
implementations from the rfunc package (as it would prevent the linker
from applying dead code elimination).

Fixes #729.


rmadar:
Thanks a lot for this, this is really *super* useful! I would like to propose few signatures/types that I personally find useful (feel free to tell me if this is too specific). But before:
 1. I would like to know: is it somehow "expensive" to generate many pre-defined functions or not?
 2. It would be nice to have the list of implemented type/signatures in the doc. Right now, I see something like this:
image
But I think we might want to tune the naming convention to make appear input type/multiplicity and output type. Why not something like: `In<T1>Ar<N1><T2>Ar<N2>...Out<T1>Ar<N1><T2>Ar<N2>`. For `func(in32, float64, float64) bool`, it would look like `InI32Ar1F64Ar1F64Ar1OutB` ... ok maybe not so readable (it would be better with `_` in this case ;-) )!


### Wish-list and motivation

My strategy is to work only with `float64` (for values) and `bool` (for cuts) as output types, to avoid too much combination. Enable pre-defined functions for slices would be also nice.  Here is the (first) full wish-list, organized by goals:

**1. Other types to `float64` conversion**
  + `func(float32) float64`
  + `func(int) float64`
  + `func(int32) float64`
  + `func(int64) float64`

**2. "Multi-branches" cuts** (I guess it's fair to let the "mixed" signatures to the user)
  + `func(int) bool` + arrities
  + `func(int32) bool` + arrities
  + `func(float32) bool` + arrities
  + `func(float64) bool` + arrities

**3. Slice manipulations** (both for reducing operation and element operation)
  + `func([]<input>) []float64`
  + `func([]<input>) float64`
  + `func([]<input>, int) float64`
(where `<input>` can be `int`, `int32`, `int64`, `float32` or `float64`)

sbinet:
let's move the discussion over there:
https://github.com/go-hep/hep/issues/737

Issue #749, opened 2020-06-04
hplot/hstack: possible issue with LogY hstack


rmadar:
I am playing with `hplot.WithLogY()` option and it works fine for `hplot.H1D`. For `hplot.Hstack`, it seems there is something odd when `HStackKind = HStackOn`. I tried to look at the hstack code, but I didn't see anything.

What I will describe here can be reproduced runing `Example_withLogScale()`, defined [here](https://github.com/rmadar/tree-gonalyzer/blob/2ac65225345f64189d8fb906270058d70700a4d2/ana/example_test.go#L131). When stack mode is off  (*i.e.* `ana.WithHistoStack(false)`), I get this:
![image](https://user-images.githubusercontent.com/30081646/83811358-f357ce80-a6b9-11ea-9d16-313a40436282.png)
(so no pb, even with bins having 0 events)

When I enable stacking (*i.e.* `ana.WithHistoStack(true)`), I get this error (which appears at the command `hplot.Save(f, ...)`):
```
panic: Values must be greater than 0 for a log scale. [recovered]
	panic: Values must be greater than 0 for a log scale.
``` 
But when I keep stack enabled and I change the x-axis ranges from `[0, 3000]` to [350, 2000], it works:
![image](https://user-images.githubusercontent.com/30081646/83811674-87299a80-a6ba-11ea-8ee8-ed7b0d71bf68.png)

sbinet:
it would seem it's when plotting the `Band` that things go awry...

```
=== RUN   Example_withLogScale
--- FAIL: Example_withLogScale (0.40s)
panic: Values must be greater than 0 for a log scale. [recovered]
	panic: Values must be greater than 0 for a log scale.

goroutine 1 [running]:
testing.(*InternalExample).processRunResult(0xc0001efd28, 0x0, 0x0, 0x17f6a04c, 0xba9240, 0xe8d3b0, 0x20300000000000)
	/home/binet/sdk/go/src/testing/example.go:89 +0x648
testing.runExample.func2(0xbfae9c15d598747e, 0x71c4fd, 0x1478fc0, 0xc0000ac590, 0xc0000ac008, 0xc0000942a0, 0xc00023fd28, 0xc0001efd58)
	/home/binet/sdk/go/src/testing/run_example.go:58 +0x10d
panic(0xba9240, 0xe8d3b0)
	/home/binet/sdk/go/src/runtime/panic.go:969 +0x175
[...]

PR #750, opened 2020-06-05
hplot: properly handle hstacks with bands and log-y

sbinet:
Fixes go-hep/hep#749.


sbinet:
@rmadar this fixes your issue.
but uncovers a new one: notice the gray band over the empty bin, on the lower left tile plot (around `-6`)?

it would seem we may need to change how we draw `H1D.Band`.
instead of reusing `hplot.Band`, we may need to introduce a dedicated plotter (probably one that is just a bunch of polygons)

rmadar:
OK I noticed the problem around `-6`, indeed. You mean that error band of histograms would need to be a different type than `hplot.Band`, at least for the plotting side?

Since we are about modifying error band of histograms, I had a suggestion in mind for a while (I might open a new issue if that's tricky): would it be possible to have hatched areas instead of full colored area? This looks way nicer for such a band:
image

sbinet:
> OK I noticed the problem around -6, indeed. You mean that error band of histograms would need to be a different type than hplot.Band, at least for the plotting side?

yep.
`hplot.ErrBand` ?
with, at its core, a `[]plotter.Polygon`.

> Since we are about modifying error band of histograms, I had a suggestion in mind for a while (I might open a new issue if that's tricky): would it be possible to have hatched areas instead of full colored area? This looks way nicer for such a band:

that's https://github.com/gonum/plot/issues/564

Issue #751, opened 2020-06-05
hplot: handle hstack+band with empty bins


sbinet:
to fix #749 , we introduced "empty bins" for the bands that display error bars, in a log-y environment.

we need to come up with a new way to handle colored/filled error bar "boxes" as this "connects" 2 bins (separated by an empty bin).
see:

![img](https://github.com/go-hep/hep/raw/24781f7c/hplot/testdata/hstack_logy_golden.png)

rmadar:
I think we could try to send the problematic bin to zero for the band as well. But of course, if we don't want to destroy the band in the two bins around the problematic one, we have to find a "trick". 

One idea could be to add two points on right of `i-1` bins and two points on the left of the `i+1` bins, with  the following x values: `xHighEdge[i-1]+epsilon` and `xLowEdge[i+1]-epsilon`. These four points would have `y=0`. Actually, I even wonder if we could take `epsilon=0` ...

OK, one has probably to try to see how it looks. If you think it's good-enough idea (this is still a bit dirty so ...), I can give a shot.

Issue #755, opened 2020-06-13
hplot: adding a generic binned graph type?


rmadar:
It is relatively usual to compute bin-by-bin quantities from various histograms, and the result is not always an histogram, *i.e.* (weighted) events counts. The most typical example is a bin-by-bin ratio of histograms, but there are more (bin-by-bin composition, bin-by-bin significance, etc ...).

The most natural option - I think - would be to write a new generic plotter, let's say `BinnedGraph`, to deal with binned information. This would mainly be a copy paste of `hplot.H1D` except that the `Hist` field would be replaced by another type (and I think all the plotting aspects are similar):
```go
type BinnedQuantity struct {
  binning hbook.Range      // I am not sure which is the best type.
  Values  plotter.Values   // y values
  Errors  plotter.YErrors  // y errros
}
```
Such a new type could have arbitrary `Values` and `Errors` and would allow store any operation between several histograms, *via* syntax like:
```go
bg := NewBinnedGraph(h.Binning)
for i:=0 ; i<h.Len() ; i++ {
  v1, e1 := h1.Value(i), h1.Error(i)
  v2, e2 := h2.Value(i), h2.Error(i)
  
  bg.SetValue(i, val(v1, v2))
  bg.SetError(i, err(v1, e1, v2, e2))
}
```

rmadar:
Putting back the discussion from https://github.com/go-hep/hep/pull/756 here
> I must say I am a bit hard-pressed to see exactly the difference w/, say, a hplot.S2D.
but it wouldn't be the first time something escaped me :)

I think the practical difference if not that large: the only real one is the ability of stacking different `BinnedGraph` while it requires some "tricks" in `hplot.S2D` (you need to either have the sames `x's` or interpolate between `x's`). This would be needed to produce the (in)famous composition plots (randomly chosen example ;-) ).

From a conceptual point of view, one has discrete values as `x's` while the other has intervals (which basically share everything with usual histograms - but the relation between `val=sum(w)` and `err = sum(w^2)`).

At the end, I am not sure what's the best between:
  + adapt `hplot.S2D` to accommodate the few user-cases (as we did w/ `HiSteps`)
  + "fork" histograms into real histograms and binned graph

I hope that's clearer !

rmadar:
[...]

PR #756, opened 2020-06-14
New plotter to handle histogram error bands

rmadar:
Fixes https://github.com/go-hep/hep/issues/751

I started with the name `BinnedErrBand` because it can be applied to any binned quantity (not only an histogram, *i.e.* event counts). This point is related to the discussion in https://github.com/go-hep/hep/issues/755 . Of course, it's just a proposal and this can be modified.


hplot/binerrband.go
```diff
@@ -0,0 +1,108 @@
...
+			plotter.XY{X: xmin, Y: y - ydo},
+			plotter.XY{X: xmin, Y: y + yup},
+			plotter.XY{X: xmax, Y: y + yup},
+			plotter.XY{X: xmax, Y: y - ydo},
+		}
+
+		// Polygon
+		// FIX-ME(rmadar): it would be better to draw only top
+		//                 and bottom horizontal lines, but
+		//                 not the vertical lines.
```
sbinet, line 80: this could be done by (presumably) providing a `nil` LineStyle to `Polygon` and then drawing the 2 top+bottom lines...

rmadar:
> thanks.
> (and apologies for the belated answer, a bit busy w/ Tomuvol/MIM...)

No problem!

> shouldn't this also be integrated into `HStack` (and/or `H1D`) ?

Indeed, I was just waiting to settle on the type of fields and APIs, to implement them into `HStack` and `H1D`.

> I am ok with using `[]hbook.Range` instead of a `[][2]float64`.
> 
> that said, perhaps this also means we'd need to have the YODA Counter class equivalent into `hbook`:
> 
> * https://yoda.hepforge.org/doxy/classYODA_1_1Counter.html
> 
> b/c, basically, what `BinnedErrBand` needs as input is a slice of values w/ :
> 
> ```go
> type Counter struct {
>     V float64
>     StdDev float64
>     XRange hbook.Range
> }
> ```

Yes, that would be quite useful (also if you think https://github.com/go-hep/hep/issues/755 is a good idea). Does that mean that this type has be propagated across every types that are involved in histograms...?

sbinet:
>> shouldn't this also be integrated into HStack (and/or H1D) ?
> Indeed, I was just waiting to settle on the type of fields and APIs, to implement them into HStack and H1D.

makes sense.

> Yes, that would be quite useful (also if you think #755 is a good idea).

I must say I am a bit hard-pressed to see exactly the difference w/, say, a `hplot.S2D`.
but it wouldn't be the first time something escaped me :)

> Does that mean that this type has be propagated across every types that are involved in histograms...?

not necessarily.
only if we do really gain something by doing that. I must admit: I am not sure.

rmadar:
> I must say I am a bit hard-pressed to see exactly the difference w/, say, a hplot.S2D.
but it wouldn't be the first time something escaped me :)

I have tried to better explained (directly in https://github.com/go-hep/hep/issues/755)

For the `Counter`-like type, I would be tempted to keep an asymmetric error (to account for Poisson uncertainties for e.g.). This is not possible in ROOT and this is quite annoying. So something like:
```go
type Counter struct {
    V float64
    XRange hbook.Range
    
    // Either this
    ErrDown, ErrUp float64

    // Or directly
    Err plotter.Error
}
```

hplot/binerrband.go
rmadar: Ha ... yes, of course! Done.

rmadar:
If you are happy with this, I'll propagate the changes to `hplot.H1D` and `hplot.HStack`.

rmadar:
We'll probably also need that the new type `BinnedErrBand` implements the `plot.GlyphBoxer` interface (cf. [this post](https://github.com/gonum/plot/wiki/Creating-Custom-Plotters:-A-tutorial-on-creating-custom-Plotters#the-glyphboxer-interface)) to have a proper axis auto-scaling, and avoid this:

image

hbook/h1d.go
```diff
@@ -728,6 +728,18 @@ scanLoop:
 	return err
 }
 
+// Counts return a slice of Count, ignoring outerflow.
+func (h *H1D) Counts() []Count {
+	cs := make([]Count, h.Len())
+	for i, bin := range h.Binning.Bins {
+		cs[i].XRange = bin.Range
+		cs[i].Val = bin.Dist.Dist.SumW
+		cs[i].Err.Low = math.Sqrt(bin.Dist.Dist.SumW2) * 0.5
```
sbinet, line 737: the way errors are filled for that slice of counts should probably documented in the `Counts()` method.
also: perhaps just use `Dist0D.errW()` ?

hbook/hbook.go
```diff
@@ -40,6 +40,17 @@ func (r Range) Width() float64 {
 	return math.Abs(r.Max - r.Min)
 }
 
+// Count is a 1-dim binned information, made of
+// a X range, a value and its uncertainties.
+type Count struct {
```
sbinet, line 45: could we also have an example for `Count` (with perhaps also modifying an example of `H1D` to use `Counts()`) ?

hplot/example_binerrband_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+import (
+	"image/color"
+	"log"
+
+	"golang.org/x/exp/rand"
+
+	"gonum.org/v1/gonum/stat/distuv"
+
+	"gonum.org/v1/plot/vg"
```
sbinet, line 15: please regroup these imports.

hplot/binerrband.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// FillColor is the color to fill the area
+	// between the top and bottom data points.
+	// Use nil to disable the filling.
+	FillColor color.Color
+}
+
+// NewBinnedErrBand creates a binned error band
+// from a slice of count.
+func NewBinnedErrBand(h *hbook.H1D) *BinnedErrBand {
```
sbinet, line 39: I think it would probably make more sense to have that function take directly a slice of `hbook.Count`.

hbook/hbook.go
rmadar: I did two examples, one with just a `hbook.Count` type and the other with `hbook.H1D` (but I didn't modify existing H1D example). Let me know if you think more (or different) examples would be useful - I am happy to add them.

hbook/h1d.go
rmadar: I have actually went for `h.Value(i)` and `h.Error(i)` - let me know if you prefer `Dist0`-based calls.

hplot/binerrband.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// Plot implements the Plotter interface,
+// drawing a colored box defined by width
+// of bins (x-axis) and error (y-axis).
+func (b *BinnedErrBand) Plot(c draw.Canvas, plt *plot.Plot) {
+
+	for _, count := range b.Counts {
+
+		// Don't do anything if both errors are zero
```
sbinet, line 52: is that true?
shouldn't we at least draw one of the top/bottom lines? (if their line width>0, that is)

rmadar:
> We'll probably also need that the new type `BinnedErrBand` implements the `plot.GlyphBoxer` interface (cf. [this post](https://github.com/gonum/plot/wiki/Creating-Custom-Plotters:-A-tutorial-on-creating-custom-Plotters#the-glyphboxer-interface)) to have a proper axis auto-scaling

Concerning this, I am not sure where it come from (since the `DataRange()` function of `BinnedErrBand` takes the errors into account). I have even tried to modify the `func (h *hbook.H1D) DataRange()` to include uncertainties, but it doesn't correct the problem.

For this PR, I think I am happy with this, but I might open another issue for the auto-scaling problem.

hplot/h1d.go
```diff
@@ -168,39 +168,40 @@ func (h *H1D) withYErrBars(yoffs []float64) *plotter.YErrorBars {
...
 func (h1 *H1D) withBand() error {
-
-	bins := h1.Hist.Binning.Bins
-	var (
-		top = make(plotter.XYs, 0, 2*len(bins))
-		bot = make(plotter.XYs, 0, 2*len(bins))
-	)
-
-	for i := 0; i < 2*len(bins); i++ {
+	/*
```
sbinet, line 171: just remove all this commented out block?

hplot/h1d.go
```diff
@@ -168,39 +168,40 @@ func (h *H1D) withYErrBars(yoffs []float64) *plotter.YErrorBars {
...
+				bin  = bins[ibin]
+				xmin = bin.XEdges().Min
+				xmax = bin.XEdges().Max
+				sumw = bin.SumW()
+				errw = bin.ErrW()
+				ymin = sumw - 0.5*errw
+				ymax = sumw + 0.5*errw
+			)
+			switch {
+			case h1.LogY && ymin <= 0:
```
sbinet, line 190: ah... this calls for some support for LogY in BinnedErrBand...

sbinet:
yes, the GlyphBoxer interface can be part of another PR :)

hplot/h1d.go
rmadar: Indeed. I have added a `LogY` field in `BinnedErrBand`, so that it allows its use even without an histogram.

hplot/binerrband.go
rmadar: Thinking about it, it does make sense to have at least the line, even if the error is 0. I'll add this.

rmadar:
> yes, the GlyphBoxer interface can be part of another PR :)

Well, I think it shouldn't be needed since the DataRange() uses `y-e` and `y+e`. And this issue was already present before the implementation of `BinnedErrBand`, so I think the issue it somewhere in `hplot.H1D` ... I am searching !

rmadar:
Ok, I have finally found (and solved) the issue for the band ... for *both* `H1D` and `HStack`.

hplot/hstack.go
```diff
@@ -119,10 +119,11 @@ func (hstack *HStack) DataRange() (xmin, xmax, ymin, ymax float64) {
 	for _, h := range hstack.hs {
 		for i, bin := range h.Hist.Binning.Bins {
 			sumw := bin.SumW()
+			errw := bin.ErrW()
 			xmax = math.Max(bin.XMax(), xmax)
 			xmin = math.Min(bin.XMin(), xmin)
-			ymax = math.Max(yoffs[i]+sumw, ymax)
-			ymin = math.Min(yoffs[i]+sumw, ymin)
+			ymax = math.Max(yoffs[i]+sumw+0.5*errw, ymax)
+			ymin = math.Min(yoffs[i]+sumw-0.5*errw, ymin)
```
sbinet, line 126: aha! nice catch.

but actually, I think the perhaps "right" fix is to uncomment the lines 140-148 below (the nice block prefixed with an ominous `// FIXME`) and replace the `if h.YErrs` with a loop over all histos and accumulate each of the band's DataRanges (if not nil).

what do you think?

rmadar: Indeed! It's bit more trickier than I anticipated, but I should be able to fix it in PR. I'll take case of it

sbinet: yeah, that's why I'd originally left it for later, being lazy and all :)

thanks for doing all this grunt work.

PR #768, opened 2020-06-25
csvutil: add support for writing slices

chin123:
This is useful for tools such as root2csv, which fail right now if the detected type is slice. Reading slices is not implemented because it comes with a few more complications such as parsing strings with quotes or the slice delimiter in them. However, if it seems appropriate I could implement that too.


csvutil/csv.go
```diff
@@ -27,6 +27,34 @@ func min(a, b int) int {
...
+	case reflect.String:
+		return rv.String(), nil
+	case reflect.Slice:
+		vals := make([]string, rv.Len())
+		for i := 0; i < rv.Len(); i++ {
+			if vals[i], err = formatValue(rv.Index(i).Interface()); err != nil {
+				return "", err
+			}
+		}
+		return "[" + strings.Join(vals, ",") + "]", nil
```
sbinet, line 52: is this a well known or widespread way to deal with non-scalar values in the CSV world?
what happens if the `csv.Writer.Comma` is `','` instead? wouldn't this confuse other tools?

csvutil/csv.go
```diff
@@ -27,6 +27,34 @@ func min(a, b int) int {
...
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		return strconv.FormatInt(rv.Int(), 10), nil
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+		return strconv.FormatUint(rv.Uint(), 10), nil
+	case reflect.Float32, reflect.Float64:
+		return strconv.FormatFloat(rv.Float(), 'g', -1, rt.Bits()), nil
+	case reflect.String:
+		return rv.String(), nil
+	case reflect.Slice:
+		vals := make([]string, rv.Len())
```
sbinet, line 46: perhaps use a `string.Builder` instead? (and take it as an argument to `formatValue`?)

csvutil/csv.go
chin123: You're right, it would cause confusion in that case. What pandas seems to do when writing non scalars is surrounding the array with `"` marks. That should resolve any ambiguity, does that sound fine?

sbinet: makes sense.
what does it do for:
- slices of slices of ints?
- slices of strings?

(so it's a pandas-only thing? or is it more widespread?)

sbinet: ok:

sbinet: ok:

```
>>> df = pd.DataFrame({'slice': [[[10,11],[2,3]],[[40,41],[5,6]]]})
>>> df.to_csv("foo.csv")
```

leads to:
```
,slice
0,"[[10, 11], [2, 3]]"
1,"[[40, 41], [5, 6]]"
```

and:
```
>>> df = pd.DataFrame({'slice': [[["10","11"],["2","3"]],[["40","41"],["5","6"]]]})
>>> df.to_csv("foo.csv")
```
to:
```
,slice
0,"[['10', '11'], ['2', '3']]"
1,"[['40', '41'], ['5', '6']]"
```

so if we "standardize" on what pandas does, we should try to replicate that behaviour (with a unit test :))
```
```

chin123: Oh, it looks like Go's csv automatically adds quotes if a comma is present in the field: https://golang.org/src/encoding/csv/writer.go?s=1503:1548#L38  
This was an interesting was to discover it :)

csvutil/csv.go
chin123: done!

chin123: not passed to formatValue though, I'm not sure how that would work...

csvutil/csv.go
```diff
@@ -27,6 +27,44 @@ func min(a, b int) int {
...
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+		return strconv.FormatUint(rv.Uint(), 10), nil
+	case reflect.Float32, reflect.Float64:
+		return strconv.FormatFloat(rv.Float(), 'g', -1, rt.Bits()), nil
+	case reflect.String:
+		if in_quotes {
+			return "'" + rv.String() + "'", nil
+		} else {
+			return rv.String(), nil
+		}
```
sbinet, line 47:
```suggestion
		if quotes {
			return "'" + rv.String() + "'", nil
		}
		return rv.String(), nil
```

csvutil/csv.go
```diff
@@ -27,6 +27,44 @@ func min(a, b int) int {
...
+	case reflect.Slice:
+		var slice strings.Builder
+		fmt.Fprintf(&slice, "[")
+		for i := 0; i < rv.Len(); i++ {
+			cur_elem, err := formatValue(rv.Index(i).Interface(), true)
+			if err != nil {
+				return "", err
+			}
+			fmt.Fprintf(&slice, "%s", cur_elem)
+			if i != rv.Len()-1 {
```
sbinet, line 57: better prepend the `,`.
ie: move that just after the `for` loop and as:
```go
if i > 0 {
    slice.WriteString(", ")
}
```

csvutil/csv.go
```diff
@@ -27,6 +27,44 @@ func min(a, b int) int {
...
+		}
+	case reflect.Slice:
+		var slice strings.Builder
+		fmt.Fprintf(&slice, "[")
+		for i := 0; i < rv.Len(); i++ {
+			cur_elem, err := formatValue(rv.Index(i).Interface(), true)
+			if err != nil {
+				return "", err
+			}
+			fmt.Fprintf(&slice, "%s", cur_elem)
```
sbinet, line 56:
```suggestion
			slice.WriteString(elem)
```

csvutil/csv.go
```diff
@@ -27,6 +27,44 @@ func min(a, b int) int {
...
+		if in_quotes {
+			return "'" + rv.String() + "'", nil
+		} else {
+			return rv.String(), nil
+		}
+	case reflect.Slice:
+		var slice strings.Builder
+		fmt.Fprintf(&slice, "[")
+		for i := 0; i < rv.Len(); i++ {
+			cur_elem, err := formatValue(rv.Index(i).Interface(), true)
```
sbinet, line 52:
```suggestion
			elem, err := formatValue(rv.Index(i).Interface(), true)
```

csvutil/csv.go
```diff
@@ -27,6 +27,44 @@ func min(a, b int) int {
...
+			if err != nil {
+				return "", err
+			}
+			fmt.Fprintf(&slice, "%s", cur_elem)
+			if i != rv.Len()-1 {
+				fmt.Fprintf(&slice, ", ")
+			}
+		}
+		fmt.Fprintf(&slice, "]")
```
sbinet, line 61:
```suggestion
		slice.WriteString("]")
```

csvutil/csv_test.go
```diff
@@ -391,7 +394,46 @@ func TestCSVWriterStruct(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("files differ: %+v\n", err)
+	}
+}
+
+func TestCSVWriterArgsSlice(t *testing.T) {
+	fname := "testdata/out-args-slice.csv"
+	tbl, err := csvutil.Create(fname)
+	if err != nil {
+		t.Errorf("could not create %s: %+v\n", fname, err)
```
sbinet, line 407:
```suggestion
		t.Fatalf("could not create %s: %+v", fname, err)
```

csvutil/csv_test.go
```diff
@@ -391,7 +394,46 @@ func TestCSVWriterStruct(t *testing.T) {
...
+		var (
+			lii = [][]int{{1, 2, 3}, {2, 3, 4}, {7, 8, 15}}
+			lss = [][]string{{"foo", "bar", "baz"}, {"abc", "def", "ghi"}, {"qwerty"}}
+			ls  = []string{"abc", "def", "ghi"}
+			f   = float64(i)
+		)
+		err = tbl.WriteRow(lii, lss, ls, f)
+		if err != nil {
+			t.Errorf("error writing row %d: %+v\n", i, err)
+			break
```
sbinet, line 427:
```suggestion
			t.Fatalf("error writing row %d: %+v", i, err)
```

csvutil/csv_test.go
```diff
@@ -391,7 +394,46 @@ func TestCSVWriterStruct(t *testing.T) {
...
+	tbl, err := csvutil.Create(fname)
+	if err != nil {
+		t.Errorf("could not create %s: %+v\n", fname, err)
+	}
+	defer tbl.Close()
+	tbl.Writer.Comma = ','
+
+	err = tbl.WriteHeader("## more complicated slices: [][]int{}, [][]string{}, []string{}, float64\n")
+	if err != nil {
+		t.Errorf("error writing header: %+v\n", err)
```
sbinet, line 414:
```suggestion
		t.Fatalf("error writing header: %+v", err)
```

csvutil/csv_test.go
```diff
@@ -391,7 +394,46 @@ func TestCSVWriterStruct(t *testing.T) {
...
+		err = tbl.WriteRow(lii, lss, ls, f)
+		if err != nil {
+			t.Errorf("error writing row %d: %+v\n", i, err)
+			break
+		}
+	}
+
+	err = tbl.Close()
+	if err != nil {
+		t.Errorf("error closing table: %+v\n", err)
```
sbinet, line 433:
```suggestion
		t.Fatalf("error closing table: %+v", err)
```

csvutil/csv_test.go
```diff
@@ -391,7 +394,46 @@ func TestCSVWriterStruct(t *testing.T) {
...
+	}
+
+	err = diff("testdata/write-results-slice.csv", fname)
 	if err != nil {
 		t.Errorf("files differ: %+v\n", err)
```
sbinet, line 438:
```suggestion
		t.Fatalf("files differ: %+v", err)
```
(yeah, this is probably one of my first packages for Go-HEP, it probably shows a bit: I didn't get all the idioms correctly)

csvutil/csv.go
```diff
@@ -172,21 +210,9 @@ func (tbl *Table) WriteRow(args ...interface{}) error {
 func (tbl *Table) write(args ...interface{}) error {
 	rec := make([]string, len(args))
```
sbinet, line 211: we could create a `strings.Builder` variable here, call `.Reset()` just after the `for i, arg := range args {` line and pass that builder (by pointer) to `formatValue`.
`formatValue` would thus only return an `error` (or `nil`)

but ok, it could be part of a subsequent PR, if you prefer.

chin123: Ok, I changed it in this PR.

sbinet:
needs https://github.com/go-hep/license/pull/12

Issue #790, opened 2020-08-03
hplot: consider adding a plot label helper


rmadar:
Trying to add labels on plot, I played a bit with https://godoc.org/gonum.org/v1/plot/plotter#Labels and I thought it might be convenient to have a new type in `hplot` to ease label manipulation. I'll try to submit a first version in a PR.

PR #791, opened 2020-08-03
Plot label helper

rmadar:
First stab for https://github.com/go-hep/hep/issues/790


hplot/label.go
```diff
@@ -0,0 +1,94 @@
...
+	Style draw.TextStyle // Style of the label
+
+	// Compute the position wrt canvas size. This
+	// feature would need to be implemented into the plotter.
+	// (There is an example in gonum/v1/plot/plotter).
+	CanPos bool
+}
+
+// NewLabel creates a new value of type Label.
+func NewLabel(x, y float64, txt string, opts ...LabelOption) Label {
```
sbinet, line 20: as is, I am not completely sure this function really adds value.
perhaps just remove it? (together w/ the LabelOption types and func-opts)

rmadar:
Thanks for the suggestions - I have implemented them. I have 3 questions:
 1. I am not sure how to handle the default `Label.Style` value (if nothing is set, the instruction `labels.TextStyle = []draw.TextStyle{l.Style}` crashes).
 2. If we don't use `hplot.NewLabel()` function, we'll have to create all labels as `&hplot.Label{...}` - since `p.Add()` take a `*plotter`. I don't know if this is a big deal, but as a user, I would prefer something symmetric for all objects (having sometime to put `&` and sometime to not put it might be confusing).
 3. I am not sure how to pass all the text style (color, font,  rotation, X/Yalign and TexHandler) in case of relative coordinates, *i.e.* using `da.FillString()` function.

hplot/label.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	// Add the text styles.
+	// FIXME(rmadar): this crashes when l.Style is not
+	//                assigned.
+	// labels.TextStyle = []draw.TextStyle{l.Style}
+
+	// Return the result
+	return labels
+}
```
sbinet, line 80:
```suggestion
func (lbl *Label) labels(dc draw.Canvas) *plotter.Labels {
    if lbl.Style == (draw.TextStyle{}) {
        sty := ...
        lbl.Style = sty
    }
    
    x := lbl.X
    y := lbl.Y
    if lbl.Normalized {
        x = dc.X(x) // well, this doesn't compile, but you get the gist of it :)
        y = dc.Y(y) // ditto
    }
    
    xys := plotter.XYLabels{
        XYs: []plotter.XY{{X:x, Y:y}},
        Labels: []string{lbl.Text},
        TextStyle: []draw.TextStyle{lbl.Style},
    }
    
    lbls, err := plotter.NewLabels(xys)
    if err != nil {
        panic(fmt.Errorf("hplot: could not create labels: %w", err))
    }
    
    return lbls
}
```

hplot/label.go
```diff
@@ -0,0 +1,85 @@
...
+		pLabels.Plot(c, plt)
+	}
+}
+
+// DataRange returns the minimum and maximum x and
+// y values, implementing the plot.DataRanger interface.
+func (l *Label) DataRange() (xmin, xmax, ymin, ymax float64) {
+	pLabels := plotterLabels(l)
+	return pLabels.DataRange()
+}
```
sbinet, line 49:
```suggestion
func (lbl Label) DataRange() (xmin, xmax, ymin, ymax float64) {
    if lbl.Normalized {
        return math.Inf(+1), math.Inf(-1), math.Inf(+1), math.Inf(-1)
    }
    return lbl.labels(nil).DataRange()
}
```

hplot/label.go
```diff
@@ -0,0 +1,85 @@
...
+	//                assigned.
+	// labels.TextStyle = []draw.TextStyle{l.Style}
+
+	// Return the result
+	return labels
+}
+
+var (
+	_ plot.Plotter    = (*Label)(nil)
+	_ plot.DataRanger = (*Label)(nil)
```
sbinet, line 84:
```suggestion
	_ plot.Plotter    = (*Label)(nil)
	_ plot.DataRanger = (*Label)(nil)
	_ plot.GlyphBoxer = (*Label)(nil)
```

hplot/label.go
```diff
@@ -0,0 +1,85 @@
...
+	pLabels := plotterLabels(l)
+	return pLabels.DataRange()
+}
+
+// GlyphBoxes returns a GlyphBoxe, corresponding
+// to the label, implementing the plot.GlyphBoxer interface.
+func (l *Label) GlyphBoxes(p *plot.Plot) []plot.GlyphBox {
+	pLabels := plotterLabels(l)
+	return pLabels.GlyphBoxes(p)
+}
```
sbinet, line 56:
```suggestion
func (lbl Label) GlyphBoxes(p *plot.Plot) []plot.GlyphBox {
    if lbl.Normalized {
        return []plot.GlyphBox{
            X: lbl.X,
            Y: lbl.Y,
            Rectangle: lbl.TextStyle.Rectangle(lbl.Text),
        }
    }
    return lbl.labels(nil).GlyphBoxes(p)
}
```

rmadar:
Thanks, I have implemented all you suggestions.

Just one thing: the `normalized` feature doesn't work because the code crashes in `func (lbl Label) GlyphBoxes(p)`, and I didn't manage to understand why. So don't be surprised if CI will fail.

(I am also planning to update the example with a custom TextStyle)

hplot/example_label_test.go
```diff
@@ -0,0 +1,48 @@
...
+		X:    -0.5, Y: 0.5,
+	}
+	l2 := hplot.Label{
+		Text: "... the right place",
+		X:    0.5, Y: -0.5,
+	}
+	l3 := hplot.Label{
+		Text: "This is the middle",
+		X:    0.5, Y: 0.5,
+		Normalized: true}
```
sbinet, line 29:
```suggestion
		Normalized: true,
	}
```

hplot/label.go
```diff
@@ -0,0 +1,107 @@
...
+	"gonum.org/v1/plot/vg/draw"
+)
+
+type Label struct {
+	Text      string         // Text of the label
+	X, Y      float64        // Position of the label
+	TextStyle draw.TextStyle // Text style of the label
+
+	// Normalized indicates whether the label position
+	//is in data coordinates  or normalized with regard
```
sbinet, line 23:
```suggestion
	// is in data coordinates or normalized with regard
```

hplot/label.go
```diff
@@ -0,0 +1,107 @@
...
+
+// Internal helper function to get plotter.Labels type.
+func (lbl *Label) labels(c draw.Canvas, p *plot.Plot) *plotter.Labels {
+
+	if lbl.TextStyle == (draw.TextStyle{}) {
+		// FIXME[rmadar]: implement the hplot.DefaultFont
+		//                and hplot.DefaultFontSize.
+		defaultFont, err := vg.MakeFont(plotter.DefaultFont, plotter.DefaultFontSize)
+		if err != nil {
+			panic("impossible to make font.")
```
sbinet, line 71:
```suggestion
			panic(fmt.Errorf("hplot: could not create font: %w", err))
```

hplot/label.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	xyL := plotter.XYLabels{
+		XYs:    []plotter.XY{{X: x, Y: y}},
+		Labels: []string{lbl.Text},
+	}
+
+	lbls, err := plotter.NewLabels(xyL)
+	if err != nil {
+		panic("cannot create plotter.Labels")
```
sbinet, line 95:
```suggestion
		panic(fmt.Errorf("hplot: could not create labels: %w", err))
```

hplot/example_label_test.go
```diff
@@ -0,0 +1,48 @@
...
+	p.X.Min = -1
+	p.X.Max = 1
+	p.Y.Min = -1
+	p.Y.Max = 1
+
+	// Adding the labels
+	p.Add(l1)
+	p.Add(l2)
+	p.Add(l3)
+
```
sbinet, line 43: perhaps also add a `plot/plotter.GlyphBoxes` plotter? (so we the glyph boxes are drawn).

hplot/label.go
```diff
@@ -0,0 +1,107 @@
...
+	pLabels := lbl.labels(draw.Canvas{}, nil)
+	return pLabels.DataRange()
+}
+
+// GlyphBoxes returns a GlyphBoxe, corresponding
+// to the label, implementing the plot.GlyphBoxer interface.
+func (lbl Label) GlyphBoxes(p *plot.Plot) []plot.GlyphBox {
+
+	if lbl.Normalized {
+		// FIXME[rmadar]: this crashes and I don't understand why.
```
sbinet, line 54: probably because `lbl.TextStyle` is the zero value of `draw.TextStyle`? (and thus the call to `lbl.TextStyle.Rectangle` is using zero values of a few of its fields)

either:
- introduce a little helper method `func (lbl *Label) style() draw.TextStyle` (and use it throughout)
- re-introduce your `NewLabel(...) *Label` function (in which we make sure to create a `TextStyle` field value with a meaningful default value) (and make `*Label` implement `plot.Plotter`, etc...)

perhaps option 2 above is better afterall (apologies for the back and forth on this matter)

hplot/label.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	return lbl.labels(draw.Canvas{}, p).GlyphBoxes(p)
+}
+
+// Internal helper function to get plotter.Labels type.
+func (lbl *Label) labels(c draw.Canvas, p *plot.Plot) *plotter.Labels {
+
+	if lbl.TextStyle == (draw.TextStyle{}) {
+		// FIXME[rmadar]: implement the hplot.DefaultFont
```
sbinet, line 67: perhaps use `hplot.DefaultStyle.Label` instead?

well, it's the font for the plot axis' labels... so font-size=12, whereas the gonum/plot/plotter.Labels default is to use a font-size of 10...

perhaps time for an overhaul of fonts handling. (again.)

sbinet:
there's also the corner (ha!) case of placing a label (say, `"Hellooooooooo"`) at `(x=1,y=1)` in normalized coordinates.
(this won't be a problem in data coordinates: `plot.DataRanger` and `plot.GlyphBoxer` will instruct `gonum/plot` to widen the data canvas window)

what should we do? (well, first, we should make a test for these corner cases)

- panic because the label string will fall off the canvas (on the right, and possibly on the top);
- have `NewLabel` return an error because the label string will fall off the canvas (on the right, and possibly on the top);
- automatically modify the label coordinates so the label string fits inside the data canvas;
- something else?

thoughts?

rmadar:
Yes, indeed! I would go for option 2.

I have another issue for the transformation of normalized coordinates into the "data" ones. The code I used:
```go
trX, trY := p.Transforms(&c)
x = float64(trX(X)) 
y = float64(trY(y))
```
is not what we need. The same stands for 
```go
dc := p.DataCanvas(c)
x = float64(dc.X(x)) 
y = float64(dc.Y(y)) 
```
Typically, the first code gives `xData, yData = 328.3, 190.8` for a position in the middle of the two axis, while it should be `0, 0` in the example.

If I am not wrong, we would need the "inverse" function `xRel := p.X.Norm(xData)`, that would give us `xData := p.X.Value(rRel)` or something. Do you know if this exists or if we should use another approach?

sbinet:
> If I am not wrong, we would need the "inverse" function xRel := p.X.Norm(xData), that would give us xData := p.X.Value(rRel) or something. Do you know if this exists or if we should use another approach?

I don't think that `p.X.Norm^-1` exists.
but we know the `Min/Max` of the axis, so coding that "inverse" function is just:
```go
dataCoord := func (xrel, xmin, xmax float64) float64 {
    return xmin + xrel*(xmax-xmin)
}
x = dataCoord(x, p.X.Min, p.X.Max)
y = dataCoord(y, p.Y.Min, p.Y.Max)
```
right?

as for the 3 choices above, ok for 2)
*but* also provide a func-opt to automatically adapt the normalized coordinates?

rmadar:
Yes, I thought about that, and then I thought "what if the axis is in log scale?". If the internal axis data remains linear, then we're good. Otherwise, we would need to handle this case too.

rmadar:
And the option to automatically adapt the coordinates is a good point.

sbinet:
> Yes, I thought about that, and then I thought "what if the axis is in log scale?". If the internal axis data remains linear, then we're good. Otherwise, we would need to handle this case too.

from the axis/plot point of view, only `min/max` are kept.
everything else is handled (stateless-ly) by `plot.Normalizer`:
- https://github.com/gonum/plot/blob/master/axis.go#L192
I think we're good.
(but perhaps also good to add a test for that as well)

(ah, and we should also probably panic if `x ∉ [0, 1]` or `y ∉ [0, 1]` when `Normalized` and document.)

rmadar:
OK very good, I'am on it.

Another question: how to I get the x and y size of a string (as it appears on the canvas, I mean, so depend on the font etc...)? I would need to know it to do something like:
```go
lgtX := xOnCan(lbl.Text)
lgtY := yOnCan(lbl.Text)

relXLgt  := p.X.norm(lgtX)
relYLgt  := p.Y.norm(lgtX)

if x + relXLgt  > 1 || y + relYLgt  > 1 {
 // recompute x and y so that it fits
}
```

sbinet:
the "bounding box" of a given string can be calculated w/ [draw.TextStyle.Rectangle](https://godoc.org/gonum.org/v1/plot/vg/draw#TextStyle.Rectangle).

rmadar:
OK I have made few modifications, but still few things missing:
 1. test with log scale axis (just matter of doing it)
 2. adding GlyphBoxes in the example (I didn't find how to plot it, since it's not a plotter apparently)
 3. the option `AutoAdujst` is not working.

For the last point, I have two problems. First, whatever the size of the text, it always fits (thanks to `DataRange()`) so I am not sure about the condition to be checked (`xRelMax>1`). But, of course, it depends if it is called before or after `Plot()` function. Second, `lbl.TextStyle.Rectangle(lbl.Text).Min` return values in `vg.Length` so I had to convert it into `float64` to call `p.X.Norm()` function on it. Are the values OK in that case (possible unit conversion issue?). I actually did some print out and I was confused.

Issue #821, opened 2020-11-17
groot: problem with slices in large file


rmadar:
Following a [discussion](https://root-forum.cern.ch/t/size-limit-in-t-buildindex-function/42318/15) in the ROOT forum, it turns out that reading back slices fails when the TTree is large.

sbinet:
I am trying to reproduce the failure with the following program:

```go
package main

import (
	"log"
	"math/rand"

	"go-hep.org/x/hep/groot"
	"go-hep.org/x/hep/groot/riofs"
	"go-hep.org/x/hep/groot/rtree"
)

func main() {
	f, err := groot.Create("o.root", riofs.WithoutCompression())
	if err != nil {
		log.Fatalf("error: %+v", err)
	}
	defer f.Close()

	var evt struct {
		Run    int64     `groot:"runNbr"`
		Evt    int32     `groot:"evtNbr"`
		NLep0  int32     `groot:"nlep0"`
		LepPt0 []float64 `groot:"lep_pt0[nlep0]"`
		NLep1  int32     `groot:"nlep1"`
		LepPt1 []float64 `groot:"lep_pt1[nlep1]"`
		NLep2  int32     `groot:"nlep2"`
		LepPt2 []float64 `groot:"lep_pt2[nlep2]"`
		NLep3  int32     `groot:"nlep3"`
		LepPt3 []float64 `groot:"lep_pt3[nlep3]"`
		NLep4  int32     `groot:"nlep4"`
		LepPt4 []float64 `groot:"lep_pt4[nlep4]"`
	}
	evt.LepPt0 = make([]float64, 0, 20)
	evt.LepPt1 = make([]float64, 0, 20)
	evt.LepPt2 = make([]float64, 0, 20)
	evt.LepPt3 = make([]float64, 0, 20)
	evt.LepPt4 = make([]float64, 0, 20)

	wvars := rtree.WriteVarsFromStruct(&evt)
	w, err := rtree.NewWriter(f, "truth", wvars)
	if err != nil {
		log.Fatalf("error: %+v", err)
	}
	defer w.Close()

	rnd := rand.New(rand.NewSource(1234))

	const N = 1e7
	for i := 0; i < N; i++ {
		if i%(N/100) == 0 {
			log.Printf("evt: %d...", i)
		}
		evt.Run = int64(i)
		evt.Evt = int32(i)
		evt.NLep0 = int32(rnd.Intn(10) + 1)
		evt.NLep1 = evt.NLep0
[...]

PR #824, opened 2020-11-20
groot/riofs: fix overlaping blocks w/ big-file mode

sbinet:
TKeys sizes computation was broken for "big files".
The code was indeed changing the `key.rvers` field to indicate
the "big file" mode was enabled, but it wasn't updating the
on-disk size of the TKey (ie: missing 8bytes).

This CL computes the correct size in one-go instead of fixing it
after the facts.

Fixes go-hep/hep#821.

### DIFF
--- a/groot/riofs/key.go
+++ b/groot/riofs/key.go
@@ -99,15 +99,16 @@ func newKey(dir *tdirectoryFile, name, title, class string, objlen int32, f *Fil
 	// if the key's payload is actually compressed, we introduce a hole
 	// with the f.setEnd call below.
 	k.nbytes = k.objlen + k.keylen
+	eof := f.end
 	if objlen > 0 {
-		k.seekkey = f.end
+		k.seekkey = eof
 		err := f.setEnd(k.seekkey + int64(k.nbytes))
 		if err != nil {
 			panic(err)
 		}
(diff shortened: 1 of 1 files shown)


sbinet:
@rmadar PTAL.

it would seem this fixes it for me.
(in my minimal reproducer at least. I was banging my head against my ill-written C++/ROOT code...)

rmadar:
Thanks a lot! Is there a change that you can merge it to master, so that I can try (more easily, let say)?

sbinet:
theoretically, you can try it out like so:

```
$> cd /some/where
$> git clone git@github.com:sbinet-hep/hep
$> cd ./hep
$> git checkout issue-821

$> cd /your/code
$> cat >> go.mod << EOF

replace go-hep.org/x/hep => /some/where/hep
EOF
```

(I'd rather not merge again something into `master` that isn't really proper :P)

rmadar:
Thanks for giving (again) the git commands (I have written them down this time). I tried and it works for me too!!

sbinet:
cool :)

`git` is a swiss-knife that must be relearned every time one reaches back for it :)

sbinet:
tagged as `v0.28.3`.

PR #828, opened 2020-11-26
hbook: introduce Rand1D

sbinet:


sbinet:
@rmadar @ebusato here's the equivalent to `TH1::GetRandom`.

PR #829, opened 2020-12-03
fmom: add SetPtEtaPhi{E,M} to PxPyPzE

sbinet:


sbinet:
@rmadar with this PR in, `fmom.PxPyPzE` should be an (almost) drop-in replacement of ROOT's `TLorentzVector`.

rmadar:
Yes, thanks for this! I am really thinking about using `fmom` instead of my home-made lorentz-vector. So this PR is more than welcome!

Issue #831, opened 2020-12-07
fmom: suggestion of adding a function to get 3-vector


rmadar:
This would be just quite convenient to have a function like:

```go                                                                                                                   
func (p fmom.P4) P3() r3.Vec {                                                                                                                                             
        return r3.Vec{X: p.Px(), Y: p.Py(), Z: p.Pz()}                                                                                                                    
}
```

sbinet:
SGTM.

rather than a method on each and every `fmom.P4` implementations, I'd rather see:
```go
// VecOf returns the spatial components of the provided 4-vector.
func VecOf(p fmom.P4) r3.Vec {
    return r3.Vec{...}
}
```

WDYT?
(want to give it a stab? +tests, of course :P)

PR #832, opened 2020-12-07
Implementing VecOf(p4) function

rmadar:
Addressing https://github.com/go-hep/hep/issues/831

### DIFF
--- a/fmom/ops_test.go
+++ b/fmom/ops_test.go
@@ -44,10 +44,6 @@ func deepEqual(p1, p2 P4) bool {
 	return Equal(p1, p2)
 }
 
-func deepEqualP3(p1, p2 r3.Vec) bool {
-	return p1.X == p2.X && p1.Y == p2.Y && p1.Z == p2.Z
-}
-
 func TestAdd(t *testing.T) {
 	for _, tc := range []struct {
 		p1   P4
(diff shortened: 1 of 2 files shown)


fmom/ops_test.go
```diff
@@ -44,6 +44,10 @@ func deepEqual(p1, p2 P4) bool {
 	return Equal(p1, p2)
 }
 
+func deepEqualP3(p1, p2 r3.Vec) bool {
+	return p1.X == p2.X && p1.Y == p2.Y && p1.Z == p2.Z
```
sbinet, line 48:
```suggestion
	return p1 == p2
```
should just work.

and thus, `deepEqualP3` doesn't carry too much of its weight.
I'd just drop it.

fmom/ops_test.go
```diff
@@ -390,3 +394,33 @@ func TestBoostOf(t *testing.T) {
...
+			p:    newPxPyPzE(NewPxPyPzE(10, 0, 10, 20)),
+			want: r3.Vec{X: 10, Y: 0, Z: 10},
+		},
+
+		{
+			p:    newPxPyPzE(NewPxPyPzE(10, 10, 0, 20)),
+			want: r3.Vec{X: 10, Y: 10, Z: 0},
+		},
+	} {
+		p := tc.p.Clone()
```
sbinet, line 418: that's not necessary, I'd just drop it (and use `tc.p` directly below)

fmom/ops_test.go
```diff
@@ -390,3 +394,33 @@ func TestBoostOf(t *testing.T) {
...
+
+		got := VecOf(p)
+
+		if !deepEqualP3(got, tc.want) {
```
sbinet, line 422:
```suggestion
		if got != tc.want {
```

fmom/ops_test.go
```diff
@@ -390,3 +394,33 @@ func TestBoostOf(t *testing.T) {
...
+			t.Fatalf("got= %#v\nwant=%#v", got, tc.want)
```
sbinet, line 423:
```suggestion
			t.Errorf("invalid spatial components for %#v: got= %#v\nwant=%#v", tc.p, got, tc.want)
```

sbinet (commented):
thanks.
I do have a few minor comments+suggestions.

fmom/ops_test.go
```diff
@@ -390,3 +394,33 @@ func TestBoostOf(t *testing.T) {
...
 	}
 }
+
+func TestVecOf(t *testing.T) {
+	for _, tc := range []struct {
+		p    P4
+		want r3.Vec
+	}{
+		{
+			p:    newPxPyPzE(NewPxPyPzE(0, 10, 10, 20)),
```
sbinet, line 404:
```suggestion
			p:    newPxPyPzE(NewPxPyPzE(0, 10, 20, 30)),
```

just to address any possible ambiguity.

fmom/ops_test.go
```diff
@@ -390,3 +394,33 @@ func TestBoostOf(t *testing.T) {
...
+			want: r3.Vec{X: 0, Y: 10, Z: 10},
+		},
+
+		{
+			p:    newPxPyPzE(NewPxPyPzE(10, 0, 10, 20)),
```
sbinet, line 409:
```suggestion
			p:    newPxPyPzE(NewPxPyPzE(10, 0, 20, 30)),
```
ditto.

fmom/ops_test.go
```diff
@@ -390,3 +394,33 @@ func TestBoostOf(t *testing.T) {
...
+			want: r3.Vec{X: 10, Y: 0, Z: 10},
+		},
+
+		{
+			p:    newPxPyPzE(NewPxPyPzE(10, 10, 0, 20)),
```
sbinet, line 414:
```suggestion
			p:    newPxPyPzE(NewPxPyPzE(10, 20, 0, 30)),
```

fmom/ops_test.go
rmadar: Ha ok, nice! 

One curiosity question on this: I though there is no operator overloading in go ... how `obj1 == obj2` can work for a "user-defined" structure?

sbinet: `==` and `!=` work for all values (except slices, the equality operator isn't defined for those: that's why one can't use slices as a key of a `map`).
so they do work nicely for `struct` values: it's just the field/field comparison (and recursively)

sbinet: (that's not the same for other operators such as `<`, `>` or `!`)

sbinet:
thanks.

PR #902, opened 2022-02-04
cmd/root2npy: improve memory usage

sbinet:
This CL modifies root2npy to process Trees column by column so only 1
numpy buffer is materialized in-memory.

```
$> benchstat ./ref.txt ./new.txt
name       old time/op    new time/op    delta
Process-8    2.48ms ± 1%    2.39ms ± 2%   -3.44%  (p=0.000 n=28+29)

name       old alloc/op   new alloc/op   delta
Process-8    3.50MB ± 1%    2.64MB ± 2%  -24.63%  (p=0.000 n=30+29)

name       old allocs/op  new allocs/op  delta
Process-8     3.41k ± 0%     3.28k ± 0%   -3.60%  (p=0.000 n=30+30)
```

### DIFF
--- a/cmd/root2npy/main.go
+++ b/cmd/root2npy/main.go
@@ -197,20 +197,6 @@ func process(oname, fname, tname string) error {
 	}
 	log.Printf("scanning leaves... [done]")
 
-	r, err := rtree.NewReader(tree, nt.args)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


sbinet:
PTAL @Jatharachill

Issue #920, opened 2022-02-25
xroot: improve performances


sbinet:
this is possibly related to #399.

consider:

```
$> time root-dump root://ccxrootdgotest.in2p3.fr:9001/tmp/rootio/testdata/SMHiggsToZZTo4L.root > /dev/null

real	1m8.557s
user	0m14.863s
sys	0m3.562s
```

running over the same file but with `http[s]://` (which, currently, downloads the whole file and then serves it locally):
```
$> time root-dump https://cern.ch/binet/big-file.root > /dev/null

real	0m5.454s
user	0m6.156s
sys	0m0.228s
```

a factor 10x is left on the floor.
(ok, not the same machine, on different networks, etc... but still)

Moelf:
I was just about to post a performance question:
```julia
julia> using UnROOT
[ Info: Precompiling UnROOT [3cd96dde-e98d-4713-81e9-a4a1b0235ce9]

julia> const txrd = LazyTree("root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleMuParked.root", "Events");
┌ Warning: Your tree is quite wide, with 32 branches, this will take compiler a moment.
└ @ UnROOT ~/.julia/dev/UnROOT/src/iteration.jl:294

julia> @time sum(txrd.nMuon[1:10^5])
  1.723536 seconds (3.45 M allocations: 173.713 MiB, 2.67% gc time, 29.63% compilation time)
0x0000000000039bbb

julia> @time sum(txrd.nMuon[1:10^5])
  1.174588 seconds (1.27 k allocations: 1.049 MiB)
0x0000000000039bbb

julia> const thttp = LazyTree("https://jiling.web.cern.ch/jiling/public/Run2012B_DoubleMuParked.root", "Events");
┌ Warning: Your tree is quite wide, with 32 branches, this will take compiler a moment.
└ @ UnROOT ~/.julia/dev/UnROOT/src/iteration.jl:294

julia> @time sum(thttp.nMuon[1:10^5])
  0.847792 seconds (146.22 k allocations: 9.412 MiB, 8.69% compilation time)
0x0000000000039647

julia> @time sum(thttp.nMuon[1:10^5])
  0.287621 seconds (752 allocations: 1.516 MiB)
0x0000000000039647

julia> @time sum(thttp.nMuon[1:10^5])
  0.279659 seconds (719 allocations: 1.516 MiB)
0x0000000000039647
```

[...]

PR #923, opened 2022-03-08
xrootd: improve read performances 10-fold

sbinet:
This CL reduces the lock contention on xrootd.File operations.

```
  $> benchstat ./ref.txt ./new.txt
  name    old time/op    new time/op    delta
  Read-8     67.2s ± 1%      5.4s ± 7%  -91.93%  (p=0.000 n=9+28)

  name    old alloc/op   new alloc/op   delta
  Read-8     343MB ± 0%     341MB ± 0%   -0.78%  (p=0.000 n=8+30)

  name    old allocs/op  new allocs/op  delta
  Read-8      277k ± 0%      288k ± 0%   +3.84%  (p=0.000 n=7+29)
```

and now:
```
  $> time root-dump root://ccxrootdgotest.in2p3.fr:9001/tmp/rootio/testdata/SMHiggsToZZTo4L.root > /dev/null

  real	0m7.279s
  user	0m8.221s
  sys	0m1.256s
```

compared to:

```
  $> time root-dump https://cern.ch/binet/big-file.root > /dev/null

  real	0m5.454s
  user	0m6.156s
  sys	0m0.228s
```

Updates #920.

### DIFF
--- a/xrootd/bench_test.go
+++ b/xrootd/bench_test.go
@@ -1,0 +1,53 @@
+// Copyright ©2022 The go-hep Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package xrootd_test
+
+import (
+	"testing"
+
+	"go-hep.org/x/hep/groot"
+	_ "go-hep.org/x/hep/groot/riofs/plugin/xrootd"
+	"go-hep.org/x/hep/groot/rtree"
+)
+
+func BenchmarkRead(b *testing.B) {
+	const (
+		fname = "root://ccxrootdgotest.in2p3.fr:9001/tmp/rootio/testdata/SMHiggsToZZTo4L.root"
+		tname = "Events"
+	)
+
+	f, err := groot.Open(fname)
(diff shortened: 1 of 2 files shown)


Moelf:
I want to understand this mutex lock a bit more. Is it not possible to read from a xrootd file handler in parallel?

sbinet:
it should. (IIRC, there's a test for that, with the `-race` detector).
what the mutex protects is the "session ID" (and a cached value of the remote "file info").
"session ID" is some UUID-like identifier used internally by xrootd (client and server) to multiplex operations.

sbinet:
> I think the answer is yes. But I don't know how it plays with the extern C call.

depending on how you call into go-hep/xrootd (presumably by generating a .so that wraps a Go type and exposes C-funcs), it shouldn't be an issue.

Moelf:
thanks. naively I would call this function from different threads.

https://github.com/JuliaPackaging/Yggdrasil/blob/38a8c9dd73fae3b39ec76fee0bb96162a1b6f138/X/xrootdgo/bundled/main.go#L59

PR #935, opened 2022-04-15
groot/riofs: export in-memory read-only RMemFile ROOT file

sbinet:
This CL export RMemFile, that provides a simple way to create in-memory
read-only ROOT files.


sbinet:
PTAL @aphecetche

PR #938, opened 2022-04-27
lhef: reduce number of allocs in Decoder

dolmen:
In the `lhef.Decoder`, allocate slices using sliced bigger chunks in order to reduce the number of allocations.

Example:
```
a := make([]int32, n)
b := make([]int32, n)
c := make([]int32, n)
```
is replaced with:
```
n32 := make([]int32, 3*n)
a := n32[:n:n]
b := n32[n : 2*n : 2*n]
b := n32[2*n:]
```

### DIFF
--- a/lhef/reader.go
+++ b/lhef/reader.go
@@ -122,12 +122,17 @@ Loop:
 		return nil, err
 	}
 
-	d.Run.XSECUP = make([]float64, int(d.Run.NPRUP))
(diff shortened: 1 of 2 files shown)


sbinet (approved):
LGTM

thanks for improving this. (it feels like good old `FORTRAN` column blocks, but hey :P)

PR #940, opened 2022-04-27
sliceop: use errors.New for constant errors

dolmen:
Use `errors.New` instead of `fmt.Errorf` when the message is a constant string.


dolmen:
2022-04-28 The [golang.org/x/exp/slices](https://pkg.go.dev/golang.org/x/exp/slices) is, like its import path implies, still experimental.

sbinet (approved):
LGTM

PR #951, opened 2022-06-22
brio-gen: cleanup files produced by tests on test success

dolmen:
In tests of brio-gen, when a failure occurs (generated output changes) the new version is kept on disk. This is the existing behaviour.

With this patch we remove the new version if there is no difference (test success). This allows to automatically remove those temporary files that are useful only for debugging a test failure.


sbinet (approved):
LGTM

PR #952, opened 2022-06-22
groot: fix go:generate portability for root-gen-type

dolmen:
Use `go run` instead of requiring the binary to be already built.


sbinet (approved):
LGTM

Issue #991, opened 2023-09-06
groot/rhist: consider adding JSON serialization for TH{1,2}x


sbinet:
this is a follow-up from https://github.com/go-hep/hep/issues/982.

We should consider providing JSON serialization of `TH{1,2}x`, compatible with the ROOT one.

PR #994, opened 2023-09-07
groot/rjson: first import

sbinet:
Fixes https://github.com/go-hep/hep/issues/991.


sbinet:
^^ @mrceyhun _ping_ ?

was this useful ?
(in the meantime, I've also added support for `TGraph`)

BTW, do you have a link to the repo that uses `go-hep.org/x/hep/groot/...` ?
(I'd like to add it to https://go-hep.org/users/#code)

PR #1018, opened 2024-10-25
groot/internal/rcompress: add parallel zlib compression

sbinet:


sbinet:
most of the performance improvements were actually coming from a zlib knob (level going from 9 to 6).
shelving this for now.

PR #1024, opened 2024-11-20
fit: fix mismatch returned val

zqkgo:


sbinet:
thanks for the PR.

actually, this file could just be totally removed (I vaguely remember this was my driver for testing the API of the at-the-time-new `fit` package), as the 2 functions it's testing are actually exercized in:

- https://pkg.go.dev/go-hep.org/x/hep/fit#example-Curve1D-Gaussian
- https://pkg.go.dev/go-hep.org/x/hep/fit#example-Curve1D-Exponential

would you mind updating your PR ?
thanks again.

